PR #4, opened 2015-06-24
Imports assets from bootstrap's woff2 file

zfoster:
See: https://github.com/twbs/bootstrap/tree/master/fonts

I'm pretty new to this, but I noticed I had to add this line to my index.js in order to get some of the glyphicons to render, as they were hitting 404s like: 
```GET http://localhost:4200/fonts/glyphicons-halflings-regular.woff2```

Let me know if this is a mistake on my end or if there's a place I can write a test for another import like this


simonihmig:
Seems to have been forgotten, thanks for your effort!

PR #6, opened 2015-06-30
Added support for ember-cli 0.2.7 along with ember-try.

kpavankumar:
Also changed the replaced deprecated syntax of {{bind-attr}} with HTMLBars Style attributes and modified {{each}} helpers


simonihmig:
Thanks, looks good!

PR #9, opened 2015-07-16
Support for Ember 1.13.x

jesenko:
Current version of ember-bootstrap uses some private api that was changed with the introduction of Glimmer engine with 1.13.x. 

This PR includes:
  * parent - children components use `nearestOfType` for communication of `childViews` is no longer supported in Ember
  * package.json and bower.json dependencies to latest stable versions.
  * fixed tests


jesenko:
All tests except one pass for build on stable ember version. `bs-button-test.js`, `button has HTML attributes` fails due to https://github.com/emberjs/ember.js/issues/11480.

PR #11, opened 2015-07-18
Fix wrong checkbox class for horizontal layout.

jesenko:


simonihmig:
Actually all form elements in a horizontal form suffer form the same bug, see https://github.com/kaliber5/ember-bootstrap/issues/14.

Will close this in favor of a complete bug fix.

simonihmig:
Fixed in v0.1.4

PR #23, opened 2015-09-27
Dynamic modal dialog & Proxy Button

danshapir:
Hey,

I've added a new modal service and component inspired by the bootstrap-for-ember original addon with lots of tweaks and major changes I've done over a year of development with this addon + made it fully 2.0 compatible. I've also changed the bs-button to use the ProxyMixin so that it can be dynamically created as the modal requires while being still being bound to the caller content.

It's fully battle proof on a few production clients.

P.S
Just in case I did a merge with your branch just a few days ago so we are fully synced :)

Thanks,
Dan


simonihmig:
Hi Dan,

thanks for your effort. However there are a couple of issues which make me unable to merge this:
* existing tests are failing
* the new functionality is not covered by tests
* the new functionality lacks API docs
* there are references to ember-bootstrap-components all over the place

Moreover there is an existing branch in development which will add modal support, so this will be the implementation that ember-bootstrap will incorporate, which is probably quite different to yours.

If you want to make your effort publicly available, I would suggest you release this as an independent modal-focused addon, maybe requiring ember-bootstrap in its dependencies!

Regards,
Simon

PR #25, opened 2015-10-22
Fix modal close action

anilmaurya:


simonihmig:
Hi there, I just can't see why this should be needed. In fact I think this would break things, at least in the dummy app inside this addon the modal page should be working without this change. Can you show me the way you use the modal component that would need this change?

anilmaurya:
@simonihmig I am using modal component inside a component. Without this change , I am getting `close` not defined for `ember-wormhole` component.

simonihmig:
@anilmaurya I guess there must be something wrong in your template, probably the nesting of modal components... can you show me an example?

anilmaurya:
@simonihmig This is only a workaround to make it work in my application. However there are some more issues which are not resolved by this fix, like close action not working without bs-modal-header component.

I am closing this pull-request for now , will come up with better solution and example.

PR #26, opened 2015-10-24
replace @each with [] to follow ember 2.0+ conventions

mlennie:
@each was causing an error.

### DIFF
--- a/addon/components/bs-button-group.js
+++ b/addon/components/bs-button-group.js
@@ -160,7 +160,7 @@ export default Ember.Component.extend(SizeClass, {
      */
     value: undefined,
 
-    _syncValueToActiveButtons: Ember.observer('value','childButtons.@each.value',function(){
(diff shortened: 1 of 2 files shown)


addon/components/bs-button-group.js
```diff
@@ -160,7 +160,7 @@ export default Ember.Component.extend(SizeClass, {
      */
     value: undefined,
 
-    _syncValueToActiveButtons: Ember.observer('value','childButtons.@each.value',function(){
+    _syncValueToActiveButtons: Ember.observer('value','childButtons.[].value',function(){
```
jcope2013: the @each should stay here since your chaining another property https://github.com/emberjs/ember.js/pull/11994

simonihmig: As @jcope2013 said I think this should be correct. Closing this.

PR #27, opened 2015-10-25
Wire up <label> for attribute used in form-element component

cyk:
Using the strategy from [this stack overflow answer](http://stackoverflow.com/a/32895607/747731?stw=2), form elements within the `form-element` component can become a labeled control.


simonihmig:
Good point, thanks for your effort! Two thinks prevent this from getting merged directly however:
* there is already a concat helper in ember, see http://emberjs.com/api/classes/Ember.Templates.helpers.html#method_concat
* tests would be needed to check this new feature for the different form elements. Can you add those?

cyk:
I removed the redundant helper and added tests for labeled control feature.

simonihmig:
Look good, just one thing: you added tests actually only for the controlType=text case, the tests for textarea and checkbox are missing the correct controlType and selector!

cyk:
Oops. Sorry about that. Thanks for your patience helping me get this PR together.

simonihmig:
@cyk Thanks, has been merged! :)

PR #31, opened 2015-11-27
Use bootstrap theme to override default bootstrap css settings.

huberts:
According to project documentation, setting 'importBootstrapTheme' to true should enable the Bootstrap Theme. It is not working, because of the order of attaching the CSS files (first goes the theme, second goes the default Bootstrap CSS file).
This pull request fixes the - imho - bug.


simonihmig:
Thanks!

PR #32, opened 2015-11-30
Load whole Bootstrap.js by default

martndemus:
And only load modules when the user configures them.

It's pretty confusing to not have all bootstrap things available from the get go and have no way the enable them.


simonihmig:
Sorry, but this won't get merged. The approach this addon follows is to implement Bootstrap's JS functionality using native ember components. Very often things won't work out of the box just by including  bootstrap.js in a SPA like an ember app. You are free to include whatever you want from the bower installation using your app's local ember-cli-build.js however.

PR #37, opened 2016-01-12
Implement renderInPlace flag on bs-modal

bendemboski:
Causes the modal to render in place instead of being wormholed.

### DIFF
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -109,6 +109,8 @@ var observeOpen = function () {
  element, which is a child of ember's root element. This will make sure that modals always overlay the whole app, and
  are not effected by parent elements with `overflow: hidden` for example.
 
+ If you want the modal to render in place, rather than being wormholed, you can set renderInPlace=true.
+
  @class Modal
  @namespace Components
  @extends Ember.Component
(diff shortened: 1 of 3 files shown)


simonihmig:
Looks pretty good. No objections to merge this, however - just out of curiosity - is there a certain use case, where this (rendering in place) would be required?

bendemboski:
My app requires that non-UI code is able to show modals (e.g. a particular error response from the server triggering a login modal), and global policies are enforced on modals, such as only allowing one to be shown at a time, or dismissing them when a transition occurs.

To achieve this, I have a service that stores the data model for the modal (the name and state for a component to be dynamically rendered using the ```{{component}}``` helper), and methods to show/hide modals, and then a single component in ```application.hbs``` that gets passed the service from the application controller and then renders the modals inside of it as the data in the service indicates.

So, I already have them in a top-level global location, and don't want the extra overhead or testing complexity of ember-wormhole. Granted this isn't that much overhead, but I don't need it so I'd prefer to just dispense with it.

simonihmig:
Ok, I see. Thanks for your support!

bendemboski:
You bet! Thanks for the awesome library!

PR #39, opened 2016-01-19
Added insertEmberWormholeElementToDom configuration option

jonnedeprez:
I made it possible to disable the automatic creation of the ember-wormhole DOM element. I'm using this addon in a context where the Ember boot process is delayed, and as a result the Ember rootElement doesn't yet exist when the initializer runs. If a developer sets this flag to false, he will need to add that DOM element manually.


simonihmig:
Thanks

PR #51, opened 2016-02-23
Remove hardcoding of `isDevelopingAddon`.

rwjblue:
By default `isDevelopingAddon` is `true` when running `ember serve`, `ember test`, or `ember test --server` within the addon itself but `false` when running inside consuming applications.

The way this was previously, always forced `isDevelopingAddon` to be `true`, which makes consuming applications fail JSHint because this addon does.


simonihmig:
Thanks @rwjblue. Didn't notice this was still left in place from the early days...

PR #60, opened 2016-03-03
Implement autofocus support

bendemboski:
When a modal is shown, its root element is given focus so keyboard events can be handled. This change allows developers to use the HTML5 autofocus attribute to indicate that an input element should be given focus, instead of the root element.

This also involved propagating the autofocus attribute through a couple of layers of for/input elements.

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -172,6 +172,15 @@ export default FormGroup.extend({
   placeholder: null,
 
   /**
+   * Control element's HTML5 autofocus attribute
+   *
+   * @property autofocus
+   * @type boolean
+   * @public
+   */
+  autofocus: false,
+
+  /**
    * Control element's name attribute
    *
    * @property name
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -101,6 +101,25 @@ var observeOpen = function () {
  The modal component supports this common case by triggering the submit event programmatically on the body's form if
  present whenever the footer's submit button is pressed, so the example above will work as expected.
 
+ ### Auto-focus
+
+ In order to allow key handling to function, the modal's root element is given focus once the modal is shown. If your
+ modal contains an element such as a text input and you would like it to be given focus rather than the modal element,
+ then give it the HTML5 autofocus attribute:
+
+  ```hbs
+ {{#bs-modal title="Form Example" body=false footer=false}}
+   {{#bs-modal-body}}
+     {{#bs-form action=(action "submit") model=this}}
+       {{bs-form-element controlType="text" label="first name" property="firstname" autofocus=true}}
+       {{bs-form-element controlType="text" label="last name" property="lastname"}}
+     {{/bs-form}}
+   {{/bs-modal-body}}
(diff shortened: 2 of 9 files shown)


addon/components/bs-form-element.js
```diff
@@ -172,6 +172,15 @@ export default FormGroup.extend({
   placeholder: null,
 
   /**
+   * Control element's HTML5 autofocus attribute
+   *
+   * @property autofocus
+   * @type string
+   * @public
+   */
+  autofocus: null,
```
simonihmig: This should be a boolean property according to HTML5 specs. Needs updating the example in the docs as well.

tests/integration/components/bs-modal-test.js
```diff
@@ -228,6 +228,89 @@ test('when modal has a form and the submit button is clicked, the form is submit
   this.$('.modal .modal-footer button[type=submit]').click();
 });
 
+test('modal element is focused when no [autofocus] element is present and fade=false', function (assert) {
```
simonihmig: I think the first two tests ("modal element is focused ...") are not needed, even a bit contra productive in the sense that you should write test for public interfaces and not (private) implementation details. Focusing the modal is an implementation detail, the feature is that pressing the escape key closes the modal. This should be testet, and this test is in place already. So just drop these two tests.

However another test that checks if closing the modal with the escape key works when an input element is autofocused would be nice!

simonihmig:
Apart from the two little remarks this one looks pretty good. Thanks for your effort so far! :)

bendemboski:
Okay, I've pushed a couple of new commits addressing your feedback. If they look good, give me a :+1: and I'll rebase them so there's a clean git history when this PR gets merged.

bendemboski:
Okay, last comment addressed, rebased and ready to merge.

simonihmig:
Great work, thanks a lot! :)

addon/components/bs-modal.js
```diff
@@ -407,6 +426,20 @@ export default Ember.Component.extend({
   _observeOpen: Ember.observer('open', observeOpen),
 
   /**
+   * Give the modal (or its autofocus element) focus
+   *
+   * @method takeFocus
+   * @private
+   */
+  takeFocus() {
+    let focusElement = this.$('[autofocus]').first();
```
simonihmig: @bendemboski Did the autofocus feature actually work for you? The tests were fine, and were all passing using PhantomJS. However they failed in Chrome! And indeed the field was never focused in Chrome. 

Which seems right because the above jQuery selector does not select anything, as the autofocus field is not a child of the modal component with the DOM, because the modal-dialog is whormholed to another part of the DOM. This line should be like this:

        let focusElement = this.get('modalElement').find('[autofocus]');

I have a fix ready, however I was just wondering a) did it work for you?, an b) why tests were passing in PhantomJS but failing in Chrome... strange...

bendemboski: Huh, yeah that's weird, not sure why they are passing in PhantomJS. The functionality works for me because I use the renderInPlace flag to disable the ember-wormhole behavior...

PR #62, opened 2016-03-04
Implement openAction

bendemboski:
openAction will fire just after the modal has been shown, but before the transition animation, and openedAction will fire after the modal is completely shown, including transition animation.

The idea is that openAction can be used to access and manipulate the DOM at a time when it's no longer display: none, so size calculations and the like can be performed, and openedAction can be used to perform actions after the entire show animation is complete, such as setting focus, starting custom in-modal animations, etc.

Note that this technically breaks backwards compatibility, as it changes the timing of when openedAction fires.

### DIFF
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -392,6 +392,19 @@ export default Ember.Component.extend({
   closedAction: null,
 
   /**
+   * The action to be sent when the modal is opening.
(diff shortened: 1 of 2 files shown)


bendemboski:
@simonihmig is this what you had in mind as far as openAction vs. openedAction?

simonihmig:
Very nice :)

PR #69, opened 2016-03-21
Create validation hooks for external libs and fix config loader

### LINKED ISSUE #68: Custom icons dont get set in form groups
I have the following in my confg/environment.js

```
  'ember-bootstrap': {
      formValidationSuccessIcon: 'icon icon-check',
      formValidationErrorIcon: 'icon icon-error',
      formValidationWarningIcon: 'icon icon-warning',
      formValidationInfoIcon: 'icon icon-info'
    }
```

I can see that Config.load is getting called from the initializer but it doesnt look like its getting set in the form group, errorIcon is looking up `Config.formValidationErrorIcon` which is its default value.

offirgolan:
Hi There,

Im one of the creators of [ember-cp-validations](https://github.com/offirgolan/ember-cp-validations) and I think this addon is fantastic. Im creating this PR to move some code around into hooks which I can easily create a 3rd party addon for anyone that wants to use ember-bootstrap + ember-cp-validations. I currently have an alpha ([ember-bootstrap-cp-validations](https://github.com/offirgolan/ember-bootstrap-cp-validations)) but I think it's a little messy and will be difficult to maintain without some important changes to this library. 

(also Resolves #68)

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -139,6 +139,19 @@ export default Ember.Component.extend({
   }),
 
   /**
+   * Validate hook that will either resolve if the model is valid or reject if it's not
+   *
+   * @param Function resolve
+   * @param Function reject
+   * @public
+   */
+  validate(resolve, reject) {
+    this.get('model').validate().then(() => {
+      return this.get('model.isValid') ? resolve() : reject();
+    }, reject);
+  },
+
+  /**
    * Submit handler that will send the default action ("action") to the controller when submitting the form.
    *
    * If there is a supplied `model` that supports validation (`hasValidator`) the model will be validated before, and
@@ -156,18 +169,12 @@ export default Ember.Component.extend({
     if (!this.get('hasValidator')) {
       return this.sendAction();
     } else {
-      return this
-        .get('model')
-        .validate()
-        .then(() => {
-          if (this.get('model.isValid')) {
-            return this.sendAction();
-          }
-        })
-        .catch(() => {
-          this.get('childFormElements').setEach('showValidation', true);
-          return this.sendAction('invalid');
-        });
+      return new Ember.RSVP.Promise((resolve, reject) => this.validate(resolve, reject)).then(() => {
+        return this.sendAction();
+      }, () => {
+        this.get('childFormElements').setEach('showValidation', true);
+        return this.sendAction('invalid');
+      });
     }
   },
 
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -139,16 +139,17 @@ export default Ember.Component.extend({
   }),
 
   /**
-   * Validate hook that will either resolve if the model is valid or reject if it's not
+   * Validate hook that will return a promise that will either resolve if the model is valid
+   * or reject if it's not.
    *
-   * @param Function resolve
-   * @param Function reject
+   * @param Object model
+   * @return Promise
    * @public
    */
-  validate(resolve, reject) {
-    this.get('model').validate().then(() => {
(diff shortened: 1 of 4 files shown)


simonihmig:
@offirgolan Thanks for your effort, and for getting involved into this! :)

I just feel we have to think about the design a little bit more, to get this right. I do like your approach to decouple the validation integration by moving this out to a separate addon. However then it would be more consistent to also remove the hard wired `ember-validations` support into another addon. This would be an breaking change obviously, but worth it to keep the API design clean imho. What do you think?

Another thing: do we really have to have the `setupValidations` method for our purpose? It uses `defineProperty`, which is a private API. Another approach:

- `bs-form-element` looks for an `errors` property (empty by default) for validation messages
- an validation supporting addon (ember-bootstrap-cp-validations for ember-cp-validations, or ember-bootstrap-validations for ember-validations) specifies to be loaded after ember-bootstrap, thus being able to override any ember-bootstrap components (hopefully that works, see https://github.com/ember-cli/ember-cli/pull/1132) 
- these addons override `bs-form-element` (supplying an `errors` computed property) and `bs-form` (implementing the `validate()` method), just by importing the base component from ember-bootstrap and reexporting the extended one, so the app sees the extended components with added validation support. No need for an initializer or reopening of classes.

Did I miss something?

Maybe I could create a separate feature branch for this, to be able to work on this together, before merging this to master. This way you could adopt https://github.com/offirgolan/ember-bootstrap-cp-validations to the new API, and I could provide another addon for ember-validations, so we see if this works in the real world, and release it when everything is ready?

Would be nice to have the fix for the config bug in a separate PR, for getting this merged in advance!

Thanks a lot again so far!

addon/components/bs-form.js
```diff
@@ -139,6 +139,19 @@ export default Ember.Component.extend({
   }),
 
   /**
+   * Validate hook that will either resolve if the model is valid or reject if it's not
+   *
+   * @param Function resolve
+   * @param Function reject
+   * @public
+   */
+  validate(resolve, reject) {
```
simonihmig: I think this would be cleaner if it would just return a Promise by itself, instead of giving the promise callbacks as arguments!?

offirgolan:
> it would be more consistent to also remove the hard wired ember-validations support into another addon

Agreed. I can make the changes and create empty hooks.

> do we really have to have the setupValidations method for our purpose

Although it would be empty until a 3rd party lib overwrites it, I think it's good to have since there is some logic (i.e. isBlank(property) ) before setup. This will allow you to roll out changes to the init method or any other code surrounding the setupValidations hook easily without having to modify the 3rd party libs. 

>  It uses defineProperty

From my experience, I find `defineProperty` to be more stable, more commonly used, and more reliable than Ember.Binding. It also allows you to define a property with Ember.computed which gives you access to readOnly, oneWay, custom computed properties, and more. 

> Maybe I could create a separate feature branch for this, to be able to work on this together

Whatever makes this easier for you! Im up for anything really :smile_cat:. I can make the changes you suggested and then we can move that into a new feature branch.

addon/components/bs-form.js
offirgolan: Totally agree!

offirgolan:
@simonihmig Changes have been made as discussed.

simonihmig:
Ok, looks fine. Thanks again!

I created a "validations" branch. Could you please create new PR targeting that branch. I will close this one now. 

When I have the "ember-bootstrap-validations" addon ready, and probably added some documentation, I will merge it to master.

You will then update your https://github.com/offirgolan/ember-bootstrap-cp-validations addon  to use the new API, right? Will then add some infos + links to the docs...

PR #70, opened 2016-03-21
Validation hooks

offirgolan:


offirgolan:
@simonihmig feel free to message me in ember-community slack if that's easier for you :smile_cat:

PR #72, opened 2016-03-21
Add beforeSubmit hook

offirgolan:
There are many times where we may want to normalize data before we validate/submit

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -139,7 +139,7 @@ export default Ember.Component.extend({
   }),
 
   /**
-   * Validate hook that will return a promise that will either resolve if the model is valid
+   * Validate hook which will return a promise that will either resolve if the model is valid
    * or reject if it's not.
    *
    * @param Object model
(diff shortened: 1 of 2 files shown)


offirgolan:
@simonihmig, I made another change to pass available arguments to their actions. This is necessary if you're dependent on a server returned error message or some specific post validation check.

simonihmig:
Thanks again @offirgolan! The hook and passing the results of validations to the actions does make sense. Why is passing the raw event object `e` necessary? It is a rather low level thing I would tend to not pass around to the app unless absolutely necessary for some use cases...

Things that should be added before merging:

- [ ] mention the beforeSubmit action and the passing of the results to docs ("Submitting the form")
- [ ] add a simple test for asserting that beforeSubmit is called

offirgolan:
@simonihmig I wasn't sure it was really necessary or not tbh but ill remove it. To follow your naming convention, do you want to call this action `before` or keep it as `beforeSubmit`?

offirgolan:
@simonihmig Added docs and tests. Also removed `e` from action calls and renamed `beforeSubmit` to `before`.

addon/components/bs-form.js
```diff
@@ -161,6 +175,7 @@ export default Ember.Component.extend({
    * messages will be shown automatically.
    *
    * @event submit
+   * @param Object result The returned result from the validate method
```
simonihmig: This added @param statement is incorrect. `e` ist the DOM submit event.

offirgolan:
@simonihmig fixed it by exposing `action` as a public action handler.

simonihmig:
Travic build was failing because of some trivial JSCS code styling issues, please fix them if still present....

offirgolan:
Should be fixed now

PR #74, opened 2016-04-03
add `novalidate` attr to attributeBindings array

kepek:


simonihmig:
A couple of things would be needed:

- make the property explicit, defining it as false by default
- add documentation (yuidoc style) for the property
- add a simple test to https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-form-test.js

simonihmig:
Replaced by #94

PR #93, opened 2016-05-27
Fix: override model on bs-form-element

jelhan:
`model: computed.alias('form.model')` sets `form.model` to model passed in {{bs-form-element}}.

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -270,7 +270,7 @@ export default FormGroup.extend({
    * @property model
    * @public
    */
-  model: computed.alias('form.model'),
+  model: computed.reads('form.model'),
 
   /**
    * The array of error messages from the `model`'s validation.


addon/components/bs-form-element.js
```diff
@@ -270,7 +270,7 @@ export default FormGroup.extend({
    * @property model
    * @public
    */
-  model: computed.alias('form.model'),
+  model: null,
```
simonihmig: It should be sufficient to just declare the property as `computed.reads('form.model')` (one way binding), without the additional code in `init()`, or do I miss something? This way it would default to `form.model`, but would allow to be overridden without changing the form`s `model`. That's what you are trying to achieve, aren't you?

jelhan: Ähh, yeah of course. Updated PR.

PR #94, opened 2016-05-27
bs-form: adds support for novalidate attribute

jelhan:
Allows to disable browser form validation which may interfere in an undesired way with application validation logic.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -72,6 +72,7 @@ const { computed } = Ember;
 export default Ember.Component.extend({
   tagName: 'form',
   classNameBindings: ['layoutClass'],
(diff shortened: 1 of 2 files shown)


addon/components/bs-form.js
```diff
@@ -143,6 +144,11 @@ export default Ember.Component.extend({
   submitOnEnter: false,
 
   /**
+   * If set to true novalidate attribute is present on form element
+   */
+  novalidate: false,
```
simonihmig: - [ ] Please add the yuidoc tags @property, @type, @default and @public for correct API docs generation. 

Apart from that it looks good, thanks!

jelhan: Done

PR #95, opened 2016-05-28
Fix: update showValidations on form elements wrapped by another component

jelhan:


simonihmig:
I see your use case here, and it certainly is a valid one. 

However I have some concerns over the implementation:
* it uses an method instead of a computed property, which lacks caching for example. And making it a CP would be more ember way I guess. I think your implementation would still allow turning this into a CP with some slight modifications, but...
* `childViews` is a private API, which I would not use anymore. Not your fault, it was used before, probably for historical reasons. However there is a different implementation possible, that is used for other components of this addon (DropDown for example) that need to be aware of their parent<>child relationship. That is using the `component-parent` and `component-child` mixins. I would prefer this one, as it is not using a private API anymore

If you have the time to update your PR, that would be great. If not, no problem, I can do that as well. I would then just pick your tests, which are fine, and add the implementation for the form and form-element components...

Cheers, and thanks for your effort so far!

jelhan:
Refactored with component-parent and component-child mixins. Those are nice ones!

addon/components/bs-form.js
```diff
@@ -154,18 +155,6 @@ export default Ember.Component.extend({
...
 
   /**
-   * An array of `Components.FormElement`s that are children of this form.
-   *
-   * @property childFormElements
-   * @type Array
-   * @readonly
-   * @protected
-   */
-  childFormElements: computed.filter('childViews', function(view) {
```
simonihmig: I would prefer to keep the `childFormElements` in place, but just redefine it to operate on the `children` property instead of on `childViews`. This way we make sure we only operate on FormElement instances, if there is some weird case where we have another child component having the child mixin that is not a FormElement! And maybe some users have subclassed bs-form and used the `childFormElements` property, this way we would not break anything...

jelhan: That makes sence. Updated PR. I guess commits should be squashed before merge...

jelhan:
I have an edge case in ember.js 1.13.13 where parent lookup works in `didInsertElement` hook but not in `willDestroyElement` hook. `this.nearestOfType(ComponentParentMixin);` return `undefined`. Child does not got removed from parent therefore. I have no idea if it's related to this PR or a general bug and what is causing it. @simonihmig do you have any hint?

jelhan:
Bug mentioned in last comment (https://github.com/kaliber5/ember-bootstrap/pull/95#issuecomment-222692509) is not related to this PR. I've opened a separate issue: #99

jelhan:
#99 has an indirect impact. Since `childFormElements` contains destroyed objects in these case `this.get('childFormElements).setEach()` executed on submit throws:

> ember.debug.js:33540 Error: Assertion Failed: calling set on destroyed object

addon/components/bs-form.js
```diff
@@ -232,7 +233,13 @@ export default Ember.Component.extend({
       let validationPromise = this.validate(this.get('model'));
       if (validationPromise && validationPromise instanceof Ember.RSVP.Promise) {
         validationPromise.then((r) => this.sendAction('action', r), (err) => {
-          this.get('childFormElements').setEach('showValidation', true);
+          // childFormElements may contain destoryed objects due to https://github.com/kaliber5/ember-bootstrap/issues/99
+          // calling setEach on destroyed object throws therefore filter out all destroyed objects before
+          // since filterBy returns a base JS Array we have to convert it into Ember.Array
+          const activeChildFormElements = Ember.A(
+            this.get('childFormElements').filterBy('isDestroyed', false)
+          );
```
simonihmig: @jelhan Is this still needed with the fix for #99 in place?

jelhan: I don't think so. Will revert 52a3131c26aaaba845b305362867f682899f100f.

PR #97, opened 2016-05-29
bs-form-element: yield validation state

jelhan:
```hbs
{{#bs-form-element |value id validationState|}}
  <div>{{validationState}}</div>
{{/bs-form-element}}
```

validationState might be 'success', 'error' or null if no validation is to be shown.

My use case is to switch bootstrap button class depending on validation state in a [button addons](http://getbootstrap.com/components/#input-groups-buttons) as suggested [here](https://github.com/twbs/bootstrap/pull/12740).

### DIFF
--- a/app/templates/components/form-element/horizontal/default.hbs
+++ b/app/templates/components/form-element/horizontal/default.hbs
@@ -2,7 +2,7 @@
     <label class="control-label {{horizontalLabelGridClass}}" for="{{formElementId}}">{{label}}</label>
     <div class="{{horizontalInputGridClass}}">
         {{#if hasBlock}}
-            {{yield value formElementId}}
+            {{yield value formElementId validation}}
         {{else}}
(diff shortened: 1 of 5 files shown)


app/templates/components/form-element/horizontal/default.hbs
```diff
@@ -12,7 +12,7 @@
 {{else}}
     <div class="{{horizontalInputGridClass}} {{horizontalInputOffsetGridClass}}">
         {{#if hasBlock}}
-            {{yield value}}
+            {{yield value formElementId validation}}
```
jelhan: Updated PR cause I missed this one.

simonihmig:
Nice! Would you just add that new block param to the documentation at https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form-element.js#L98-L102, please!

jelhan:
Added block param to example but without use case. A good one didn't occurred to me.

simonihmig:
Thanks!

PR #98, opened 2016-05-30
adjust position of feedback items if there is an input group

jelhan:
> Icons, labels, and input groups
> Manual positioning of feedback icons is required for [...] input groups with an add-on on the right. [...] For input groups, adjust the right value to an appropriate pixel value depending on the width of your addon.
> Source: http://getbootstrap.com/css/#forms-control-validation

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -3,7 +3,7 @@ import FormGroup from 'ember-bootstrap/components/bs-form-group';
 import Form from 'ember-bootstrap/components/bs-form';
 import ComponentChild from 'ember-bootstrap/mixins/component-child';
 
-const { computed, defineProperty } = Ember;
+const { computed, defineProperty, observer, on, run } = Ember;
 
 const nonTextFieldControlTypes = Ember.A([
   'checkbox',
@@ -519,5 +519,49 @@ export default FormGroup.extend(ComponentChild, {
       defineProperty(this, 'value', computed.alias(`model.${this.get('property')}`));
       this.setupValidations();
     }
-  }
+  },
+
+  /*
+   * adjust feedback icon position
+   *
+   * Bootstrap documentation:
+   *  Manual positioning of feedback icons is required for [...] input groups
+   *  with an add-on on the right. [...] For input groups, adjust the right
+   *  value to an appropriate pixel value depending on the width of your addon.
+   */
+  adjustFeedbackIcons: on('didInsertElement', observer('hasFeedback', 'formLayout', function() {
+    run.scheduleOnce('afterRender', () => {
+      // validation state icons are only shown if form element has feedback
+      if (this.get('hasFeedback') && !this.get('isDestroying')) {
+        // form group element has
+        this.$()
+          // an input-group
+          .has('.input-group')
+          // an addon or button on right side
+          .has('.input-group input + .input-group-addon, .input-group input + .input-group-btn')
+          // an icon showing validation state
+          .has('.form-control-feedback')
+          .each((i, formGroups) => {
+            // clear existing adjustment
+            this.$('.form-control-feedback').css('right', '');
+            let feedbackIcon = this.$('.form-control-feedback', formGroups);
+            let defaultPositionString = feedbackIcon.css('right');
+            Ember.assert(
+              defaultPositionString.substr(-2) === 'px',
+              '.form-control-feedback css right units other than px are not supported'
+            );
+            let defaultPosition = parseInt(
+              defaultPositionString.substr(0, defaultPositionString.length - 2)
+            );
+            // Bootstrap documentation:
+            //  We do not support multiple add-ons (.input-group-addon or .input-group-btn) on a single side.
+            // therefore we could rely on having only one input-group-addon or input-group-btn
+            let inputGroupWidth = this.$('input + .input-group-addon, input + .input-group-btn', formGroups).outerWidth();
+            let adjustedPosition = defaultPosition + inputGroupWidth;
+
+            feedbackIcon.css('right', adjustedPosition);
+          });
+      }
+    });
+  }))
 });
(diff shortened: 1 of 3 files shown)


jelhan:
parseInt() doesn't seem to fix it for travis / phantomJS 1.9... Will choose another strategy.

Also found another issue: Observer fires on input changes. Therefore already adjusted position is used as default and validation icon moves more and more to the left. Will fix this one two and add another test.

jelhan:
Last commit passes in PhantomJS 2.1, Chrome, Firefox, Opera, Edge and IE 11. Would you mind upgrading PhantomJS?

simonihmig:
Having PhantomJS 2.x on TravisCI does not seem to be that straight forward unfortunately, seems to require manually downloading at setting it up. Will try to add the necessary steps some time soon...

Regarding your PR: I think it should be possible to refactor this a little bit, as to not change the feedback icon's position via jQuery as it is now, but through defining your positioning logic as a computed property, that returns an inline css string, which is added to the template at https://github.com/kaliber5/ember-bootstrap/blob/master/app/templates/components/form-element/feedback-icon.hbs. Something like this:

    <span class="form-control-feedback {{iconName}}" style={{feedbackPosition}} aria-hidden="true"></span>

As an example of what I mean: http://balinterdi.com/2016/02/03/binding-style-attributes-warning-in-ember.html

I think it feels a bit more like "the ember way" (TM), having ember in control of all DOM manipulation (would even work with Fastboot this way).

Otherwise: 👍

jelhan:
Binding a computed property to a style attribute requires `style-src: unsafe-inline` CSP rule isn't it? [ember-ci-csp-style](https://www.npmjs.com/package/ember-cli-csp-style) may address this issue. But I didn't used it yet and we would add one more dependency. Should I refactor to use it?

simonihmig:
Hm, yes I think so. Just by having a quick look at ember-cli-csp-style I could not see how they do things differently without having an issue with CSP rules however...

I would say no to adding that dependency. `bs-collapse` already uses inline styles, so that "issue" is already present, independent of this PR. We maybe can track that issue, but I would go forward with this PR anyway!

jelhan:
You are right. There are quite a lot components requiring CSP `unsafe-inline`. Had just a short test with dummy app. Having strict CSP rules (default ones from ember-cli-content-security-policy) pages accordion, collapse and progress violating CSP.

I don't like to add another CSP violation by this PR. CSP is an issue for me. Already present violations should be removed. Since I'm currently not using the affected components this isn't critical for me. But it would if it's introduced to form components.

Would you be fine with keeping current implementation which modifies DOM elements directly and therefore not causing CSP issues? Otherwise I could investigate how ember-ci-csp-style deal with the issue. I guess it's just about using component attributeBindings for style. But this would require to rewrite feedback-icon as a component (currently it's a partial).

addon/components/bs-form-element.js
```diff
@@ -505,5 +505,49 @@ export default FormGroup.extend({
...
+        this.$()
+          // an input-group
+          .has('.input-group')
+          // an addon or button on right side
+          .has('.input-group input + .input-group-addon, .input-group input + .input-group-btn')
+          // an icon showing validation state
+          .has('.form-control-feedback')
+          .each((i, formGroups) => {
+            // clear existing adjustment
+            this.$('.form-control-feedback').attr('style', '');
```
jelhan: CSP is a hard case. This line violates CSP if `style-src` does not allow `unsafe-inline`.

simonihmig:
Sorry, has been a little quiet from my side... 

I just can't figure out under which circumstances the CSP style-src rule applies. Having inline styles at all does not seem to be the issue, as something like `$(...).css(...)` as you have used it also adds an `style` attribute to the Dom node. And `bs-collapse` uses `attributeBindings` for style, but as it seems still triggers an error... 

I don't know, maybe you can enlighten me?

I would really prefer a consistent solution that is not using jQuery or direct DOM manipulation. Solutions like would not work as expected in a Fastboot environment for example, which I try to support...

jelhan:
I'm not an expert for CSP but I will have a try.

<style> element and HTML style attributes are only allowed by `unsafe-inline`. This also counts for style attributes added by JavaScript. Therefore `this.$('.form-control-feedback').attr('style', '');` violates a restrictive CSP style-src policy. But updating style properties directly via CSS Object Model (CSSOM) does not violate it. This is was jQuery css method does:

> When using .css() as a setter, jQuery modifies the element's style property. For example,  `$( "#mydiv" ).css( "color", "green" )` is equivalent to `document.getElementById( "mydiv" ).style.color = "green"`.

I played a little bit around with ember-cli-csp-style and noticed that master is broken. I was wrong by guessing `attributeBindings` does the trick. It violates CSP.. In last stable release (which still seem to work with Ember 2.6) it manipulates CSSOM directly: https://github.com/BryanCrotaz/ember-cli-csp-style/blob/v1.13/addon/mixins/csp-style.js#L11, https://github.com/BryanCrotaz/ember-cli-csp-style/blob/v1.13/addon/mixins/csp-style.js#L51-L59

I'm not quite sure if ember provides any way to avoid CSP violation when binding style attribute. I asked for a recommendation at [discuss.ember.js](http://discuss.emberjs.com/t/binding-style-attribute-and-content-security-policy-best-practice/10921).

simonihmig:
Great, thanks for your explanation, I think I got it! ;)

Probably there is no easy solution immediately applicable, so I'm fine with leaving it the way it is now until we find a better way...

There is just the broken PhantomJS 1.9 test still standing in the way to getting this merged. I see that there is just a minor 1px difference between the expected and the actual positioning. Maybe this could be "fixed" by asserting it with a small tolerance, like +- 2px?

jelhan:
Fixed travis/PhatomJS 1.9. I decided for a more restrictive way to avoid false positives.

Also squashed commits together and rebased against master since #103 introduced a merge issue. Would love to see this one merged and discuss how to handle style attribute and CSP after some feedback at http://discuss.emberjs.com/t/binding-style-attribute-and-content-security-policy-best-practice/10921.

simonihmig:
Merged. 

Vielen Dank für die ganzen "contributions"! 😄  👍

jelhan:
Vielen Dank für das großartige Addon! :+1:

PR #101, opened 2016-06-03
import map file along with bootstrap-theme.css

williamherry:
otherwise chrome will complaint about failed to parse srouce map


simonihmig:
Thanks!

PR #103, opened 2016-06-07
adds support for sr-only label

jelhan:
Introduces an option to add an label but hide it except for screen readers.

Motivation:

> Always add labels
> Screen readers will have trouble with your forms if you don't include a label for every input. For these inline forms, you can hide the labels using the .sr-only class. There are further alternative methods of providing a label for assistive technologies, such as the aria-label, aria-labelledby or title attribute. If none of these is present, screen readers may resort to using the placeholder attribute, if present, but note that use of placeholder as a replacement for other labelling methods is not advised.
> http://getbootstrap.com/css/#forms-inline

Not quite sure about the naming. Perhaps `srOnlyLabel` would me better?

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -114,11 +114,24 @@ export default FormGroup.extend(ComponentChild, {
   /**
    * Text to display within a `<label>` tag.
    *
+   * You should include a label for every form input cause otherwise screen readers
+   * will have trouble with your forms. Use `invisibleLabel` property if you want
+   * to hide them.
+   *
    * @property label
    * @type string
    * @public
    */
   label: null,
+
+  /**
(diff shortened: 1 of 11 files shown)


simonihmig:
What about something like `invisibleLabel`, with an explanation of what it does (hiding it just visually, but making it accessible for screen readers) in the js docs? I feel that would be more intuitive...

Also, maybe you call me pedantic, it's just a code style thing: would you please use double quotes in the template files! ;) `{{if labelSrOnly "sr-only"}}`

jelhan:
`invisibleLabel` sounds good. Coding style matters. :-) Updated PR.

simonihmig:
I see no update!? Forgotten to push? :)

jelhan:
Ups, even forgotten to commit the changes. :-)

PR #105, opened 2016-06-23
allow value 0 (Number) as valid value for a bs-button

arohr:
An example like this works now:

    {{#bs-button-group value=opmode type="radio"}}
      {{#bs-button value=0}}AUTO{{/bs-button}}
      {{#bs-button value=1}}REMOTE{{/bs-button}}
      {{#bs-button value=2}}LOCAL{{/bs-button}}
    {{/bs-button-group}}


simonihmig:
Thanks!

PR #108, opened 2016-06-29
Add 'expanded' class name for accordion item

Charizard:
Also, add expanded class if the accordion is expanded


simonihmig:
Could you just change the quotation marks to double quotes for the purpose of consistent code style, please?

Charizard:
Actually, that is the fix 😕 

`class="panel-heading {{if collapsed "collapsed"}}"` this won't work as class is also surrounded by double quotes.

simonihmig:
Sorry for the delay, this is merged now. Thanks!

PR #116, opened 2016-07-18
useIcons default should be false if a custom control is used

jelhan:
`controlType` defaults to `text` even if a custom control is used. Therefore `useIcons` was `true` for custom control types as default. Referring docs that wasn't intended.

`!nonTextFieldControlTypes.contains(this.get('controlType'));` had returned `false` for all controlTypes except `text`. Refactored to less complex code.

This should be considered a breaking change.

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -427,7 +427,8 @@ export default FormGroup.extend(ComponentChild, {
    * @public
    */
   useIcons: computed('controlType', function() {
-    return !nonTextFieldControlTypes.contains(this.get('controlType'));
+    return !this.get('hasBlock') &&
+           !nonTextFieldControlTypes.contains(this.get('controlType'));
   }),
 
   /**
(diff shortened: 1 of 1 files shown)


jelhan:
Sorry, I forgot that `controlType` supports inputTypes like `email`, `password` etc. which renders as `input` elements. `useIcons` should be `true` for them also. Will remove my refactor from PR.

`useIcons` doesn't seem to be covered by tests very well since they all passed.

jelhan:
Due to #117 manual testing is not such easy.

jelhan:
It's not my day:

> hasBlock is only available in template contexts (its implemented as a template layer keyword), and is not intended to be available in the components JS code.
> https://github.com/emberjs/ember.js/issues/11741#issuecomment-121248949

Therefore proposed fix is **not working**.

@simonihmig Before putting even more time into this one. Are you willing to accept a PR with a change like suggested? Otherwise we could just add a note to documentation that useIcons should be set to `false` if a custom control isn't rendered as input element.

simonihmig:
@jelhan I think it would be useful `useIcons`to be false if it is a custom control, at least by default. But I am also not so sure about how to implement that, taking the `hasBlock` issue into account...

jelhan:
We could combine the `hasBlock` template helper with a check if `useIcons` has been passed to component to support override. [Here](https://ember-twiddle.com/2823cd9aaa0ecf80ae6ff1350020eecb?fileTreeShown=false&numColumns=2&openFiles=components.my-component.js%2Ctemplates.application.hbs) is a Ember Twiddle to demonstrate the approach.

jelhan:
I guess a work-a-round like the one mentioned in https://github.com/kaliber5/ember-bootstrap/pull/116#issuecomment-234584657 isn't the way to go. Closing this one in favor of doc update in #118

PR #118, opened 2016-07-27
doc(FormElement): useIcons and custom controls

jelhan:


simonihmig:
That solution is ok for me, feeling sorry for you for the lost time with the other PR... ;) Thanks anyways! 👍

PR #121, opened 2016-08-03
Perform cleanup with modal willRemoveElement

sukima:
The removal of the `modal-open` class is handled in the `hideModal` function which also performs proper math to remove padding that was previously added when the modal opened. Since this happens when `hide()` is called (via the `open` property observer) everything works. But if you programmatically remove the modal via ember *without* using the `open` property but using `{{#if …}}` like the docs impart the cleanup is not performed.

Unfortunately the `willRemoveElement` is the only hook and it does not perform the same cleanup that `hide` does and the result is that the scroll bar padding is not reset.

The result was that closing the modal via Ember from the outside of the component caused the right side of the page to squish (`padding-right`) and be compounded each time a modal closes from ember. This was noticed in Firefox but Chrome was not affected (I have no idea why).

An example to reproduce this:

Template
--------

```handlebars
{{#if showModal}}
  {{#bs-modal title="Simple Dialog"}}
    Hello world!
  {{/bs-modal}}
{{/if}}
```

Component
---------

```javascript
import Ember from 'ember';

export default Ember.Component.extend({
  showModal: false,

  infinateModalLoop: Ember.on('didInsertElement', function () {
    this.toggleProperty('showModal');
    Ember.run.later(this, 'infinateModalLoop', 200);
  })
});
```

### DIFF
--- a/tests/integration/components/bs-modal-test.js
+++ b/tests/integration/components/bs-modal-test.js
@@ -502,3 +502,38 @@ test('Renders in place (no wormhole) if renderInPlace is set', function(assert) 
   assert.equal(this.$('.modal').length, 1, 'Modal exists.');
   assert.notEqual(this.$('.modal').parent().attr('id'), 'ember-bootstrap-modal-container');
 });
+
+test('Removes "modal-open" class when component is removed from view', function(assert) {
+  this.set('renderComponent', true);
+  this.render(hbs`{{#if renderComponent}}{{#bs-modal title="Simple Dialog"}}Hello world!{{/bs-modal}}{{/if}}<div id="ember-bootstrap-modal-container"></div>`);
+
+  let done = assert.async();
+
+  // wait for fade animation
+  setTimeout(() => {
+    assert.ok($('body').hasClass('modal-open'), 'body element has "modal-open" class.');
+
+    this.set('renderComponent', false);
+
+    assert.ok(!($('body').hasClass('modal-open')), 'body element does not have "modal-open" class.');
+    done();
+  }, transitionTimeout);
+});
+
+test('Resets scroll bar when component is removed from view', function(assert) {
+  document.body.style.paddingRight = '50px';
+  this.set('renderComponent', true);
+  this.render(hbs`{{#if renderComponent}}{{#bs-modal title="Simple Dialog"}}Hello world!{{/bs-modal}}{{/if}}<div id="ember-bootstrap-modal-container"></div>`);
+
+  let done = assert.async();
+
+  // wait for fade animation
+  setTimeout(() => {
+    document.body.style.paddingRight = '0px';
+    this.set('renderComponent', false);
+
+    assert.equal(document.body.style.paddingRight, '50px', 'paddingRight restored to 50px');
+    document.body.style.paddingRight = '0px';
+    done();
+  }, transitionTimeout);
+});
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -690,6 +690,7 @@ export default Ember.Component.extend({
   willDestroyElement() {
     Ember.$(window).off('resize.bs.modal');
     Ember.$('body').removeClass('modal-open');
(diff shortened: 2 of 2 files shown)


sukima:
How did test explode? I changed **one** line!!!

sukima:
Please Help

simonihmig:
`hideModal` does so much more than just reseting the scrollbar adjustments, it fades out the background element in an asynchronous way (fade animation). That does not make any sense to do in the `willDestroyElement` hook, as the component will be non-exsiting when that asynchrounous code is executed. The test failures are probably caused by this.

I think just fixing the `model-open` class and calling `resetScrollbar` should probably do it. 

However I would like to have a test covering this case. If you are willing to continue working on this and add a proper test, that would be great! If not, then please just file an issue for this, and I will try to tackle this!

Thanks for your support in either case! :)

sukima:
Thank you! I'll probably look into it tomorrow.

sukima:
@simonihmig How about now? Let me know if you need more fixups.

sukima:
Damn it WTF! this worked on my machine.

sukima:
I don't think it is possible for me to get these tests passing. I see errors where EmberENV is missing! Is Ember-canary not compatible with Travis? And Ember-beta seems to die on every `afterEach` I'm at a loss.

@alexwilkinson: would you happen to have any insight?

simonihmig:
@sukima Yes, there were a couple of errors in Travis, canary was crashing PhantomJS 1.9. The other PR #123 had the same issues. I fixed them over the weekend by upgrading to ember-cli 2.7.0 / PhantomJS 2.x. The changes are in `master`. You would have to rebase onto it, hopefully that should fix them!

sukima:
**Yes** Tests pass!

Doggie dance

tests/integration/components/bs-modal-test.js
```diff
@@ -502,3 +502,44 @@ test('Renders in place (no wormhole) if renderInPlace is set', function(assert)
...
+  let done = assert.async();
+
+  // wait for fade animation
+  setTimeout(() => {
+    assert.ok($('body').hasClass('modal-open'), 'body element has "modal-open" class.');
+
+    this.set('renderComponent', false);
+
+    // wait for fade animation
+    setTimeout(() => {
```
simonihmig: The second `setTimeout` in both tests should be omitted. Make the assertion directly after setting `renderComponent` to false. 

We do not have a problem when the modal is closed programmatically, by setting `open` to false, in which case the fade out transition is made. Only when it is "hard" removed as you do it within your template. In this case the fade out transition is never started, so let's test the correct cleanup immediately!

Apart from that this looks good! Thanks!!!

sukima: Oh. Interesting. The docs would benefit explaining this distinction. Often I use the `{{#if}}` method because it make more cognitive sense especially if the modal will be re displayed again.

One of the odd things I run into with `open` is a syncing between setting that to false and the user clicking the backdrop and the model is closed. If I'm not careful to adjust the property that `open` is bound to then it won't observe it being set back to `true` later.

However, if I wrap the modal in `{{#if}}` then it makes more sense, to me anyway, to remember to add a `closedAction=(action (mut theOpenBoundProperty) false)`.

Anyway I'll fix the tests.

sukima: Speaking of which, I wonder if it would be worth animating the fade in the `willRemoveElement` hook. But I presume that there is problems with this.

simonihmig: No chance. Ember will remove the component from DOM immediately, there is nothing left to be animated! ;) That's why a model within an `{{#if ...}}` block won't have a chance to fade out, unless you trigger the fade out with setting `open` to false and set the if-condition to false in the `closed` action (that is called when the transition has finished) to fully remove it from DOM...

simonihmig:
Merged. Thanks again!

Issue #48, opened 2016-02-09
navbar example


joepadmiraal:
Can you please add a navbar example with the collapse component to the documentation?

Thanks,
Joep

zoamel:
+1 For that

plainsimple123:
Wow! What a coincidence I'm also looking for that. +1.

plainsimple123:
Well, after relentless searching I found the example in ember-bootstrap/tests/dummy/app/templates/collapse.hbs. Enjoy @joepadmiraal and @zoamel!

It would be great if that is shown in the published documentation. But nonetheless, great job to kaliber5!

joepadmiraal:
@plainsimple123 Thanks for pointing to that example, but that is not exactly what I am looking for.
I'm looking for an example which includes the navbar.

I currently have something like this, but it is not working correctly.
For example when the navbar is closed in a small window and the window is made larger, the navbar stays invisible.

```
<div class="navbar-header">
  {{#bs-button class="navbar-toggle" toggle=true active=notCollapsed action="toggle"}}
    <span class="sr-only">Toggle navigation</span>
    <span class="icon-bar"></span>
    <span class="icon-bar"></span>
    <span class="icon-bar"></span>
  {{/bs-button}}
  <span class="navbar-brand active">Ember.js example</span>
</div>

{{#bs-collapse collapsed=collapsed}}
  <nav class="navbar-collapse bs-navbar-collapse" role="navigation">
    <ul class="nav navbar-nav">
        {{#active-link}} {{#link-to 'groups'}}Groups{{/link-to}} {{/active-link}}
        {{#active-link}} {{#link-to 'students'}}Students{{/link-to}} {{/active-link}}
    </ul>
  </nav>
{{/bs-collapse}}
```

simonihmig:
@plainsimple123 I plan to integrate the demo app from tests/dummy into the official documentation page (http://kaliber5.github.io/ember-bootstrap/), hopefully soon.

simonihmig:
@joepadmiraal Can you please show me a code example of what you want to achieve using the original bootstrap.js? I am not sure I unterstand what your goal is...

joepadmiraal:
@simonihmig Basicaly its just [this navbar example](http://getbootstrap.com/components/#navbar-default) from the official Bootstrap site.

When the browser window is made smaller the navbar will hide the buttons and show a button to collapse and expand the button list.

kwasielew:
[...]

Issue #86, opened 2016-05-06
Add support for contextual components


simonihmig:
Useful for Form > FormElement and Modal > ModalHeader/Body/Footer

srvance:
Also useful for Navbar and supporting components.

simonihmig:
Closed in favor of #158

PR #123, opened 2016-08-07
Adds navbar support

srvance:
This addresses issue #48. It adds a family of bs-navbar* components usable as contextual components under bs-navbar. I have written full documentation, although I have not run the generation from them. I have written integration tests for everything. The acceptance test is not much different from what was generated because I haven't looked into how to trigger responsive behaviors yet. The use of {{hash}} introduces some backward compatibility issues. I believe it was introduced in Ember 2.3 and can be polyfilled back, but I don't know if the same is true of contextual components.


simonihmig:
Thanks @srvance for that large contribution! I will probably need some time for reviewing this, as I am pretty busy right now...

I also have to think about the contextual component thing. I did have this on my roadmap, see https://github.com/kaliber5/ember-bootstrap/issues/86. However ember-bootstrap currently is supporting all ember versions starting with 1.13, so this needs to be addressed somehow.

Oh, and I noticed there were some failing tests in Travis!

srvance:
Glad to contribute! It was fun and I wanted it for myself. :)

Navbar failed in travis prior to Ember 2.3 for the contextual component reasons you'd expect. We can unroll the contextual component work for the time being if that's an obstacle. It exposes a bit more of the plumbing to the user because we can't do the component currying and requires an extra yield action, but it's not too bad. We could also document how to override the templates for people who want the contextual form.

Most of the failures are in beta and canary for large parts of the project. They only affect the navbar in one test for bs-navbar-toggle that probably relates to the button failures; it extends. There are also some failures in bs-progress-bar in the default and 2.3 scenarios.

tests/integration/components/bs-navbar-test.js
```diff
@@ -0,0 +1,83 @@
...
+});
+
+test('it handles inverse navbars properly', function(assert) {
+  this.render(hbs`{{bs-navbar type="inverse"}}`);
+
+  assert.ok(this.$('nav').hasClass('navbar-inverse'), 'the navbar has the navbar-inverse class');
+  assert.notOk(this.$('nav').hasClass('navbar-default'), 'the navbar does not have the navbar-default class');
+});
+
+test('it handles the toggling action propoerly', function(assert) {
```
simonihmig: This test seems to be broken. First there is no `#clickme` element in your rendered template. This is not catched in the test probably because of the second error: your use of `assert.async` is wrong. It returns a callback that you have to call when your async stuff has finished. See https://api.qunitjs.com/async/ and https://github.com/kaliber5/ember-bootstrap/blob/fa66b53f862d2c893c06872fbe31c4a88d0cf567/tests/integration/components/bs-collapse-test.js#L31-L39 for examples.

This is probably also the cause of some other tests (progress bar) to fail, because your assertions here are run outside of the context of this test, as the test does not wait for your timeouts...

tests/acceptance/bs-navbar-test.js
```diff
@@ -0,0 +1,13 @@
...
+// TODO: Write this
+moduleForAcceptance('Acceptance | bs navbar');
+
+test('visiting /navbars', function(assert) {
+  visit('/navbars');
+
+  andThen(function() {
+    assert.equal(currentURL(), '/navbars');
+  });
+});
```
simonihmig: I don't think we need an acceptance test here. It should all be covered by integration tests

addon/components/bs-navbar.js
```diff
@@ -0,0 +1,87 @@
...
+ {{/bs-navbar}}
+ ```
+
+ ### Navbar styles
+
+ The component supports the default bootstrap navbar styling options "default" and "inverse" through the `type`
+ property. Bootstrap navbars [do not currently support justified nav links](http://getbootstrap.com/components/#navbar-default),
+ so those are explicitly disallowed.
+
+ Other bootstrap navbar variations, such as forms, buttons, fixed to top, etc. can be supported through direct use of
```
simonihmig: I think we should support the styling options of the `bs-navbar` itself through properties, rather than setting class names. Something like a `position` property maybe?

addon/templates/components/bs-navbar.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="container-fluid">
```
simonihmig: Hm, this does not have to be fluid, you could also want to have a `.container` container! Maybe introduce a `fluid` boolean property?

simonihmig:
Some more general notes:

1. I think dropping support for ember < 2.3 at this point does not make sense, since the value added through the use of contextual components is not that big. I had some similar cases with other components where you would want the component to call an action on its logical "parent" component, and not on the component/controller of its default action context. See https://github.com/kaliber5/ember-bootstrap/blob/master/addon/mixins/modal-closer.js for an example where the action target object is "hardwired" to a parent component, using the `nearestOfType` method. One could apply this to the toggle button, without having the user to do the "wiring"...

2. I am not sure we actually need those components like `bs-navbar-header` or `bs-navbar-brand`, i.e. all components that actually do not add any functionality besides their default markup (class name), as writing `{{#bs-navbar-header}...{{/bs-navbar-header}}` does not make thinks easier that just writing `<div class="navbar-header">...</div>`. It might even add some performance and/or memory footprint overhead. That's why I did not try to add all those things like grids, badges, labels and so on, that you could implement just as well using plain HTML markup. So this would be a little bit inconsistent here. 
For the `bs-navbar-brand` component in particular: the current implementation does not allow linking to internal routes, or would force a complete app reload. One would probably want to use the `link-to` component instead, to link to e.g. the `index` route and perform an internal route transition: `{{#link-to "index" class="navbar-brand"}}...{{/link-to}}`

tests/integration/components/bs-navbar-test.js
srvance: Makes sense. I'm getting re-acquainted with qunit. I used collapse as my basis, but erroneously decided I could inline `done`. Fixing.

tests/acceptance/bs-navbar-test.js
srvance: I agree. The navbar integration test covers it pretty well. Will remove.

srvance: We should reconsider this if we decide to back off on the contextual components. The navbar test won't be nearly as comprehensive on the multi-component behaviors in that case.

srvance:
1. I'll take a look at that. Did you see the previous version that exposed some of the plumbing? Will still need to expose the `collapsed` property, I think.

2. I asked myself the same thing and mainly added them for the consistency. I buy the keeping it minimal and overhead arguments and will remove them.

addon/components/bs-navbar.js
srvance: Done.

addon/templates/components/bs-navbar.hbs
srvance: Done.

tests/acceptance/bs-navbar-test.js
simonihmig: You can test components with sub components as well in an integration test, see https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-accordion-test.js#L44-L59 as an example.

simonihmig:
Thanks for the instand changes, that's great! Tests do run fine now for ember > 2.3, now let's tackle the other versions...

I think one could create a binding for the `collaped` property without the user having to manage this. See an example here: `bs-form-element` has a `form` property that is a reference to its "parent" component (in this case a form): https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form-element.js#L528-L536. And it then "reads" a property of its parent: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form-element.js#L294

This could work here in a similar way: the `bs-navbar-content` component creating a reference to its parent `bs-navbar` and reading its `collapsed` property. What do you think?

srvance:
Not done yet. Just had to break for the day. Should finish this up today.

I'm fine with this technique until we can do contextual components. It strikes me as a clever necessity rather than a desirable solution from a design aesthetics perspective. I'll see how things look as I unroll the contextual components to decide whether to use that technique for `collapsed`.

srvance:
BTW, what's your opinion on how I implemented the container[-fluid] change? Would you prefer a `classNameBindings` approach with a computed property instead?

srvance:
I ended up directly wiring navbar-content to the navbar using that technique, but still yielded `collapsed` for testing purposes. Is there another way to get at component state from within the test? See https://github.com/srvance/ember-bootstrap/blob/f1ceaa4ee7fc65e9527b024aa35b694301199bf8/tests/integration/components/bs-navbar-test.js#L47 for where it's relevant.

srvance:
All set! Let me know if you think we need to do more.

simonihmig:
@srvance Looks good, with the exception of yielding `collapsed` as you mentioned. I would write the test in a way that it tests if the `bs-navbar-content` is actually collapsed, that is it has/has not the `in` class. Just as in the `bs-collapse` test!

srvance:
@simonihmig I considered that, but was thinking I'd rather test the behavior added in `bs-navbar` itself rather than consequential behavior. The `yield` doesn't hurt anything and doesn't have to be acknowledge or used in normal usage. Do you mind if we keep it that way? I'm worried that the other approach couples the `bs-navbar` test to `bs-navbar-content` behavior.

simonihmig:
Hm, I would probably not really mind it, but I would still prefer the other way! ;) After all these are integration tests, and not unit tests, so complete test isolation is not an objective here, IMHO. 

a) the different navbar components are connected to each other and not meant to be used separately, so testing if they are "cooperating" in the expected way seems fine to me. 
b) testing for the `in` class means testing for the actual behavior of the component(s), that is it unveils some content, rather than testing for some internal implementation detail (the `collapsed` property) 

At least that is my personal view of it... :)

srvance:
Your reasoning is sound. I have a slight preference for the way it is, but could go either way. I'd feel more strongly if it were a unit test. The change is on its way.

srvance:
Looks like Phantom died at the start of canary due to a missing variable. Any way we can kick it off again?

simonihmig:
@srvance Did a forced rebuild twice in TravisCI, both died as well, at the exact same step (canary). No idea what that means... 

Could you please just make another dummy commit, maybe a forced rebuild is not cleaning things up in the right way...

srvance:
@simonihmig Done. I found a documentation mistake to correct.

srvance:
@simonihmig Looks like same thing again. :(

srvance:
@simonihmig It looks like somehow require.js isn't present.

simonihmig:
It seems there is an issue with ember canary, that is totally unrelated to TravisCI as well as your PR. The same happens on `master` when running `ember try:each` locally. Will try to fix this...

simonihmig:
@srvance I have fixed the issue now with an upgrade to ember-cli 2.7.0 and PhantomJS 2 on TravisCI. Could you please rebase your branch to master. That should fix the problems in this PR as well, so we can get this finally merged!

srvance:
The ember-cli 2.7.0 upgrade caused some issues that were all over the Slack channel yesterday. Fixed here.

simonihmig:
Merged this. Will release this as 0.9.0 asap. Huge thanks for your effort!!!

PR #127, opened 2016-08-18
feat(FormElement): adds `showValidationOn` events

jelhan:
E.g. a checkbox or a radio group should show validation immediately after change. Therefore make it configurable trough `showValidationOn` property.

Also fixes existing tests for validation.

### DIFF
--- a/tests/integration/components/bs-form-element-test.js
+++ b/tests/integration/components/bs-form-element-test.js
@@ -268,43 +268,108 @@ test('adjusts validation icon position if there is an input group', function(ass
 });
 
 test('shows validation errors', function(assert) {
-  this.set('errors', Ember.A());
+  this.set('errors', Ember.A(['Invalid']));
+  this.set('model', Ember.Object.create({ name: null }));
   this.render(hbs`
-      {{bs-form-element property='name' elementId='child' showValidation=true hasValidator=true errors=errors}}
+      {{bs-form-element property='name' elementId='child' hasValidator=true errors=errors model=model}}
   `);
-  assert.ok(
-    this.$('form .has-error').length === 0,
+  assert.notOk(
+    this.$('.form-group').hasClass('has-error'),
     'validation errors aren\'t shown before user interaction'
   );
   Ember.run(() => {
-    this.set('errors', Ember.A([
-      'Invalid'
-    ]));
+    this.$('input').trigger('focusout');
   });
   assert.ok(
     this.$('.form-group').hasClass('has-error'),
-    'validation errors are shown when errors are present (child)'
+    'validation errors are shown after user interaction when errors are present (child)'
   );
   assert.equal(this.$('.form-group .help-block').text().trim(), 'Invalid');
+  Ember.run(() => {
+    this.set('errors', Ember.A());
+  });
+  assert.notOk(
+    this.$('.form-group').hasClass('has-error'),
+    'form group isn\'t shown as having errors if there aren\'t any'
+  );
 });
 
 test('shows validation warnings', function(assert) {
-  this.set('warnings', Ember.A());
+  this.set('warnings', Ember.A(['Insecure']));
+  this.set('model', Ember.Object.create({ name: null }));
   this.render(hbs`
-      {{bs-form-element property='name' elementId='child' showValidation=true hasValidator=true warnings=warnings}}
+      {{bs-form-element property='name' elementId='child' hasValidator=true warnings=warnings model=model}}
   `);
-  assert.ok(
-    this.$('form .has-warning').length === 0,
+  assert.notOk(
+    this.$('.form-group').hasClass('has-warning'),
     'validation warnings aren\'t shown before user interaction'
   );
   Ember.run(() => {
-    this.set('warnings', Ember.A([
-      'Insecure'
-    ]));
+    this.$('input').trigger('focusout');
   });
   assert.ok(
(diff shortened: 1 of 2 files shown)


jelhan:
Would love to discuss the api cause passing an array isn't easily possible with default template helpers. Of course if you have an array template helper current syntax is nice:

```hbs
  {{bs-form-element showValidationOn=(array 'change' 'focusOut')}}
```

We could use a plain object cause `hash` helper is part of core. But this would require to pass a value. Since it should be default to false for absent keys the only useful value would be `true`. This would make syntax ugly I guess.

```hbs
  {{bs-form-element showValidationOn=(hash change=true focusOut=true)}}
```

So what do you think? Should we suggest using an array template helper or switch `showValidationOn` to a plain object?

Another possibility would be to also support a single event passed as `String` beside an `Array`:

```hbs
  {{bs-form-element showValidationOn='change'}}
  {{bs-form-element showValidationOn=(array 'change' 'focusOut')}}
```

This would reduce the need of an array helper to use cases where you have to support multiple events but would increase code complexity.

srvance:
The `hash` helper wasn't introduced until Ember 2.3 and right now ember-bootstrap is compatible back to 1.13.

What do you think of comma-separated list syntax (e.g., `showValidationOn='change,focusOut'`) and parsing it?

jelhan:
@srvance Api design discussed here will not change compatibility. You could always use a property or a template helper which is not part of core.

I don't like comma-separated lists. If it's a list it should be defined as one. But perhaps that's just a coding style question.

Personally I'm tending towards an `Array`. I think it's intuitive to handle a list of values as an Array. I bet an `array` template helper will make it to core sometime (or `hash` will be removed again). Note that [ember-array-helper](https://github.com/kellyselden/ember-array-helper) already exists.

srvance:
There's ember-hash-helper-polyfill, too. If the API requires it, should ember-bootstrap add it as a dependency?

jelhan:
It's not required. Test added by this PR demonstrate how a property could simply be used. Only if you like to define the value in template an array template helper (or hash template helper if we choose plain object) is required.

jelhan:
Also note that there are already some components having an array property: `value` in `ButtonGroup` and `choices` in `FormElement`.

tests/integration/components/bs-form-element-test.js
```diff
@@ -268,43 +268,82 @@ test('adjusts validation icon position if there is an input group', function(ass
...
+    'form group isn\'t shown as having warnings if there are\'t any'
+  );
+});
+
+test('events enabling validation rendering are configurable per `showValidationOn`', function(assert) {
+  this.set('errors', Ember.A(['Invalid']));
+  this.set('model', Ember.Object.create({ name: null }));
+  this.set('showValidationOn', ['change']);
+  this.render(hbs`
+      {{bs-form-element property='name' elementId='child' hasValidator=true errors=errors model=model}}
```
simonihmig: `showValidationOn= showValidationOn` is missing!

tests/integration/components/bs-form-element-test.js
```diff
@@ -268,43 +268,82 @@ test('adjusts validation icon position if there is an input group', function(ass
...
+  `);
+  assert.notOk(
+    this.$('form-group').hasClass('has-warning'),
```
simonihmig: you are testing for the `has-warning` which will only be present if the `warnings` property is set. You have `errors` assigned, so check for `has-error` class!

tests/integration/components/bs-form-element-test.js
```diff
@@ -268,43 +268,82 @@ test('adjusts validation icon position if there is an input group', function(ass
...
+    this.$('input').trigger('focusout');
+  });
+  assert.notOk(
+    this.$('form-group').hasClass('has-warning'),
+    'events not present in `showValidationOn` are ignored'
+  );
+  Ember.run(() => {
+    this.$('input').trigger('change');
+  });
+  assert.notOk(
```
simonihmig: This should be `assert.ok`. That's probably why the test succeeds although you have missed assigning `showValidationOn` (see above)! ;)

tests/integration/components/bs-form-element-test.js
```diff
@@ -268,43 +268,82 @@ test('adjusts validation icon position if there is an input group', function(ass
...
+    this.$('form-group').hasClass('has-warning'),
+    'events present in `showValidationOn` trigger validatoin'
```
simonihmig: typo

simonihmig:
I am in favor of @jelhan's last suggestion: accept an array as well as a string. 

Arrays (or array-like objects) are certainly acceptable, every component that iterates over a collection to e.g. display a list of things expects an array. In this case it is certainly a bit strange to have to add an property to your controller/parent component to supply the events array, as this array will be static and not dynamic data. But I think this case is quite exotic when you accept a string, as you probably will want to watch for either `focusOut` or `change`, not both of them.

srvance:
I agree with 'as well as'. I read it as 'instead of', probably erroneously.

jelhan:
Thanks for code review! I guess it was really to late that night...

Fixed the tests and added `string` support. The api now supports both `String` and `Array`:

```hbs 
{{bs-form-element validationOn='change'}}
{{bs-form-element validationOn=(array 'change' 'focusOut')}}
```

simonihmig:
Thanks @jelhan! Published as 0.9.1

PR #143, opened 2016-10-12
.idea directory shouldn't be part of node package

salmanasiddiqui:


simonihmig:
This is strange, `.idea` is part of `.gitignore` and as such should also be ignored by npm. Maybe it's because of the missing slash? https://github.com/kaliber5/ember-bootstrap/blob/master/.gitignore#L20

Though git seems to accept this...

simonihmig:
Thanks anyways!

salmanasiddiqui:
@simonihmig 
https://github.com/npm/npm/issues/3763#issuecomment-22255799

`npm` respects `.gitignore` only if there is no `.npmignore`

Welcome :wink:

simonihmig:
Oh, did't know that, thanks for teaching me this! ;)

Issue #140, opened 2016-10-06
Add tree shaking support


simonihmig:
Add options to `ember-cli-build.js` to include/exclude certain components in the build. 

Example: https://github.com/DockYard/ember-composable-helpers/blob/master/index.js

PR #147, opened 2016-11-03
Add tree shaking support #140

huberts:
Implementation based on DockYard/ember-composable-helpers, resolving issue #140 

It uses the same configuration method as [the EmberComposableHelpers pointed by you](https://github.com/DockYard/ember-composable-helpers#configuration):

```
//your-bootstrap-app/ember-cli-build.js

/* global require, module */
var EmberApp = require('ember-cli/lib/broccoli/ember-app');

module.exports = function(defaults) {
    var app = new EmberApp({
        'ember-bootstrap': {
          only: ['bs-button', 'bs-form'],
          except: ['bs-accordion']
        }
    });

    return app.toTree();
};

```

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,9 @@ var path = require('path'),
   util = require('util'),
   extend = util._extend,
   mergeTrees = require('broccoli-merge-trees'),
-  Funnel = require('broccoli-funnel');
+  Funnel = require('broccoli-funnel'),
+  intersection = require('./lib/intersection'),
+  difference = require('./lib/difference');
 
 var defaultOptions = {
   importBootstrapTheme: false,
@@ -46,6 +48,9 @@ module.exports = {
       app.import(path.join(bootstrapPath, 'fonts/glyphicons-halflings-regular.woff2'), {destDir: '/fonts'});
     }
 
+    this.whitelist = this.generateWhitelist(options);
+    this.blacklist = this.generateBlacklist(options);
+
     if (!process.env.EMBER_CLI_FASTBOOT) {
       app.import('vendor/transition.js');
     }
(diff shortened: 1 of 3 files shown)


simonihmig:
Cool, looks good so far!

As there is no automatic test for this feature (would be doable but a bit more complicated), have you actually tested this in a real app, does it really exclude all the source code of the excluded component?

huberts:
You can check it right here, in [sandbox app](https://github.com/huberts/ember-bootstrap-sandbox). It makes vendor.js file 100kb lighter.
The downside is, I had to explicitly pass to 'ONLY' section not only components used in code (bs-button, bs-form, bs-navbar, bs-navbar-toggle, bs-navbar-toggle, bs-navbar-nav, bs-navbar-item), but also the ones that are dependencies for them (bs-form-element, bs-form-group, bs-collapse).
What do you think? Maybe it would be nice to NOT explicitly list these dependencies?

simonihmig:
Yeah, good point. Would be cool to have that, would probably require to setup a little dependency tree, something like:

```js
var dependencies = {
  'bs-dropdown': ['bs-collapse', ...],
  ...
};
```

The `generateWhitelist` method would then have to check those dependencies (recursively) and add all those to the whitelist.

If you are willing to contribute that, that would be awesome, but in this case I would see tests for this as a requirement (there were already some node tests for the composable helpers addon here: https://github.com/DockYard/ember-composable-helpers/tree/master/test).

But I am also fine with getting this merged as it is now, and maybe adding the dependency feature later.

huberts:
Well, give me some time and I will upgrade the solution with dependencies. And with tests as well.

simonihmig:
@huberts may I ping you on this one again. A lot has changed here in master since the initial PR was created, so there are some merge conflicts. Do you still want to work on this? (including or excluding dependency support)

simonihmig:
Closing this because this is not up to date anymore and because of lack of feedback. @huberts if you want to continue work on this, feel free to reopen!

Issue #148, opened 2016-11-03
html5 attributes support


basz:
```
{{#bs-button title="Cancel what you are doing..." type="default"}}
    Cancel
{{/bs-button}}
```
renders as;

```
<button id="ember1391" type="button" class="ember-view btn btn-default">
    Cancel
</button>
```

simply adding 'title' to attributeBindings in addons/components/bs-button solves it.
```
attributeBindings: ['disabled', 'buttonType:type', 'title'],
```
But there might be more attributes; http://www.quackit.com/html_5/tags/html_5_global_attributes.cfm

Is there a reason these (or a subset of) are not implemented? I can easily create a PR, but I can't do tests and I would like to know if it would considered at all...

PR #149, opened 2016-11-03
Addition/support for title attribute on buttons

basz:
#148

### DIFF
--- a/addon/components/bs-button.js
+++ b/addon/components/bs-button.js
@@ -91,7 +91,7 @@ export default Ember.Component.extend(ComponentChild, TypeClass, SizeClass, {
    */
   classTypePrefix: 'btn',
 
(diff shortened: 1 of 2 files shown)


basz:
Hmm, not so sure if this is desired as the use of the title tag is actually discouraged these days.

https://www.paciellogroup.com/blog/2010/11/using-the-html-title-attribute/

simonihmig:
Thanks for your support! 

I would leave the decision about using the title attribute to the user, so I am fine with adding this.

However I would like to ask you to add the `title` property to the component definition, even with a `null` value, and to add a simple doc block so we have some documentation about it. E.g. see https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-button.js#L105-L113

simonihmig:
Cool, thanks! :)

PR #151, opened 2016-11-03
handle active state while null value for bs-button-group

apokinsocha:
```
{{#bs-button-group value=testProp type="radio"}}
    {{#bs-button value=1}}1{{/bs-button}}
    {{#bs-button value=2}}2{{/bs-button}}
    {{#bs-button value=3}}3{{/bs-button}}
{{/bs-button-group}}

```

If I set **testProp** back to null, I'm looking to get the **bs-button-group** to reset so that the active children (in the case of a radio; the single child target) corresponding **bs-button**'s **active** state to false.

Perhaps this is not considered a bug, but either way this PR fixes this use case for me and tests I have performed pass.

P.S; great ember addon

### DIFF
--- a/addon/components/bs-button-group.js
+++ b/addon/components/bs-button-group.js
@@ -202,7 +202,7 @@ export default Ember.Component.extend(ComponentParent, SizeClass, {
           } else {
             let lastActive = this.get('lastActiveChildren.firstObject');
             if (lastActive) {
-              lastActive.set('active', true);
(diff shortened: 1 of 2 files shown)


simonihmig:
Thanks for your contribution! To make sure this works and that future refactorings don't break this again, I would like to have a test that covers this specific case.

Would you be able to add one to https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-button-group-test.js?

Issue #155, opened 2016-11-07
bs-modal does not close anymore


iamsebastian:
Hello people,

Since I've upgraded to `0.11.1` from `0.9.1`, the `bs-modal` does not close anymore. Not by clicking outside the modal nor the default close button nor the `x` in the corner.

The modal is implemented inside a component, to show errors on the backend.

I've inspected through the components in the ember-inspector and found out, that the `in` property gets not set to `false`, after the `open` prop was set to `false`.

The modal ~closes fine~ fades out, if I not only overgive in handlebars the `open`-boolean (`open=theBoolThatHandlesThisInTheUpperComp`), but also the `in`-boolean (`open=theBoolThatHandlesThisInTheUpperComp in=theBoolThatHandlesThisInTheUpperComp`). But that ~is just a workaround~ was just a test for now. For sure, the backdrop keeps in the DOM and no interaction is possible. (Yeah, I know it's a private property).

Used versions (I've tried with two ember-cli versions):
- 1.13.12
- 2.9.1

Other versions:

```json
Ember             : 1.13.10
Ember Data        : 2.9.0
jQuery            : 1.11.3
Ember Simple Auth : 1.1.0
```

Update (after an hour): I upgraded ember-cli, ember-data, aso. to the actual version (~2.9), but the error still exists.

### My component in application.hbs

```hbs
{{#backend-error-modal open=openBackendErrorModal}}
  {{backendErrMsg}}
{{/backend-error-modal}}
```

### backend-error-modal comp

```hbs
{{#bs-modal open=open in=open header=false body=false footer=false submitAction=(action "submit") title=title as |modal|}}
  {{#bs-modal-header}}
    <h4 class="modal-title text-danger"><i class="glyphicon glyphicon-{{headerIcon}}"></i> {{title}}</h4>
  {{/bs-modal-header}}
  {{#bs-modal-body}}
    {{yield}}
  {{/bs-modal-body}}
  {{bs-modal-footer closeTitle="Schließen" submitTitle="Zum Login"}}
{{/bs-modal}}
```

```js
import Ember from 'ember';

export default Ember.Component.extend({
  headerIcon: 'comment',
  // Can be overriden via "title" argument on integration of component.
  title: 'Serverfehler',

  actions: {
    close() {
      console.log('Called "close" on backend error modal.');
      this.set('open', false);
    },
    submit() {
      console.log('Called submit on backend error modal.');
      this.set('open', false);
[...]

PR #156, opened 2016-11-07
Relates to #155

iamsebastian:
Relates to #155

As I've explained in #155, the modal dialog is not closing anymore. Maybe, if you have issues with glimmer2 and you are adding these `div`s inbecause of these, then there can be a way to detect the glimmer version? Modals in Ember `1.13.10` are definitely broken, when adding these `div`s.

### DIFF
--- a/app/templates/components/bs-modal.hbs
+++ b/app/templates/components/bs-modal.hbs
@@ -4,10 +4,8 @@
   {{yield this}}
 {{/bs-modal-dialog}}
 
-<div>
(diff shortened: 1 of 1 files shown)


iamsebastian:
Just as a note, with this pull request, I just wanted to mark, where the issue #155 is related to.

simonihmig:
Removing the divs did not fix the problem with Ember 1.13 for me, see #155. Closing this

Issue #129, opened 2016-08-24
Apply DDAU pattern everywhere


simonihmig:
Some (older) components are not designed with DDAU in mind. Have to fix that.
**Will probably be a breaking change!**

Let's collect all issues here:
- [ ] ButtonGroup: `value` should be oneway, call `change` action
- [ ] Alert: dismiss

simonihmig:
Closed in favor of #158

Issue #104, opened 2016-06-15
Migrate to one way controls


offirgolan:
I believe that all current input controls should eventually be converted to one-way controls. 

See here: https://github.com/DockYard/ember-one-way-controls

Not only will it be a step in the right direction, but it will also provide some much needed functionality as well as better control on input values.

simonihmig:
Thanks @offirgolan for bringing this up. This also came to my mind recently, but good to have this tracked here... 

Maybe one should completely decouple `bs-form-element` at some time from the used control and make "Custom Controls" as mentioned here http://kaliber5.github.io/ember-bootstrap/api/classes/Components.FormElement.html the default way of using this!?

One could also make an "addon addon" (like those for the validation support) for replacing the default controls with those one-way controls...

offirgolan:
@simonihmig I think that would be good. I would still make sure that form composability with this addon will be as simple as it currently is.

simonihmig:
Closed in favor of #158

PR #158, opened 2016-11-08
Road to ember-bootstrap 1.0

simonihmig:
This is a Work in Progress PR to design, discuss and track the changes that are planned for a stable 1.0 release of ember-bootstrap. **All users are encouraged to participate with comments and suggestions!**

## Scope of a 1.0 release

The basic idea of a first stable 1.0 release is not to add any new components or other features. Instead its focus will be to improve and stabilize its public API, make it coherent across components and follow modern ember best practices ("idiomatic ember"). 

In particular it should address these current issues:
* Apply the **DDAU pattern** everywhere: some older components violate this by changing public properties through two-way bindings instead of just triggering an action
* Remove the need for **private APIs**: components that follow a parent/child relationsship make use of private APIs (see the [child](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/mixins/component-child.js) and [parent](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/mixins/component-parent.js) mixins). 
* Make child components (e.g. a `NavItem` component can only be used within a `Nav`parent component) kind of **private**

The latter two issues can be solved by using the **contextual components** feature of Ember 2.3+

### Breaking change

_ember-bootstrap 1.0 will be a backwards incompatible change, because of the proposed API changes. But this will clear the way to have a stable API with strict semver compliance starting at 1.0._

## To Do

Fix DDAU (replaces #129):
- [x] Button: `active` state
- [x] ButtonGroup: `value` should be one way, call action
- [x] Alert: `visible` should be one way
- [x] FormElement: use one way controls (replaces #104)
- [x] Modal: `open` property is modified when `autoClose` is set
- [x] Accordion: `selected` property is modified
- [x] Tabs: `activeId` property is modified

Apply contextual components (replaces #86):
- [x] ButtonGroup -> Button
- [x] Modal -> ModalHeader/ModalBody/ModalFooter
- [x] Nav -> NavItem
- [x] Navbar -> NavbarNav/NavbarToggle/NavbarContent
- [x] Dropdown -> DropdownToggle/DropdownButton/DropdownMenu
- [x] Tab -> TabPane
- [x] Accordion -> AccordionItem
- [x] Form -> FormElement
- [x] Progress -> ProgressBar

Other:
- [x] Unify action names: `on<Event>`. Remove use of default action (`action`) in favor of speaking name.
- [x] Drop `bs-select`
- [x] Drop `bs-input` and `bs-textarea` as they only support two-way bindings
- [x] Move all templates from app/templates to addon/templates
- [ ] update validation addons


srvance:
@simonihmig I can restore the contextual components for navbar in the next few days. It's still in the commit history. Any other areas in particular you'd like me to take on?

simonihmig:
Hey @srvance, that would be really nice! 

As stated in my check list, I also started to rename action names to `on<Event>`  for more consistent naming. And I changed action invokation from `sendAction` calls to directly calling close action. See this commit: https://github.com/kaliber5/ember-bootstrap/pull/158/commits/05a9606a7b36295befea7c28781c737f35102c4f

When you are ready, just submit a PR to the "release-1.0" branch, not master!

srvance:
Will do! Watch for it by the end of the upcoming US holiday weekend, if not sooner.

srvance:
@simonihmig I'm thinking of nesting the navbar components. what do you think?

Also, I'm updating to ember-cli@2.9.1 and ember-2.9.0.  I'd like to eliminate the deprecation warnings on `npm install`.

simonihmig:
Yes, I forgot to mention that by myself: nesting! :) If you look at the release-1.0 branch, the `bs-modal` components has been refactored already, and all the sub components (here header/body/footer) are nested under bs-modal. Basically every contextual component that is yielded should be regarded as kind of private (in the sense that a user should not create it directly) and be nested under its public parent.

srvance:
@simonihmig So it inherits the template, after all!

Moxide:
Just a word of encouragement for you.
Big task ! The "one way controls" will be a great feature.

simonihmig:
@Moxide thanks! Most of it is done though, more or less just the one way controls are missing... :)

rstudner:
Just wanted to stop in and leave a huge THANK YOU comment to everyones hard work.  1.0 feels so close!

simonihmig:
This is directly related to the refactorings I was making for ember-bootstrap 1.0. For anyone reading this, please give some feedback! :)
https://gist.github.com/simonihmig/f4cada751b89b192f608bcb83ed1c22a
https://twitter.com/simonihmig/status/811648173932445696

simonihmig:
Just released v1.0.0-alpha, fyi!

PR #160, opened 2016-11-15
Updated ember-wormhole

simonihmig:


seawatts:
If we upgrade all the way there might need to be some changes
https://github.com/yapplabs/ember-wormhole/releases/tag/0.5.1

It would be great to get this in as it is now breaking fastboot.

PR #167, opened 2016-11-22
Release 1.0: Navbar contextual components

srvance:
Update to ember-cli@2.9.1 and ember@2.9.0
Update conventional-changelog to gulp-conventional-changelog to get rid of deprecations
Update the ember-try config
Reinstate navbar contextual components, making them nested as well and treating nav-item as a non-nested contextual component


.npmignore
```diff
@@ -1,4 +1,3 @@
-/.idea
```
simonihmig: Please leave that as it is, because I am using an IDEA based IDE! ;)

config/ember-try.js
```diff
@@ -1,2 +1,60 @@
+/*jshint node:true*/
 module.exports = {
```
simonihmig: e-b uses the "versionCompatibility" feature of ember-try (https://github.com/ember-cli/ember-try#versioncompatibility) to check for all supported ember versions. So there is no need to specify any scenarios here. Therefore also the changes to `.travis.yml` should be reverted.

addon/components/bs-navbar.js
```diff
@@ -7,38 +7,40 @@ import layout from '../templates/components/bs-navbar';
...
+ Uses the following components by a contextual reference:
 
- - [Components.NavbarContent](Components.NavbarContent.html)
- - [Components.NavbarToggle](Components.NavbarToggle.html)
- - [Components.NavbarNav](Components.NavbarNav.html)
+ - [Components.NavbarHeader](Components.NavbarHeader.html
+ - [Components.NavbarContent](Components.NavbarContent.html
+ - [Components.NavbarToggle](Components.NavbarToggle.html
+ - [Components.NavbarBrand](Components.NavbarBrand.html
+ - [Components.NavbarNav](Components.NavbarNav.html
```
simonihmig: the closing parenthesis seems to be missing for all the links! :)

addon/components/bs-navbar.js
```diff
@@ -7,38 +7,40 @@ import layout from '../templates/components/bs-navbar';
...
-       {{#bs-nav-item}}{{#link-to "collapse"}}Collapse{{/link-to}}{{/bs-nav-item}}
-       {{#bs-nav-item}}{{#link-to "modal"}}Modals{{/link-to}}{{/bs-nav-item}}
-       {{#bs-nav-item}}{{#link-to "progress"}}Progress bars{{/link-to}}{{/bs-nav-item}}
-       {{#bs-nav-item}}{{#link-to "navs"}}Navs{{/link-to}}{{/bs-nav-item}}
-       {{#bs-nav-item}}{{#link-to "navbars"}}Navbars{{/link-to}}{{/bs-nav-item}}
-     {{/bs-navbar-nav}}
-   {{/bs-navbar-content}}
+   {{#navbar.content}}
+     {{#navbar.nav}}
+       {{#navbar.nav-item}}{{#link-to "alert"}}Alert{{/link-to}}{{/navbar.nav-item}}
```
simonihmig: You have the nav-item yielded by the navbar itself, probably because the nav itself is not yet refactored to use contextual components, right? 

In this case I would then change this to to have the actual nav yield the nav-item, as soon as the nav has been refactored!

addon/components/bs-navbar/toggle.js
```diff
@@ -0,0 +1,19 @@
...
+ * @class NavbarToggle
+ * @namespace Components
+ * @extends Components.Button
+ * @public
+ */
+export default BsButtonComponent.extend({
+  layout,
+
+  classNames: ['navbar-toggle', 'collapsed'],
+  classNameBindings: ['collapsed::active'],
```
simonihmig: How come you are using the active class here? I have not found this to be documented on the bootstrap docs.

addon/components/bs-navbar.js
srvance: Actually, since `e-b-navbar-nav` is derived from `e-b-nav` and templates aren't inherited, the nav/nav-item CC change won't inherit. I talked to some core folks, and components not inheriting templates is a deliberate choice. I think it would be bad to hack it in.

I debated whether to put it under the nav (e.g. `navbar.nav.item` or do it the way I did as `navbar.item`. This seemed simpler and sufficiently encapsulated, but it's a judgment call.

addon/components/bs-navbar.js
srvance: Oops!

simonihmig:
Thanks again! Highly appreciated! :)

srvance:
Gladly! Any other part of this you'd like me to help with?

PR #168, opened 2016-11-23
supporting dropdown to open on hover (opt-out)

gokatz:


simonihmig:
Thanks @gokatz for contributing! 

Unfortunately I have to say that I will not merge this, because:
* on hover dropdown are not supported in the original bootstrap, and this addon tries to match bootstrap as closely as possible/useful
* adding custom app-specific behaviour like this one is pretty easy, by overriding the original dropdown component inside the user's app folder, or by creating a new one that extends from it.

Hope you understand, and thanks again for taking the time!

PR #172, opened 2016-12-01
Addition/readonly attribute

basz:
This PR adds support for the readonly attribute, which differs from the disabled attribute

- A read-only field can't be modified, but, unlike disabled, you can tab into it, highlight it, and copy its contents.
- readonly is only relevant for type text, search, url, tel, email, number, password, and the date/time input types. It is also valid on <textarea>
- readonly is ignored for type hidden, range, color, checkbox, radio, file, button, submit and image

### DIFF
--- a/addon/components/bs-form-element.js
+++ b/addon/components/bs-form-element.js
@@ -206,6 +206,15 @@ export default FormGroup.extend(ComponentChild, {
   disabled: false,
 
   /**
+   * Control element's HTML5 readonly attribute
+   *
+   * @property readonly
+   * @type boolean
+   * @public
+   */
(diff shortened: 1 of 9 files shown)


simonihmig:
Looks good in general! But you have modified the template for the select element only for the vertical form layout, not horizontal and inline. Probably forgotten?

basz:
That must be a mistake - select does not support read only anyway and its deprecated afaik.

Ill fix that soon

simonihmig:
@basz Thanks a lot for this contribution! 👍

basz:
You are welkome! Thanks for merging!

PR #176, opened 2016-12-13
removed all unneeded .gitkeep files

RyanNerd:
The .gitkeep files exist for folders that do not have any content (including other folders) in them so that git will include the folder in the commit/push. Empty folders do not get committed or pushed by git is the reason.

I removed all the .gitkeep files that were not in use.

This has no impact on logic and is just an [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder) thing I have when I see unneed code or files.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,2 @@
+/api
+CHANGELOG.md
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,2 +1,0 @@
-/api
-CHANGELOG.md


simonihmig:
Funny variant of OCD that you have! ;)

Thanks anyway, will happily merge. However there is one issue: you also removed the `docs/.gitignore` file, but this would still be needed (some gulp task for generating docs will put files in there, that should not be commited).

simonihmig:
Thanks again!

PR #180, opened 2016-12-27
Fixed the Collapse example in the 1.0.0-release branch

RyanNerd:
**This is intended/targeted for the 1.0.0-release branch**
The collapse example button was not responding to mouse clicks.


simonihmig:
Thanks for catching this! But actually it should be `onClick=(action "toggle")`! Can you update?

simonihmig:
Fixed this in your branch already and merged it. Thanks!

PR #185, opened 2016-12-28
Restore the label for the vertical checkbox form element

srvance:
Restore the label for the vertical checkbox form element that was removed in dda8e3605efe46d9f6583c826dd955e60daafcb8.


simonihmig:
Thanks!

PR #192, opened 2016-12-30
Remove deprecated Ember.K

cibernox:
This PR was generated automatically using ember-k-codemod

### DIFF
--- a/addon/components/bs-accordion.js
+++ b/addon/components/bs-accordion.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import layout from '../templates/components/bs-accordion';
 import listenTo from '../utils/listen-to-cp';
 
-const { K: noop } = Ember;
+const {} = Ember;
 
 /**
  Bootstrap-style accordion group, with collapsible/expandable items.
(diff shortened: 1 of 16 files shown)


addon/components/bs-accordion.js
```diff
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import layout from '../templates/components/bs-accordion';
 import listenTo from '../utils/listen-to-cp';
 
-const { K: noop } = Ember;
+const {} = Ember;
```
simonihmig: Hey cool, I knew I would have to update things after the RFC was accepted, but you were faster! Thanks! :)

Unfortunately the codemod will leave empty destructuring statements like the above, which causes failing JSHint tests. I can fix that manually, but maybe there is a way to fix that also in the codemod?

cibernox: Interesting. I didn't consider this. Seems that the codemod should be smart enough to do that

cibernox:
@simonihmig I've made another PR with the updated codemod and now empty destructuring is removed. I created another one because actually I do this with a script, that I run agains ALL ember-addons.

PR #193, opened 2016-12-31
Remove deprecated Ember.K

cibernox:
This PR was generated automatically using ember-k-codemod

### DIFF
--- a/addon/components/bs-accordion.js
+++ b/addon/components/bs-accordion.js
@@ -2,8 +2,6 @@ import Ember from 'ember';
 import layout from '../templates/components/bs-accordion';
 import listenTo from '../utils/listen-to-cp';
 
-const { K: noop } = Ember;
-
 /**
  Bootstrap-style accordion group, with collapsible/expandable items.
  See http://getbootstrap.com/components/#btn-groups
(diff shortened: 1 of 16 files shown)


simonihmig:
Thanks, dear codemod bot! :)

wukongrita:
@cibernox would you kindly to make this PR for v 0.x too ?
I'm still using the v 0.x, but the ember already 2.12 and got the same warning (DEPRECATION: Ember.K)

thanks before..

simonihmig:
@wukongrita This was done using an automated approach, redoing this manually for 0.x is probably not something @cibernox should be bothered with. Apart from that, I am a bit hesitant to add "fixes" like that to the 0.x branch, since all further development should go to the latest version. Fixes for 0.x should be restricted to real bugs or eventual security issues. But in this case it is not even a bug, as it is just a deprecation, still continues to work till Ember 3.0 I assume.

If that is really important to you, feel free to create a PR, from and to the `release-v0.x` branch (https://github.com/kaliber5/ember-bootstrap/tree/release-v0.x). You can use https://github.com/abuiles/ember-watson to automate that.

PR #206, opened 2017-01-20
Support Bootstrap 4

simonihmig:
This is a work in progress, still a lot to do. Supersedes https://github.com/kaliber5/ember-bootstrap/pull/181

## To Do

- [x] separate component trees for BS3 and BS4
- [x] Implement BS4 components (build on https://github.com/kaliber5/ember-bootstrap/pull/181)
- [x] Proper option for selecting the Bootstrap version to be used
- [x] Don't import glyphicons in BS4
- [x] Restrict preprocessors for BS4
- [x] import static BS4 CSS
- [x] import BS4 SASS
- [x] Assess whether blueprint handles switching from BS3 to BS4 and back
- [x] separate tests for BS4
- [x] Consolidate BS3/4 documentation under BS3 docs
- [x] Run BS4 tests on CI
- [ ] Fix form validation markup test
- [x] Update documentation for BS4 migration

## Components to update

- [x] Accordion
- [x] Alert
- [x] Button
- [x] ButtonGroup
- [x] Collapse
- [x] ContextualHelp
- [x] Dropdown
- [x] Dropdown.Divider
- [x] Form
- [x] FormElement
- [x] FormElement partials
- [x] Form validation feedback icons
- [x] FormGroup
- [x] Modal
- [x] Modal Simple
- [x] Nav
- [x] Nav.Dropdown
- [x] Nav.item
- [x] Nav.LinkTo
- [x] Navbar
- [x] Popover
- [x] ProgressBar
- [x] Tab
- [x] Tooltip
- [x] `nav-link` active state (affects Navs and Tabs) (deferred and captured in #248)

## Open Items

* Why aren't default tabs initial state outlined properly? Diagnosis: The `active` class needs to go on the element with the `nav-link` rather than the `nav-item`. See [this](https://github.com/twbs/bootstrap/issues/21223) and [this](https://github.com/twbs/bootstrap/issues/20156).
  * Change 1caa0ebb2c9aabfb9c0f9896301b1a00a532ee82 has a little leakage of BS4 classes into common templates with `nav-link`. To fix the `active` state issue, there's even more. However, it's a pretty complex template to duplicate and i'm not sure refactoring out a new component is worth it.
  * The same kind of issue exists with dropdown items within a nav.


index.js
```diff
@@ -16,46 +20,71 @@ var defaultOptions = {
 module.exports = {
   name: 'ember-bootstrap',
 
-  included: function included(app) {
+  included(app) {
```
srvance: What versions of Node are we supporting? Does the support for Ember 2.3+ also include ember-cli 2.3+? They only dropped Node 0.12 support in ember-cli 2.10, so can we rely on ES6 syntax? I don't think Babel is used for ember-cli, is it? This was pointed out when I made similar changes in the PR against my fork.

simonihmig: I think we can, at least the stuff that is supported in node 4.x, which is the next version to 0.12. See http://node.green. I think Ember-cli and other addons are doing that transition now.

simonihmig: Btw, already did some changes on the master branch that also moved to ES6 syntax, see #208. These are some things I wanted to do also in preparation for the BS4 changes...

index.js
```diff
@@ -16,46 +20,71 @@ var defaultOptions = {
...
   },
+  
+  getBootstrapVersion() {
+    return 'bs3'; // @todo replace with dynamic config
+  },
+
+  treeForAddon() {
+    let tree = this._super.treeForAddon.apply(this, arguments);
+    let bsVersion = this.getBootstrapVersion();
+    let otherBsVersion = bsVersion === 'bs3' ? 'bs4' : 'bs3';
```
srvance: Let's refactor this to a method, as well.

index.js
srvance: I'm good with that. I just want to make sure we're not invalidating our backwards compatibility targets.

index.js
```diff
@@ -16,46 +20,71 @@ var defaultOptions = {
...
     }
 
     if (options.importBootstrapTheme) {
       app.import(path.join(bootstrapPath, 'css/bootstrap-theme.css'));
-      app.import(path.join(bootstrapPath, 'css/bootstrap-theme.css.map'), {destDir: 'assets'});
+      app.import(path.join(bootstrapPath, 'css/bootstrap-theme.css.map'), { destDir: 'assets' });
     }
 
     // Import glyphicons
     if (options.importBootstrapFont) {
```
srvance: This will also become bs3-only.

srvance:
Want me to apply BS4 changes on this branch then?

simonihmig:
> Want me to apply BS4 changes on this branch then?

Yes, that would be awesome!

There was some overlap with the work I did and your last commits. I just pushed some changes...

Now the proper BS4 assets are also imported when the `bootstrapVersion` is set to 4! I chose to install bootstrap@4 via npm in this case. I think there is a trend of getting rid of bower deps, and this would make it easy when sometime BS3 support would maybe eventually be dropped...

BTW, you used ES6 restructuring in you last commit, and that does not seem to be supported by node 4, thus the failing test. See me node.green link!

srvance:
> BTW, you used ES6 restructuring in you last commit

I realized overnight that might be the cause. Thanks for finding that. The error wasn't very clear and I'm running Node 6 so couldn't reproduce it.

simonihmig:
@srvance No problem. Wasn't aware of that Node 4 limitation myself until investigating what we can use now when dropping node 0.12 support. This was occurring only within the Fastboot tests, I added a debug flag there so the error messages in the Travis log should be a bit more understandable now...

srvance:
@simonihmig Any thoughts on how to deal with documentation differences between the two?

addon/components/bs4/bs-collapse.js
```diff
@@ -1,2 +1,5 @@
-export { default } from 'ember-bootstrap/components/base/bs-collapse';
+import Collapse from 'ember-bootstrap/components/base/bs-collapse';
 
+export default Collapse.extend({
+  classNameBindings: ['collapse', 'in:show', 'collapsing']
```
srvance: @simonihmig Looking for an opinion here. I overrode the `classNameBindings` that were in the base, reusing the `in` method. For this kind of thing should we generalize the method name and have `classNameBindings` defined separately in the bs3- and bs4-specific components?

I had a surprise at first because we already have a `show` method that isn't designed to support a class name binding.

srvance:
@simonihmig How do you want to handle the form-element partials? They're not separated, but they need to be.

simonihmig:
> Any thoughts on how to deal with documentation differences between the two?

For the API docs, this would be straight forward as we have separate class files where we can put the YUIDoc annotations. Would just need a way to generate just another API docs variant. Will add a To Do...

> How do you want to handle the form-element partials? They're not separated, but they need to be.

Oh, I hoped not! :) Will then have to setup some separate trees for them as well. Or refactor them to components?

> Do we treat BS3 as the base or do we refactor BS3-unique things into that tree

I would put s much as possible (all common things) into the base component. But if e.g. a class is only used for a BS3 component, and another class for BS4, then add those classes to the bs3/bs4 classes extending base. So we don't have to "override" things in BS4 components that we don't need there. Does that make sense?

addon/components/bs4/bs-collapse.js
simonihmig: Sounds good. `classNameBindings` is problematic in a way, as it is a `mergedProperty` (http://emberjs.com/api/classes/Ember.Object.html#property_mergedProperties). So overriding does not work properly. So yeah, remove from base what is not needed in bs4, and add that to bs3. Make sense?

srvance:
> For the API docs, this would be straight forward as we have separate class files where we can put the YUIDoc annotations. Would just need a way to generate just another API docs variant.

I don't know how well the docs inherit and merge. Would we copy down the docs and adjust them or is there a way to do it additively? The usage summaries will have the biggest conflicts as some options may have been removed while others added. The same could apply for the sample app.

> Oh, I hoped not! :) Will then have to setup some separate trees for them as well. Or refactor them to components?

Do you have a preference for separate partial trees vs. components? Search for `form-element` [here](https://github.com/kaliber5/ember-bootstrap/pull/181/files) to see the nature of the changes.

srvance:
Just pushed an example of pulling BS3-specifics out of base. Let me know what you think.

srvance:
Similar change for dropdown. Would prefer to rename the `open` property to `isOpen`. I think it only affects the internal actions and maybe some tests.

simonihmig:
> Just pushed an example of pulling BS3-specifics out of base. Let me know what you think.

Looks good to me!

> Do you have a preference for separate partial trees vs. components? 

Hm, not really. Was using partials because they all share the same context, so separate components felt like a bit of overhead. Maybe neglectable... For now will setup separate partials trees as well.

srvance:
Sounds good on partials. I'll wait for that until I see your changes.

Build failure was a stall in the fastboot build.

simonihmig:
> Would prefer to rename the open property to isOpen

Go ahead!

srvance:
@simonihmig There are several new contextual components I want to create: `dropdown.divider`, `nav.dropdown`, `nav.link`. These should also go to master as API changes in preparation unless we want to merge this before we finish the 1.0 alpha series.

srvance:
@simonihmig Please take a look at the `nav.dropdown` change. There's a little bit of compatibility leakage in the use of the `inNav` property for the dropdown and dropdown toggle, but I deemed it harmlessly better than the duplication necessary to avoid it.

simonihmig:
> There are several new contextual components I want to create: dropdown.divider, nav.dropdown, nav.link. 

nav.link should already be there, created that some time ago "on your request" for this reason. I am fine with the others as well. Wenn adding them also for BS3 we must make sure the docs are properly updated (API + dummy/demo app). 

> These should also go to master as API changes in preparation unless we want to merge this before we finish the 1.0 alpha series.

This branch should be merged into master within the alpha stage, so it's ok to just add them here!

> Please take a look at the nav.dropdown change.
 
I am fine with that!

stevesims:
Really looking forward to seeing this land.

One quick comment - whilst you've ticked off that the Accordion components have been updated, they are still using `panel` classes, which have been dropped in favour of `card` in BS4.

If I get a chance I'll make a PR to address that later today

srvance:
@stevesims That would be awesome. Really the checklist covers whether I've gotten it to be functional and apparently styled correctly. I'm sure there are nuances of the styling I've missed. It's quite a large endeavor, so any help would be appreciated.

srvance:
> nav.link should already be there

`nav.link-to` is already there. I think we should create a component for links that are anchor tags rather than `link-to` helper usages. In addition to requiring the `nav-link` class, I'm finding they've changed which element needs the `active` styling for it to work properly.

stevesims:
@srvance managed to find the time to make a PR (#213)

Very glad to see you've done the changes needed to get modals working - I was going to do that next.  😀

srvance:
@stevesims I'm cranking through things pretty quickly. Let me know if you want something reserved.

simonihmig:
There are now separate trees for the partials! fyi

Btw, two failing tests, for alert and modal...

addon/components/base/bs-alert.js
```diff
@@ -30,7 +30,7 @@ const {
  */
 export default Ember.Component.extend(TypeClass, TransitionSupport, {
   layout,
-  classNameBindings: ['alert', 'fade', 'in', 'alert-dismissible'],
+  classNameBindings: ['alert', 'fade', 'dismissable:alert-dismissible'],
```
simonihmig: this should be 'dismissible:alert-dismissible'!

srvance: Doh!

srvance:
Sorry about the failures. I'm working in a funny state until we figure out how to get the dummy app and tests working against both versions.

stevesims:
@srvance no need to reserve anything for me  😀 
I'm a bit distracted with some other things right now, but will let you know if I spot something else that I feel needs attention and if it's something quick I'll submit another PR

stevesims:
Just noticed that the build options weren't getting merged in properly, so the option to enable bs4 in my app's `ember-cli-build` file was getting ignored - oops!

PR #220 fixes this

srvance:
@simonihmig Thank you for the testing enablement! It's making a world of difference.

srvance:
@simonihmig The overwhelming majority of test differences are turning out to be simple alternative classes, such as `in`/`show`, `top`/`tooltip-top`. In many cases, I'm duplicating significant test logic simply to make simple class replacements. What do you think about adding a helper to facilitate data alternatives, similar to our `versionDependent` discussion but with a better name?

simonihmig:
@srvance I have a deal for you: you come up with a better name, and I will implement it. How's that? 😜

srvance:
Yeah, naming is hard. :)

We could even cover around 75% of the differences simply with a `visibleClass` helper.

srvance:
I don't mind implementing it. I just figured some design discussion should happen first.

srvance:
With this push, all tests except accordion and form element pass for BS4. I'm working on forms right now. The accordion work needs to be looked at to see if it should have the `card-group` to replace `panel-group`.

srvance:
The only failing test for BS4 is for form element.

joshuairl:
Anybody else get infinite loops when components error?
I'm thinking it's this plugin that is causing it. :-/

I'm not familiar with the plugins mixins and things to know if it could be overriding something in base ember...

srvance:
@joshuairl I haven't seen or heard of it. The mixins are pretty straightforward. Nothing special there.

What kinds of errors? When e-b components error or components in general? Which components? Using BS3 or BS4? What makes you think it might be e-b causing it? If it's not specific to this branch, you should open an issue rather than appending to this PR.

joshuairl:
@srvance Sorry about this.
I had an outdated version of the BS4 codebase.
How does one run this latest branch and enable the bootstrapVersion: 4 instead of 3?

srvance:
@simonihmig Following our discussion, I don't think it's practical to proxy all of the [anchor tag's attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes) and the [global attributes](https://developer.mozilla.org/en-US/docs/HTML/Global_attributes). Given the discussions in Bootstrap about the relationship between `nav-link` and `active`, I'd like to defer addressing the issue until the next Bootstrap release and forget about the `nav.link` component for now.

bgentry:
Quick question on this... Does bootstrap-sass still need to be installed? I'm still getting the following line added to my package.json even though I'm using v4:

```
"bootstrap-sass": "^3.3.7"
```

bgentry:
I noticed some differences from the bootstrap v4 docs with the modal component. In all the modal examples on the official bootstrap 4 docs, the modal title is an `h5`, not an `h4`. Also in those examples, the close button comes _after_ the title in the DOM. Compare the following expected vs. actual results when using the `bs-modal` invocation directly as it appears on the docs page:

Actual | Expected
-- | --
screen shot 2017-02-07 at 3 24 52 pm | screen shot 2017-02-07 at 3 24 05 pm

Presumably this is due to implementation differences between v3 and v4. It'd be easy to change it to the v4 way, but that would break compatibility. Would we need to add version-specific clauses to the modal's hbs to make this work perfectly on both versions?

srvance:
@bgentry Thanks for pointing that out! Only the one that corresponds to your Bootstrap version actually makes it into the build, but I just pushed a fix to make the build only error when `bootstrap-sass` is missing with Bootstrap 3.

index.js
```diff
@@ -65,12 +74,16 @@ module.exports = {
...
           this.ui.writeLine(chalk.red('Npm package "bootstrap-sass" is missing, but required for SASS support. Please run `ember generate ember-bootstrap` to install the missing dependencies!'));
           return false;
         }
         break;
       case 'less':
+        if (this.getBootstrapVersion() === 4) {
+          this.ui.writeLine(chalk.red('There is no Less support for Bootstrap 4! Falling back to importing static CSS. Consider switching to Sass for preprocessor support!'));
+          return false;
+        }
         if (!('bootstrap' in bowerDependencies)) {
```
srvance: @simonihmig I think this is wrong now. Does the `bootstrap-sass` distro also include LESS dependencies?

srvance:
@bgentry Thanks for pointing out the modal issue. Looks like it's documented in the [Bootstrap migration notes](https://v4-alpha.getbootstrap.com/migration/#modal) and I missed it. I'm guessing their change from `h4` to `h5` is probably because they increased the default font size overall. I'll get a fix in for this right away.

bgentry:
@srvance thanks for all the quick fixes! 🙌  I'll make sure to report anything else I notice as I get my app using the components.

index.js
simonihmig: No, `bootstrap-sass` does not contain Less. But what exactly is wrong here?

srvance: I was going to say it wasn't checking the bootstrap version, but hadn't fully thought through the effect of the conditional above it.

srvance:
@simonihmig Would you prefer to merge the master changes back up before I tackle the BS4 form changes or should I?

srvance:
In `bs-form.element.label.hbs` why aren’t the classes applied to the block form use of `<label>`? It seems block form is only used for horizontal and vertical checkboxes. It would really simplify things if I could make the `label` component with a `tagName` and `classNameBindings` and it seems like the omission of the classes might be a bug.

srvance:
For `bs-form.element.errors` what do you think of having `show` control visibility classes rather than a template conditional? I just made a change that needs to set the inner `span` class with a variable and it seemed like something better done with `classNames` but the conditional prevents that.

simonihmig:
> In bs-form.element.label.hbs why aren’t the classes applied to the block form use of <label>

The block form is used when wrapping the `<label>` around the form control (checkbox). In this case there is no class on the label, at least for BS3: http://getbootstrap.com/css/#checkboxes-and-radios

> For bs-form.element.errors what do you think of having show control visibility classes rather than a template conditional?

I would prefer to have no extraneous markup when there are actually no errors

srvance:
@simonihmig I just pushed a change to run a couple BS4 scenarios in testing in addition to all of the normal try scenarios. Do we want to run BS4 tests against all versions or just representative ones? Right now, I chose latest release and oldest (2.3.1). I could see value in canary and beta, as well.

I wanted to get something running to start.

srvance:
Looks like BS4 puts classes on the checkbox label. I'll differentiate the components.

simonihmig:
Yeah, makes sense. I would add at least `beta`. Maybe even add the LTS versions (like in the default ember-try config), so 2.4 and 2.8. If we were using some private APIs, they could have been dropped in one of those LTS releases, so testing them maybe makes sense as well.

Using the `versionCompatibility` feature of ember-try for BS3 leads to a lot of testing (like all minor versions between 2.3 and release, + beta + canary). On the one hand it is more a best practice to test using the above mentioned scheme (LTS + release + beta + canary), on the other hand having tests for all versions did catch some errors in the past (like errors caused by ember-wormhole, which uses private APIs, see #160). Not so sure about that...

srvance:
It would be easy to run the complete set of versions against BS4 with a second try config. It's just CI cycles.

srvance:
I did some experiments in bs3. Adding `control-label` to checkbox input labels in horizontal and inline layouts doesn't have any visible impact. Although it's not in the BS3 examples, it seems a strange inconsistency, especially with `sr-only`. 

I filed issue twbs/bootstrap#21998 to ask.

simonihmig:
@srvance it does have some unintended impact on a horizontal layout, adding some unwanted padding
image

srvance:
@simonihmig Looks like horizontal forms have been eliminated as their own thing.

https://v4-alpha.getbootstrap.com/components/forms/#using-the-grid
https://v4-alpha.getbootstrap.com/migration/#forms

You're supposed to use the grid classes. I don't think we should get in the middle of that, especially when you consider all the responsive breakpoint possibilities. It makes me question whether we should even add the `row` class to the `form-group` for horizontal layouts.

I'm thinking we drop it entirely for BS4. What do you think?

srvance:
Also, inline forms look pretty bad without the explicit padding classes they use in [the examples](https://v4-alpha.getbootstrap.com/components/forms/#inline-forms). I can't think of a good way to build that into the components.

Given twbs/bootstrap#21581, I'd like to defer finishing forms until the BS4 beta.

srvance:
@simonihmig I changed the remaining failing test for BS4 into a BS3-only test. Can you look at it to see if you can get it working for BS4? It's about the validation markup. If you comment out the first assertion, the rest of the test passes, so it's just about the init.

srvance:
I propose we merge can carry a small to do list forward.

simonihmig:
@srvance yeah, the icon position test is actually quite annoying, it's failing in chrome because of some pixel offsets, but passing in PhantomJS. Would like to fix that anyways, or even drop it when not possible...

I will do a more detailed review given that it seems you have migrated all components now. Happy to merge this soon! 😁

RSchwan:
I tried your implementation and fund some styling errors when the validation for a form element fails ([See docs](http://v4-alpha.getbootstrap.com/components/forms/#examples)):

Actual | Expected
------------ | -------------
010058be011de51c0f9a463d12d86e26 | ce56916d26f6eaee13802672938aefcc
dc18e20756985a1fcb76b0e7a4cd0b75 | 0d60c9d3083f24c2195e604e891ee457

I also think the styling is off when the validation is successful.

Awesome work btw 🎉

srvance:
Thank you for both the report and the appreciation, @RSchwan!

I'm not as familiar with the forms, and especially the validation, so I've asked @simonihmig to look it over. From what you're showing I see a difference in `has-error` vs. `has-danger` and `form-control-danger`.

simonihmig:
@RSchwan seems you are right. BS3 expects a `has-error` class while it is `has-danger` in BS4. 
@srvance This CP should be customized for BS4: https://github.com/kaliber5/ember-bootstrap/blob/bs4/addon/components/base/bs-form/element.js#L568. Maybe just add some private props that define the class names, so one has only to customize the class name props for BS3 and BS4, not the whole CP implementation?

srvance:
I fixed the class issue, but there's also most likely an icon issue since BS4 doesn't include glyphicons.

RSchwan:
@srvance It looks like BS4 includes some icons directly as embedded svg:
c127289e29411e837772016decf1cb6d
You just have to add the `form-control-danger` class to the input.
Btw the error text should also have the class `form-control-feedback`.
Maybe also have a look into the BS4 docs. They have some good examples: [http://v4-alpha.getbootstrap.com/components/forms/#examples](http://v4-alpha.getbootstrap.com/components/forms/#examples)

srvance:
Thanks, @RSchwan. I'm digging into it.

simonihmig:
@srvance merged master, and fixed some API docs issues. Reviewed it, looks absolutely great. I think when the last remaining issues are sorted out, we should be ready to merge! 👏

srvance:
Excellent. I think the validation icons is the only issue. We need to wait for BS4 beta for the nav-link active issue and we'll handle that separately. Does that sound right to you?

simonihmig:
I am fine with that. 

Also we must add some instructions, probably to the 'Getting started' guide for how to switch to BS4. Do you see a need for some kind of migration guide, besides setting the `bootstrapVersion` option and rerunning the default blueprint for existing projects?

srvance:
Agreed. I think the most minimal migration guide would tell people they need to use the contextual components as much as possible and would point to the BS4 migration page. This would be not only for BS4 migration, but for e-b 1.0 migration. A more complete one would enumerate e-b changes. How far do you think we should go to meet your vision for user support?

srvance:
Nice!

simonihmig:
@srvance just updated the "getting started" section of the demo app to allow for some better documentation. Stills needs some work. If you feel you want to add something there, feel free to do so! :)

Btw, lost so much time trying to increase the base font size to 16px (as it was before). Seems not to be possible, as the styles are also used for test runs, and setting the font size to `px` values breaks for BS4 which expects `rem`... 😡

srvance:
@simonihmig Will do on the documentation. It's taking me awhile to see a path through the forms work.

You'd think they would have defined the various font sizes relative to some small set of reference sizes. The [migration doc](https://v4-alpha.getbootstrap.com/migration/#global-changes) say the base size was changed to 16px from a previous 14px. It looks like there are at least 12 variables you have to change to increase the font metric globally. Should we file an issue to make it all computed by default?

simonihmig:
> Will do on the documentation. It's taking me awhile to see a path through the forms work.

That would be great. I would think we should keep this small, like what the breaking changes are when switching from BS3 to BS4 (regarding ember components), how to do that (set config + run blueprint). Certainly not duplicate the BS4 docs itself, just reference them and keep it more or less minimal. We can still work in details if we see users struggle at certain points...

What's the deal with the form stuff, you want to document something there?

Regarding the font sizes the problem ist just that the `variables.scss` is meant to be used for the demo app running on BS3, but the same file is compiled also with the BS4 styles when running BS4 tests. But BS4 expects font sizes in `rem`, when setting a `px` value there, it will break the build (see the failed Travis run yesterday) as it cannot convert from `px` to `rem`.

So when you have something like this:

```sass
$foo: 10rem;
$bar: $foo + 5rem;
```

This should work as long as `$foo` is in `rem` units, with `px` the `$bar` assignment will break.

It's tricky... maybe should have a separate repo for the demo/docs app and use the dummy app just for the few acceptance tests!?

srvance:
> What's the deal with the form stuff, you want to document something there?

There are significant differences between the feedback mechanisms in BS3 & BS4. It's mostly about the internal complexity of the implementation and that it has some decisions that align closely with the BS3 implementation that are different in BS4. I haven't used it before, so part of it is learning curve, as well. The class usage has simplified and the icons are handled completely differently.

I tried to get the Bootstrap folks to add a variable to indicate Bootstrap version that could be used for conditionals. I was going to use it to conditionally define `icon-bar` before we had the version-specific templates and I created the inline version of the navbar toggle. That would help with your font size issue, as well.

> maybe should have a separate repo for the demo/docs app and use the dummy app just for the few acceptance tests!?

That seems to be the way the larger projects are going, including Ember itself. For Ember, you put Ember, Ember Data, and the web site repos into side-by-side directories to build the production site.

simonihmig:
> There are significant differences between the feedback mechanisms in BS3 & BS4. It's mostly about the internal complexity of the implementation and that it has some decisions that align closely with the BS3 implementation that are different in BS4.

Ok, I thought it was about the documentation effort, but than your comments are not related to that...

If the current implementation is kind of misaligned to BS4, I can have a look and see if I can optimize this. What's the biggest issue there?

srvance:
Take a look at my PR. I did this part on a separate branch. I wonder if we should use the contextual components to give more layout control back to the user, simplifying the internal implementation.

1. `has-feedback` was removed. There's one mention of it in the Bootstrap files, but I think it may be vestigial.
1. Feedback icons are no longer separate elements, but background-image SVGs. The markup goes on the input element now. I'm not sure how we should handle if someone wants to us FA or glyph instead of changing the SVG variable. That's one argument for giving more control.
1. We don't support the `form-text` at the moment and if they put it in the block, it will come between the validation feedback and the field, which probably isn't what they want. Another argument for more control.

An argument against giving the control is the sensitivity of the formatting to the layout orientation, but I think only the demo app probably cares about being able to switch between orientations automatically. I suspect most people have a single orientation that is responsive. Do you know of anyone using the ability to dynamically switch layouts?

simonihmig:
> Do you know of anyone using the ability to dynamically switch layouts?

No, certainly not. That's just for the showcase. But I do think that putting a form together should be as simple as possible for the user, so having a totally easy way to just define the form fields without bothering for how the markup should be structured (that kind of includes validation) is one of the biggest benefits of the form element component. So regarding simplicity I would say yes, make it as simple as possible, but for the *user*, not necessarily for us *implementors*! ;) 

So unless you have some idea, how both cases of simplicity can be improved, I would stick to the current solution. 

> Feedback icons are no longer separate elements, but background-image SVGs. The markup goes on the input element now. I'm not sure how we should handle if someone wants to us FA or glyph instead of changing the SVG variable. That's one argument for giving more control.

Without having really looked at it or tried it out, but I think if the icon in BS4 comes from the plain CSS, it should be even easier to customize, just by customizing the CSS. I guess that's the most elegant way of customizing it, if it does not involve code changes, just CSS!?

> We don't support the form-text at the moment and if they put it in the block, it will come between the validation feedback and the field, which probably isn't what they want. Another argument for more control.

As I already said in the PR, I would like to support this out of the box at some time...

srvance:
> So unless you have some idea, how both cases of simplicity can be improved, I would stick to the current solution.

What do you think of having `label`, `feedback`/`errors`, `help`/`text`, and for BS3 `icon` as contextual components of `element`? We could use a helper to encapsulate the complexity of calculating the right component name based on layout and type. That would move the decisions about inclusion from internal logic to simply how the user composes the element and allow simple and flexible layout choices. The question of the label being a peer to the content versus wrapping around it is the only hitch I immediately see, which might mean the label isn't contextual.

Yes, the icons are as simple as changing the CSS. I don't know how that would work with FA or Glyph, and it makes the config settings irrelevant.

simonihmig:
We might yield those components. But this would be a breaking change, as the block form of `bs-form/element` was used before just for using a custom control element, things like label, error would still be added automatically. So when switching to yielded components, this would not make sense anymore...

I think we could explore that in a separate PR and see if it makes sense. But this should not block merging `bs4`, right?

srvance:
> I think we could explore that in a separate PR and see if it makes sense. But this should not block merging bs4, right?

I completely agree.

NullVoxPopuli:
this is great! thanks for working on this!

PR #207, opened 2017-01-20
Add dialog class names

ryanrishi:
This will allow you to do something like this:

```hbs
{{bs-modal title="Hello World" dialogClassNames="foo bar"}}
```

and have a modal with `foo` and `bar` classes added to the dialog.

### DIFF
--- a/addon/components/bs-modal-simple.js
+++ b/addon/components/bs-modal-simple.js
@@ -15,7 +15,7 @@ import layout from '../templates/components/bs-modal-simple';
  This will automatically create the appropriate markup, with a modal header containing the title, and a footer containing
  a default "Ok" button, that will close the modal automatically (unless you return false from `onHide`).
 
- A modal created this way will be visible at once. You can use the `{{#if ...}}` helper to hide all modal elements form
+ A modal created this way will be visible at once. You can use the `{{#if ...}}` helper to hide all modal elements from
  the DOM until needed. Or you can bind the `open` property to trigger showing and hiding the modal:
 
  ```hbs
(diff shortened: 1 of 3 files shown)


simonihmig:
First, thanks for contributing!

You can already pass a class with the standard `class` property, see https://github.com/kaliber5/ember-bootstrap/blob/master/addon/templates/components/bs-modal.hbs#L11. This is passed to the `.modal` element (not the `.modal-dialog` as in your case). But for styling this would be sufficient I think? 

Do you have a use case that really *requires* setting the class to the `.modal-dialog` element?

ryanrishi:
You're right, it makes more sense to put it on the `.modal` element.

It looks like the when I have a template like this:

```hbs
{{#bs-modal
class="foo bar"}}
{{! some custom bs-modal-body and bs-modal-footer
{{/bs-modal}}
```

then `foo bar` is being added to the base component's DOM, _not_ the wormhole. Confirming w/ jQuery, if I do $('.foo'), the result is `div#ember3095.foo.bar.ember-view`, and it should be adding the class on `#ember-3095-modal`.

Thanks for your help, I really appreciate it and love this addon! We're using v0.7.2, FWIW.

simonihmig:
> We're using v0.7.2, FWIW.

This will probably explain it. I am not sure when this was added, but probably later. It *does* work with the current 1.0 alpha release! Should also work with the latest pre-alpha version (0.11.3). I suggest you try that with a more recent version!

ryanrishi:
Ahh, upgrading to `1.0.0-alpha.3` fixed this. Thanks for your help!

PR #210, opened 2017-01-22
Add input as available event for showValidationOn

elgordino:
I need to show validations as soon as the user presses a key so I added this action in addition to 'change' and 'focusOut'. Could you consider accepting this PR please.

Thanks!

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -501,7 +501,7 @@ export default FormGroup.extend({
 
   /**
    * Event or list of events which enable form validation markup rendering.
-   * Supported events: ['focusOut', 'change']
+   * Supported events: ['focusOut', 'change', 'keyPress']
    *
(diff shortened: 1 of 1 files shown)


simonihmig:
Thanks for contributing!

But what is your use case here that requires `keyPress`? Wouldn't the `change` event work as well?

elgordino:
Hi. I thought `change` would work too, but I find that it only adds the validation when the element loses focus AND the value has changed.

simonihmig:
Yes, that's how `change` works. I see you want the validation to display immediately. But I think `input` would then be the appropriate event (http://caniuse.com/#feat=input-event).

Would you mind changing your PR to `input`?

elgordino:
Thanks `input` is indeed more appropriate! I've updated the PR.

simonihmig:
Thanks again!

PR #213, opened 2017-01-24
update bs-accordion to use card classes on bs4

stevesims:


addon/templates/components/bs4/bs-accordion/item.hbs
```diff
@@ -0,0 +1,12 @@
+<div role="tab" {{action onClick value}} class="card-header {{if collapsed "collapsed" "expanded"}}">
```
srvance: I think we should also move the common accordion item to the bs3 tree since it references the `panel` classes.

addon/templates/components/bs4/bs-accordion/item.hbs
```diff
@@ -0,0 +1,12 @@
...
+    <h5 class="card-title mb-0">
```
srvance: What is the `mb-0` class? Is this something that should have some tunability?

addon/templates/components/bs4/bs-accordion/item.hbs
stevesims: ok cool - i'll do that shortly

addon/templates/components/bs4/bs-accordion/item.hbs
stevesims: `mb-0` applies `margin-bottom: 0`.  This follows the example in the Bootstrap 4 docs for their Accordion - without it the header ends up being too tall

addon/templates/components/bs4/bs-accordion/item.hbs
stevesims: Replaced the commit with one that has the common accordion item in the bs3 tree

simonihmig:
Nice, thanks @stevesims!

stevesims:
@srvance just wondering if you're happy with this now that i've moved the bs3 template?

srvance:
Great work, @stevesims! Thank you!

PR #215, opened 2017-01-25
BS4 test separation

simonihmig:
This is the result of a quick spike on how to do the separate tests story. The idea here is:

* you can override the bootstrap version setting (usually specified in `ember-cli-build.js`) via the environment variable `BOOTSTRAPVERSION`. 
* This is not to be used by normal users, but allows us to have a build with a specific version: `BOOTSTRAPVERSION=4 ember test`. This could be added to an ember-try scenario
* there is a test helper in `tests/helpers/bootstrap-test.js` that accepts an optional version number as first argument. If not specified runs test with any version (for tests that are the same regardless of BS version), if specified runs test if the version matches, skips otherwise
* See the slightly modified test file for `bs-button`

@srvance what do you think?


tests/dummy/config/environment.js
```diff
@@ -18,7 +18,8 @@ module.exports = function(environment) {
       // when it is created
     },
 
-    unstableApiURL: 'http://simonihmig.github.io/ember-bootstrap/'
+    unstableApiURL: 'http://simonihmig.github.io/ember-bootstrap/',
+    bootstrapVersion: process.env.BOOTSTRAPVERSION || 3
```
srvance: I'd pull this logic down into a conditional, both to restrict the environment and to make changing the version simple for users.

tests/helpers/bootstrap-test.js
```diff
@@ -0,0 +1,28 @@
...
+    default:
+      throw new Error('Unsupported invocation of testForBootstrap function');
+  }
+
+  if (bsVersion && bsVersion !== currentBootstrapVersion) {
+    // Skip test for different Bootstrap version
+    return;
+  }
+  return test(name, fn);
+}
```
srvance: Clever! I wondered if there was a "skip" version of `test` if you omit the function (like in RSpec), but it looks like not, at least according to the QUnit docs.

tests/integration/components/bs-button-test.js
```diff
@@ -1,19 +1,48 @@
...
 import hbs from 'htmlbars-inline-precompile';
 import Ember from 'ember';
 
 moduleForComponent('bs-button', 'Integration | Component | bs-button', {
   integration: true
 });
 
-test('button has correct CSS classes', function(assert) {
-  this.render(hbs`{{#bs-button type="primary" size="lg" active=true block=true}}Test{{/bs-button}}`);
+test(3, 'button has correct default markup', function(assert) {
```
srvance: I really like this as a solution. The only thing that bugs me a little about it is the mysterious magic number. What if we also exposed functions (e.g., `testBS3`) that curried the version in for better readability?

srvance:
This is a very clever approach. It can cover all of our cases and is simple to use. I like it.

In some of the tests where only the class changes, like with the transition from `in` to `show`, should we also have a helper that gives you finer control, e.g., `versionDependent(bs3Value, bs4Value)` or `versionDependent({3: bs3Value, 4: bs4Value})`? It probably needs a better name than my examples and wouldn't be as well tagged in the output.

index.js
```diff
@@ -34,7 +34,11 @@ module.exports = {
     }
     this.app = app;
 
-    let options = extend(defaultOptions, app.options['ember-bootstrap']);
+    let options = extend({}, defaultOptions, app.options['ember-bootstrap']);
+    if (process.env['BOOTSTRAPVERSION']) {
```
simonihmig: Would be possible somehow I think. But why would we need to do that? I would keep this feature really private, not meant to be used by users.

tests/dummy/config/environment.js
simonihmig: Not sure what exactly you mean? Like guarding against arbitrary values?

tests/helpers/bootstrap-test.js
simonihmig: There is this: https://api.qunitjs.com/QUnit.skip/
But if I correctly understand this it marks the test then as skipped (like pending, not yet implemented), but still outpouring it in the test suite. Not what we want, I think. When running tests for BS3, we can regard all BS4 specific tests as non existing, and vice versa. So this should do it. IMO

tests/integration/components/bs-button-test.js
simonihmig: Yes, that was my first idea. But the I decided for this one. Not sure why actually. :) 
If you feel that looks better, I can change that. Probably more readable indeed...

simonihmig:
Regarding the `versionDependent` thing: would be pretty easy to make that happen. But I am not so sure we really should. The next thing would maybe the desire to have the bootstrap version available as a var, to make conditionals etc. But that's putting too much into the tests I think. For the tests I think we can be a bit loose about things like DRY, and just duplicate. I case we drop BS3 some time, the tests can stay the same, no cleanup needed. 

Just my first thought though, I am certainly not dogmatic about that...

index.js
srvance: I usually prefer to safeguard against this kind of thing being triggered in production, especially when it's mainly there for testing support. It's not meant to be used, but it's in code that users will see, especially the exposure in `environment.js`.

tests/dummy/config/environment.js
srvance: This is the location where a user would define the version they want to use. Leave that simply as a number and put the environment variable override in the `if(environment ===` blocks. If users aren't supposed to use this, keep it away from the areas we expect them to modify.

tests/integration/components/bs-button-test.js
srvance: If you looked at the tests without being familiar with the way we've been thinking, this would look mysterious. I think it would make less of a learning curve for potential contributors if we spell it out.

simonihmig: Agreed.

tests/dummy/config/environment.js
simonihmig: But this is the config of the addon's dummy app, users should not see this. And unless they don't add the same logic into their own config file, they will not be able use that. It's a bit different for the `index.js`, as this also affects the user's app. But usually they won't look into that file really...

srvance: You're right. I forgot that part.

tests/helpers/bootstrap-test.js
srvance: Some of the other frameworks have an implicit skip if you define a test without a function, which would give us another way to have two arguments, but it's not an issue with qunit.

srvance:
Let's keep it simple for now. If we have too much duplication or boilerplate around the differences, we can address it then. As you pointed out above with the environment, this is really internal to the project.

simonihmig:
Just updated the test functions to use the more verbose `testBS3`/`testBS4`...

srvance:
Looks good to me.

simonihmig:
Thanks for reviewing, merged.

PR #220, opened 2017-01-30
fix how options are merged

stevesims:
`util._extend` doesn’t support more than two arguments for merging in values, which was meaning that only the default `ember-bootstrap` options would be used, with any overrides from the app being ignored

this fix ensures that the app’s `ember-bootstrap` options don’t get ignored

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,7 @@ module.exports = {
     }
     this.app = app;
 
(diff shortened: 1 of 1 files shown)


simonihmig:
Oups, changed that recently in that branch without really checking. Good catch, thanks!

PR #231, opened 2017-02-09
[bs4] use default button type for modal close button

bgentry:
Instead of explicitly specifying type="default", just let the bs-button component use its own default type. This ensures that the button gets the correct default classes in both Bootstrap 3 and 4.


simonihmig:
Makes sense, thanks!

srvance:
Good catch, @bgentry!

PR #233, opened 2017-02-10
[WIP] Assess & improve the blueprint

srvance:
The default blueprint is run when `ember-bootstrap` is installed and can also be run separately. Currently it installs all of the dependencies necessary for e-b. With the Bootstrap 4 support, there are more dependencies and permutations of Bootstrap versions and preprocessors.

This PR investigates whether it works correctly in those situations and explores where we should go from there. Phase 1 assesses whether the blueprint works correctly for the various upgrade paths needed to support #206 with 2 Bootstrap versions (3, 4) and 3 preprocessor options (none, LESS, SASS). Phase 2 investigates whether we can improve the overhead of the current solution by installing only the packages necessary to support the chosen configuration.

# Phase 1: Install a functional `ember-bootstrap`

The `bs4` branch approach is to install `bootstrap#^3.3.7` in `bower.json` and `bootstrap#^4.0.0-alpha.6` and `bootstrap-sass#^3.3.7` in `package.json` and use build-time tree manipulation based on the `bootstrapVersion` setting in `index.js`. This results in additional overhead to install `ember-bootstrap`, but exactly what is needed in the production distribution.

* `e-b` = ember-bootstrap (None, <bs4, bs4)
* `BS` = Bootstrap (None, 3, 4)
* `Pre` = CSS preprocessor (None, LESS, SASS)
  * `LESS` = Assumes `ember-cli-less`
  * `SASS` = Assumes `ember-cli-sass`

## Installation Scenarios

Normal usage when `ember-bootstrap` has not been installed invokes the blueprint during `ember install ember-bootstrap`.

From/To (e-b/BS/Pre) | bs4/3/None | bs4/3/LESS | bs4/3/SASS | bs4/4/None | bs4/4/SASS |
---- | ---- | ---- | ---- | ---- | ----
None/None/None | Add bootstrap3 in bower.json and bootstrap4 in package.json | No direct path | No direct path | No direct path | No direct path
None/None/LESS | No direct path | Add bootstrap3 in bower.json and bootstrap 4 in package.json; create app.less | No direct path | No direct path | No direct path
None/None/SASS | No direct path | No direct path | Add bootstrap-sass and bootstrap4 in package.json; create app.scss | No direct path | No direct path
None/3/None | Update bootstrap3 in bower.json and add bootstrap4 in package.json | No direct path | No direct path | No direct path | No direct path
None/3/LESS | No direct path | Update bootstrap3 in bower.json and add bootstrap 4 in package.json; append import to app.less | No direct path | No direct path | No direct path
None/3/SASS | No direct path | No direct path | Add bootstrap-sass and bootstrap4 in package.json; append import to app.scss | No direct path | No direct path
None/4/None | Downgrade to bootstrap3 in bower.json and add bootstrap4 in package.json | No direct path | No direct path | No direct path | No direct path
None/4/SASS | No direct path | No direct path | Add bootstrap-sass and update bootstrap4 in package.json; append import to app.scss | No direct path | No direct path

### Notes

* Anything installed in `bower.json` will be untouched except for updating the "bootstrap" version if applicable. Most notably, this includes "bootstrap-sass".
* The blueprint makes no modifications to `ember-cli-build.js`
* The blueprint makes the wrong assumptions if an alternate preprocessor is present.
* There is no direct route to change preprocessors or to use Bootstrap 4 via the blueprint.

## Configuration Change Scenarios

Once `ember-bootstrap` has been installed, the blueprint is invoked with `ember generate ember-bootstrap`.

From/To (e-b/BS/Pre) | bs4/3/None | bs4/3/LESS | bs4/3/SASS | bs4/4/None | bs4/4/SASS |
---- | ---- | ---- | ---- | ---- | ----
<bs4/3/None |  |  |  |  |  
<bs4/3/LESS |  |  |  |  |  
<bs4/3/SASS |  |  |  |  |  
bs4/3/None | No change |  |  |  |  
bs4/3/LESS |  | No change |  |  |  
bs4/3/SASS |  |  | No change |  |  
bs4/4/None |  |  |  | No change |  
bs4/4/SASS |  |  |  |  | No change

## Other configurations

The following and other packages aren't explicitly addressed at this time, although their effects are if they install the packages addressed above.

* `ember-cli-bootstrap-sassy`
* `ember-cli-bootstrap3-sass`
* `ember-cli-postcss`
* `ember-cli-css-preprocess`

# Phase 2: Optimize Installation

Install only the necessary bootstrap and preprocessor packages and automatically manage the `bootstrapVersion` in `index.js`.

### DIFF
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -83,7 +83,7 @@ module.exports = {
       let source = fs.readFileSync(file, 'utf-8');
       let build = new BuildConfigEditor(source);
       let newBuild = build.edit('ember-bootstrap', settings);
-      fs.writeFileSync(newBuild.code());
+      fs.writeFileSync(file, newBuild.code());
     } else {
       this.ui.writeLine(chalk.red(`Could not find ${file} to modify.`));
     }
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -24,8 +24,8 @@ module.exports = {
   description: 'Configure ember-bootstrap',
 
   availableOptions: [
-    { name: 'bootstrap-version', type: Number, default: 3, aliases: ['bv'] },
-    { name: 'target-preprocessor', type: String, aliases: ['tp'] }
+    { name: 'bootstrap-version', type: Number, default: 3, aliases: ['bootstrap', 'bv'] },
+    { name: 'preprocessor', type: String, aliases: ['pp'] }
   ],
 
   works: 'insideProject',
@@ -34,31 +34,31 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let { bootstrapVersion, targetPreprocessor } = option;
+    let { bootstrapVersion, preprocessor } = option;
     if (bootstrapVersion !== 3 && bootstrapVersion !== 4) {
       throw new SilentError('Bootstrap version must be 3 or 4');
     }
 
-    if (targetPreprocessor && !validPreprocessors.includes(targetPreprocessor)) {
+    if (preprocessor && !validPreprocessors.includes(preprocessor)) {
       throw new SilentError(`Valid preprocessors are: ${validPreprocessors.join(', ')}`);
     }
 
-    if (!targetPreprocessor) {
+    if (!preprocessor) {
       let dependencies = this.project.dependencies();
       if ('ember-cli-sass' in dependencies) {
-        targetPreprocessor = option.targetPreprocessor = 'sass';
+        preprocessor = option.preprocessor = 'sass';
       } else if ('ember-cli-less' in dependencies) {
-        targetPreprocessor = option.targetPreprocessor = 'less';
+        preprocessor = option.preprocessor = 'less';
       } else {
-        targetPreprocessor = 'none';
+        preprocessor = 'none';
       }
     }
 
-    if (bootstrapVersion === 4 && targetPreprocessor === 'less') {
+    if (bootstrapVersion === 4 && preprocessor === 'less') {
       throw new SilentError('You cannot use LESS with Bootstrap 4');
     }
 
-    this.ui.writeLine(chalk.green(`Installing for Bootstrap ${bootstrapVersion} using preprocessor ${targetPreprocessor}`));
+    this.ui.writeLine(chalk.green(`Installing for Bootstrap ${bootstrapVersion} using preprocessor ${preprocessor}`));
   },
 
   afterInstall(option) {
@@ -86,7 +86,7 @@ module.exports = {
   },
 
   adjustBootstrapDependencies(option) {
-    let { bootstrapVersion, targetPreprocessor } = option;
+    let { bootstrapVersion, preprocessor } = option;
     let dependencies = this.project.dependencies();
     let bowerDependencies = this.project.bowerDependencies();
     let promises = [];
@@ -102,7 +102,7 @@ module.exports = {
         promises.push(this.removePackageFromBowerJSON('bootstrap-sass'));
       }
       promises.push(this.addPackageToProject('bootstrap', bs4Version));
-    } else if (targetPreprocessor === 'sass') {
+    } else if (preprocessor === 'sass') {
       if ('bootstrap' in dependencies) {
         promises.push(this.removePackageFromProject('bootstrap'));
       }
@@ -121,14 +121,14 @@ module.exports = {
   },
 
   addPreprocessorStyleImport(option) {
-    let { targetPreprocessor } = option;
+    let { preprocessor } = option;
     let importStatement = '\n@import "ember-bootstrap/bootstrap";\n';
 
-    if (targetPreprocessor === 'none') {
+    if (preprocessor === 'none') {
       return;
     }
 
-    let extension = targetPreprocessor === 'sass' ? 'scss' : 'less';
+    let extension = preprocessor === 'sass' ? 'scss' : 'less';
 
     let stylePath = path.join('app', 'styles');
     let file = path.join(stylePath, `app.${extension}`);
@@ -147,7 +147,7 @@ module.exports = {
 
   addBuildConfiguration(option) {
     let file = 'ember-cli-build.js';
-    let { bootstrapVersion, targetPreprocessor } = option;
+    let { bootstrapVersion, preprocessor } = option;
     let settings = {
       bootstrapVersion
     };
@@ -156,7 +156,7 @@ module.exports = {
       settings.importBootstrapFont = false;
     }
 
-    settings.importBootstrapCSS = (targetPreprocessor === 'none');
+    settings.importBootstrapCSS = (preprocessor === 'none');
 
     if (fs.existsSync(file)) {
       this.ui.writeLine(chalk.green(`Added ember-bootstrap configuration to ${file}`));
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -111,13 +111,10 @@ module.exports = {
       }
       promises.push(this.addPackageToProject('bootstrap-sass', bs3Version));
     } else {
-      if ('bootstrap' in dependencies) {
-        promises.push(this.removePackageFromProject('bootstrap'));
-      }
       if ('bootstrap-sass' in dependencies) {
         promises.push(this.removePackageFromProject('bootstrap-sass'));
       }
-      promises.push(this.addBowerPackageToProject('bootstrap', bs3Version));
+      promises.push(this.addPackageToProject('bootstrap', bs3Version));
     }
 
     return rsvp.all(promises);
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -34,7 +34,7 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let bootstrapVersion = option.bootstrapVersion;
+    let bootstrapVersion = option.bootstrapVersion = parseInt(option.bootstrapVersion);
     let preprocessor = option.preprocessor;
 
     if (bootstrapVersion !== 3 && bootstrapVersion !== 4) {
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -11,28 +11,33 @@ const bs3Version = '^3.3.7';
 const bs4Version = 'next';
 
 module.exports = {
+  name: 'ember-bootstrap',
+
+  availableOptions: [
+    { name: 'bootstrap-version', type: Number, default: 3 }
+  ],
+
   normalizeEntityName() {
   },
 
-  afterInstall() {
-    return this.addDependencies()
-      .then(() => this.addPreprocessorImport());
+  afterInstall(options) {
+    return this.addDependencies(options)
+      .then(() => this.addPreprocessorImport())
+      .then(() => this.addBuildConfiguration(options));
   },
 
-  addDependencies() {
+  addDependencies(options) {
     let dependencies = this.project.dependencies();
 
-    let promises = [
-      this.addPackageToProject('bootstrap', bs4Version)
-    ];
-
-    if ('ember-cli-sass' in dependencies) {
-      promises.push(this.addPackageToProject('bootstrap-sass', bs3Version));
-    } else {
-      promises.push(this.addBowerPackageToProject('bootstrap', bs3Version));
+    if (options.bootstrapVersion === 4) {
+      return this.addPackageToProject('bootstrap', bs4Version);
     }
 
-    return rsvp.all(promises);
+    if ('ember-cli-sass' in dependencies) {
+      return this.addPackageToProject('bootstrap-sass', bs3Version);
+    }
+
+    return this.addBowerPackageToProject('bootstrap', bs3Version);
   },
 
   addPreprocessorImport() {
@@ -60,6 +65,25 @@ module.exports = {
         this.ui.writeLine(chalk.green(`Created ${file}`));
         return writeFile(file, importStatement);
       }
+    }
+  },
+
+  addBuildConfiguration(options) {
+    let file = 'ember-cli-build.js';
+    let includeFonts = options.bootstrapVersion === 3;
+    let config = `
+        'ember-bootstrap': {
+            importBootstrapTheme: false,
+            importBootstrapCSS: true,
+            importBootstrapFont: ${includeFonts},
+            bootstrapVersion: ${options.bootstrapVersion}
+        },`;
+
+    if (fs.existsSync(file)) {
+      this.ui.writeLine(chalk.green(`Added ember-bootstrap configuration to ${file}`));
+      return this.insertIntoFile(file, config, { after: 'new EmberApp(defaults, {' });
+    } else {
+      this.ui.writeLine(chalk.red(`Could not find ${file} to modify.`));
     }
   }
 };
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -92,23 +92,21 @@ module.exports = {
     let bowerDependencies = this.project.bowerDependencies();
     let promises = [];
 
+    if ('bootstrap' in bowerDependencies) {
+      promises.push(this.removePackageFromBowerJSON('bootstrap'));
+    }
+    if ('bootstrap-sass' in bowerDependencies) {
+      promises.push(this.removePackageFromBowerJSON('bootstrap-sass'));
+    }
+
     if (bootstrapVersion === 4) {
-      if ('bootstrap' in bowerDependencies) {
-        promises.push(this.removePackageFromBowerJSON('bootstrap'));
-      }
       if ('bootstrap-sass' in dependencies) {
         promises.push(this.removePackageFromProject('bootstrap-sass'));
-      }
-      if ('bootstrap-sass' in bowerDependencies) {
-        promises.push(this.removePackageFromBowerJSON('bootstrap-sass'));
       }
       promises.push(this.addPackageToProject('bootstrap', bs4Version));
     } else if (preprocessor === 'sass') {
       if ('bootstrap' in dependencies) {
         promises.push(this.removePackageFromProject('bootstrap'));
-      }
-      if ('bootstrap' in bowerDependencies) {
-        promises.push(this.removePackageFromBowerJSON('bootstrap'));
       }
       promises.push(this.addPackageToProject('bootstrap-sass', bs3Version));
     } else {
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -7,9 +7,16 @@ const path = require('path');
 const writeFile = rsvp.denodeify(fs.writeFile);
 const chalk = require('chalk');
 const BuildConfigEditor = require('ember-cli-build-config-editor');
+const SilentError = require('silent-error'); // From ember-cli
 
 const bs3Version = '^3.3.7';
 const bs4Version = 'next';
+
+const validPreprocessors = [
+  'none',
+  'less',
+  'sass'
+];
 
 module.exports = {
   name: 'ember-bootstrap',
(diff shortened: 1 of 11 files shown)


srvance:
@simonihmig What do you think of this as a proof of concept? I've added a `--bootstrap-version` option to the blueprint. Based on that, it decided which dependency to add. It also adds the configuration options to `ember-cli-build.js` using the Bootstrap version to customize the `bootstrapVersion` and `importBootstrapFont` settings.

Caveats and Thoughts:

1. You have to remove the entire configuration to re-run the blueprint with this approach. A real implementation would have to be more surgical than `insertIntoFile` allows, so it would have to be done with recast, like the route generator.
1. The Blueprint API only allows you to add to `bower.json` while you can both add and remove from `package.json`. If we put all bootstrap dependencies in `package.json`, then the blueprint could remove the unnecessary ones as well as add the necessary ones.
1. We could add flags to override the preprocessor and even install a new one if we wanted. Imagine something like `ember install ember-bootstrap --bootstrap-version=4 --preprocessor=sass --skip-preprocessor-installation` for someone who was using `postCSS` for their SASS processing already. Of course the defaults on all of these would be the current sensible behaviors.

simonihmig:
>  I've added a --bootstrap-version option to the blueprint. Based on that, it decided which dependency to add. 

So this is already aiming at "Phase 2" then, right?

>  A real implementation would have to be more surgical than insertIntoFile allows, so it would have to be done with recast, like the route generator.

Yeah, the way it is now it is very error prone. Would not add anything if the `before` string was not found (a different line break would be enough), or it would duplicate the `ember-bootstrap` key if already existing. This was my primary concern why I chose to install all dependencies for BS3 and BS4, so the process of changing the BS version can be a manual one, instead of doing that programmatically. What's your idea then, would you want to  go the `recast` route then?

> If we put all bootstrap dependencies in package.json, then the blueprint could remove the unnecessary ones as well as add the necessary ones.

The problem with this - I think we briefly touched this point before - is that you cannot have two different versions in `package.json` of the same package ('bootstrap' in this case). So when we install both versions as it is now, we can only install the one with bower and the other with npm (or both with bower, I think that would be possible as well). However if we go the "Phase 2" route and only install the really needed deps, we could drop bower and install bootstrap only with npm (which would be cool, as everybody is trying to get rid of bower)

The extensive table "Installation Scenarios" is only applicable to "phase 1", right? And it should pretty much match the current Blueprint, doesn't it?

If you prefer to go the "phase 2" route, there could be an alternative to the probably quite complicated recast implementation. What if we require the user to change the Bootstrap version by setting it manually in `ember-cli-build.js`, but then requiring to rerun the default blueprint. This way for the first installation the blueprint could install just the dependencies needed for BS3 (the default for now). When the user sets `bootstrapVersion` to 4 and reruns the blueprint, it would remove the unneeded and add the new dependencies. When the user forgets to rerun the blueprint, we should be able to detect that mismatch of `bootstrapVersion` and dependencies (in `index.js` when staring a build) and issue a warning message.

srvance:
> So this is already aiming at "Phase 2" then, right?

Correct. I think the way you came up with to load multiple packages solved the problems I saw when I ran the blueprints before.

> The extensive table "Installation Scenarios" is only applicable to "phase 1", right? And it should pretty much match the current Blueprint, doesn't it?

Yes. I needed to do that analysis for myself to make sure I understood the big picture to verify it was now behaving correctly.

> Yeah, the way it is now it is very error prone.

Definitely. This is just proof of concept. If we agree on this direction, I would do it with `recast`. It shouldn't be very difficult based on what I did with the route generator. The AST markers to find and modify the settings are pretty straightforward.

With this approach, we would advocate that people use the blueprint to switch configurations, so multiple versions wouldn't be necessary. The blueprint installs exactly what's necessary, with options to include the preprocessor or not. If we created an `ember-cli-remove-bower-package-polyfill` to parallel a PR to `ember-cli`, we wouldn't even need to restrict ourselves to `package.json`.

I think the more robust `recast` implementation will also be necessary to help people upgrading `ember-bootstrap`. I'm not terribly worried about the complexity of it now that I've gone through the learning curve once and know where to find good examples.

simonihmig:
So, if you are willing to lead that effort with `recast`, then I am all in. Regarding the remove from bower feature, it should be easy to add as it is just some JSON that needs to be read, modified and written back to disk. See this example: https://github.com/tomdale/ember-cli-addon-tests/blob/master/lib/utilities/pristine.js#L196-L205

srvance:
Sounds good. Let's take that direction, but as for bs4 readiness, I don't think this should be a blocker. The current blueprint is sufficient to get started.

srvance:
@simonihmig Here's what I'm thinking for the CLI design:

* `--bootstrap-version=[3|4]`
  * default: 3
  * Affects which bootstrap dependency to install
  * Constrains preprocessor choices
* `--preprocessor=[none|sass|less]`
  * default based on what's installed constrained by `bootstrap-version`
  * Affects which bootstrap dependency to install
  * Affects which `app.*` style file to create/modify
  * Overrides installed preprocessor and issues instructions for next steps
* `--install-preprocessor`
  * default `false`
  * Adds the preprocessor dependency
  * Should this take an argument like `[none|sass|less|postcss]` and default to `none`?
    * We don't have to support `postcss` immediately, but arguments leave that option open.
    * Alternatively, we could take a comma-separated list of preprocessor packages in addition to our aliases.
    * We could also support a `same` alias that just used the specified `preprocessor`. I'll have to look if you can have optional arguments to the parameter that would let us interpret the presence of the parameter with no argument that way.
  * Should this remove other preprocessors

Additional items:

* All dependencies in `package.json`
* We'll remove old dependencies where possible.
* In the first pass we can just issue instructions for the things we don't do automatically, like removing a dependency from `bower.json` or installing a preprocessor that isn't installed yet.
* I think these parameters are all we need to calculate the appropriate config, dependencies, and `app.*`.
* We'll only set the config properties that are necessary to be set or changed by the declared parameters.

simonihmig:
@srvance looks good. Some questions / ideas:

* having both `--preprocessor` and `--install-preprocessor` might lead to confusion. For example what happens when `--install-preprocessor`  is set but not `--preprocessor`? And what happens when setting `--preprocessor` to e.g. 'sass' (assume it is not installed already) but not adding `--install-preprocessor`, this would leave the app in a somewhat broken state (having a `app.scss` but no ember-cli-sass). Can't we drop `--install-preprocessor` and just assume that we should install the preprocessor (if not done already) when using `--preprocessor`? 
* some wild idea: the `ui` lib of ember-cli has some advanced features link `ui.prompt` to create interactive workflows (see https://github.com/ember-cli/console-ui and https://github.com/sboudrias/Inquirer.js). Maybe this could be used to add some guidance, and ask the user some questions that resemble the CLI options, like asking "Which version of Bootstrap do you want to use?" which would be a list-type question (https://github.com/sboudrias/Inquirer.js#question) with '3' and '4' as selectable options. Based on the answers one could provide some better guidance, like removing the 'less' option of the following preprocessor question when BS4 has been selected. Here is an example of an addon that uses this feature quite extensively: https://github.com/offirgolan/ember-addon-genie

srvance:
@simonihmig The thing I didn't communicate well is that it would provide a lot of feedback and argument checking. Invalid combinations would error out immediately in `beforeInstall`. The intent of `--preprocessor` is to make it so you don't have to have the preprocessor installed before running the blueprint. Perhaps `--preprocessor-target` is more accurate. I didn't want to get into what to do if someone wanted to use multiple preprocessors, migrate over to the new preprocessor incrementally, or just install the preprocessor after running the blueprint.

None of this is an argument for the proposal, just clarification. You bring up good points that it could be confusing. I really like the idea of it being interactive, or at least an interactive mode (`--interactive`, `-i`). I'd like it to also be fully scriptable, though, so I think sensible arguments are important. I'm also thinking perhaps we should make it a full command. Let me play with it some more and I'll see if I can come up with a sensible combination. Deferring `--install-preprocessor` is probably a good idea for now.

srvance:
@simonihmig What do you think of `--target-preprocessor` with messages letting people what they need to do manually? 

Then we would follow on with work for installing preprocessors and interactive mode as future increments.

simonihmig:
What would that do, just install the bootstrap deps based on this setting rather than checking which preprocessor is already installed (like now), without installing the preprocessor addons if required?

srvance:
It would do that, manipulate the `app.*` and give instructions on the remainder. Basically, it would decouple the preprocessor choice from the installed preprocessor. We would fall back to the installed if no selection were provided.

srvance:
@simonihmig How does this look? Just `--bootstrap-version` and `--target-preprocessor` with package removal and manual instructions.

srvance:
We could probably add some instruction feedback for `app.*` and some documentation in the guides, but I wanted to see what you thought so far.

I'd like to defer interactive mode and preprocessor installation for now.

blueprints/ember-bootstrap/index.js
```diff
@@ -6,60 +6,151 @@ const fs = require('fs');
...
+      }
       promises.push(this.addPackageToProject('bootstrap-sass', bs3Version));
     } else {
+      if ('bootstrap' in dependencies) {
+        promises.push(this.removePackageFromProject('bootstrap'));
+      }
+      if ('bootstrap-sass' in dependencies) {
+        promises.push(this.removePackageFromProject('bootstrap-sass'));
+      }
       promises.push(this.addBowerPackageToProject('bootstrap', bs3Version));
```
simonihmig: When installing only the really needed dependencies based on the options I think we can drop installing Bootstrap via bower, and use the npm package, as everybody is moving away form bower?

blueprints/ember-bootstrap/index.js
```diff
@@ -6,60 +6,151 @@ const fs = require('fs');
 const path = require('path');
 const writeFile = rsvp.denodeify(fs.writeFile);
 const chalk = require('chalk');
+const BuildConfigEditor = require('ember-cli-build-config-editor');
+const SilentError = require('silent-error'); // From ember-cli
```
simonihmig: As I am not aware of `silent-error`, what does it do?
Also if we do use it, we should add it explicitly to our dependencies.

blueprints/ember-bootstrap/index.js
```diff
@@ -6,60 +6,151 @@ const fs = require('fs');
...
+    let promises = [];
 
-    let promises = [
-      this.addPackageToProject('bootstrap', bs4Version)
-    ];
-
-    if ('ember-cli-sass' in dependencies) {
+    if (bootstrapVersion === 4) {
+      if ('bootstrap' in bowerDependencies) {
+        this.ui.writeLine(chalk.white.bgBlue('ACTION: Manually remove \'bootstrap\' from bower.json'));
```
simonihmig: What's the idea here? I think we should be able to do this automatically, even if there is no API for this in ember-cli.

simonihmig:
@srvance Looks very good so far! 

If we have just one option to specify the preprocessor, can we just name it `--preprocessor` instead of `--target-preprocessor`? And why not install the preprocessor right away, it's just a `addPackageToProject` call that is needed, or?

blueprints/ember-bootstrap/index.js
srvance: It's part of ember-cli, which is why I didn't declare it as a dependency. It displays the error message in red without the stack trace.

blueprints/ember-bootstrap/index.js
srvance: It seemed a good example of how we might provide feedback before we invested in removing bower packages. I'll look into the bower removal now.

blueprints/ember-bootstrap/index.js
srvance: I agree. That's one of the things I wanted to discuss with you. I'll make that change.

srvance:
I changed it to `--target-preprocessor` based on our discussion above in anticipation of having a separate option to install it. I thought this clarified that it wasn't necessarily installing it. I could also make `--preprocessor` an alias.

I'm conflicted about installing the preprocessor automatically. It's probably the right thing for the 80% case. In the remainder, some people use multiple preprocessors, some people need time to convert their usage, and some people will want to use an alternate preprocessor like postcss. For people converting from less to sass (and I suppose vice versa, but I see that as unlikely given where Bootstrap is going), should we also remove the old preprocessor? I certainly don't think we should remove the `app.less` if they're converting.

I thought this was a simple, backwards-compatible way to wait and see how people use it without blocking options, but I'm fine with being more progressive about it if you want.

If we do, we should probably use `addAddonToProject` where we know it's an ember-addon.

srvance:
We'll also need to change the way multi-version testing is set up.

blueprints/ember-bootstrap/index.js
srvance: Bower removal done.

blueprints/ember-bootstrap/index.js
srvance: Done.

simonihmig:
> I'm conflicted about installing the preprocessor automatically...

But wouldn't the other 20% of users just have to not use the `--(target-)preprocessor` at all? Because when they really use multiple preprocessors or postcss, can our blueprint then do anything useful other than saying _"Hey, we cannot figure out the right think to do here, you have to do this by yourself!"_ ?

So, when the user already uses a preprocessor or wants to use one, he can just specify `--preprocessor` and the blueprint will set things up and install the preprocessor addon if needed. But when they don't or they want to manage that on their own, they just skip the option?

> If we do, we should probably use addAddonToProject where we know it's an ember-addon.

Right

> We'll also need to change the way multi-version testing is set up.

Why that?

srvance:
While I was playing with the work so far, I was convincing myself we should cater to the 80%, as well. I'll go ahead and implement that. There's always `git reset` if they don't like what it does. You're right that the 20% probably don't need the help. If we're installing the preprocessor automatically, I'll shorten the option again.

We'll need to change the multi-version testing to run the blueprint before testing now rather than setting the environment variable. The environment variable only works because all of the dependent Bootstrap packages are already installed.

simonihmig:
> While I was playing with the work so far, I was convincing myself we should cater to the 80%, as well. I'll go ahead and implement that. There's always git reset if they don't like what it does. You're right that the 20% probably don't need the help. If we're installing the preprocessor automatically, I'll shorten the option again.

Great!

> We'll need to change the multi-version testing to run the blueprint before testing now rather than setting the environment variable. The environment variable only works because all of the dependent Bootstrap packages are already installed.

Make sense! 😬
Can we run the blueprint on our own addon, would that work?

srvance:
> Can we run the blueprint on our own addon, would that work?

Good question. Maybe we should keep them installed in the addon itself as `devDependencies`. I haven't tested the blueprint against the addon itself. I'll experiment with this.

srvance:
As long as we leave the dependencies in our `bower.json` and `package.json`, we don't need to change the testing strategy. I know the blueprint wouldn't find the dummy app's `ember-cli-build.js` properly right now.

I think the implementation is complete. I need to write some docs and test that we can invoke the blueprint options during `ember install`. I've verified all the permutations running the blueprint using `npm link`, but I need to pull it off the branch to install it with options.

srvance:
You can't pass blueprint options to `ember install`, so I'll just document the blueprint use.

srvance:
I think that covers it. It occurred to me that we could register an `ember bootstrap` command that could be used instead of the default blueprint. They could share most of the code. I'm find waiting on that to see if we get feedback.

simonihmig:
Saw the ember-cli issue you created. Would be cool if the `install` command would allow to pass the options, but I guess that will take some time to land, so we should not depend on that for now...

Some remarks:
* as we don't install any bootstrap assets from bower anymore, we can remove bootstrap from bower unconditionally (I think right now it would not get removed from bower.json for bs=3 and preprocessor=less|none, right?)
* we still import bootstrap assets from `bower_components` [here](https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L106) and [here](https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L111), which would need to be fixed. Probably because of this the FastBoot test is failing (the only tests that are affected by the blueprint changes, because those e2e tests actually do an `ember install ember-bootstrap` during testing)
* regarding the documentation, I would like to see the verbose version at http://www.ember-bootstrap.com/#/getting-started/assets, the README should be short and refer to the docs page for details
* I started to write some tests for the blueprint here https://github.com/kaliber5/ember-bootstrap/blob/master/node-tests/blueprints/ember-bootstrap-test.js. With the recent changes these are kind of obsolete, but we should add some tests probably for the new functionality. The tests were not working so far, because of some issue with the blueprint test helpers. But that should be fixed by now, will have to look at this again. 

Don't want to leave all the work here to you, if you want me to tackle some things just say so! I could for example add the blueprint tests, or maybe add the interactive/prompt feature to the blueprint? (which would make even more sense now because of the so far missing options support in `ember install`)

srvance:
I'll fix the dependency management, the imports, and the docs. Do you mind doing the blueprint tests? I saw them but wasn't sure how to handle some of the effects or how to interpret the note on the bug.

Do you want the interactive mode to be part of this PR or later? Also, what do you think of the idea of a command?

simonihmig:
> I'll fix the dependency management, the imports, and the docs. 

Great!

> Do you mind doing the blueprint tests?

Sure

> Do you want the interactive mode to be part of this PR or later? 

We can do that later, if you prefer.

> Also, what do you think of the idea of a command?

What would be the benefit? Sure, they command looks a bit nicer, like `ember bootstrap:configure` instead of `ember g ember-bootstrap`, which does not really tell what it does. But apart form that? And don't we get some problems because things like `addAddonToProject` are part of the Blueprint API, and not available as a command, are they?

srvance:
I didn't realize those weren't available in the command. Let's scratch that idea.

Let's wait on the interactive mode. It seems a nice independent increment.

srvance:
We won't see the impact of these changes on the tests that depend on `ember install ember-bootstrap` until we publish another version.

simonihmig:
@srvance The FastBoot tests were failing because there is no destructuring in Node 4, the latest commit should fix this. fyi

node-tests/blueprints/ember-bootstrap-test.js
```diff
@@ -28,74 +28,151 @@ function createStyleFixture(name) {
...
-      .then(() => emberGenerate(args, (file) => {
-        expect(file('app/styles/app.scss')).to.not.exist;
-        expect(file('app/styles/app.less')).to.not.exist;
-      }));
-  });
+  describe('import styles', function() {
+
+    describe('based on existing preprocessor', function() {
+
+      it('skips blueprint when no preprocessor present', function() {
```
srvance: It doesn't skip the blueprint. It just won't create an `app.{less|scss}`. There will still be other effects, like an impact on the package to import and the build configuration.

simonihmig: Sure. That was taken from the old existing tests, will rename.

simonihmig:
> Are you working on those?

Yes :) Already started with them, brought the kid to bed, and will finish them now...

simonihmig:
Lost quite some time debugging some strange behaviour... anyways, tests are there. I don't test so far for package removal, might add that later...

simonihmig:
Now it also checks that npm packages get uninstalled. With all the possible combinations this leads to quite a number of different scenarios. I am not testing really all possible ones, as this would get way too much. Even now there are quite a lot, already starting to loose the overview... ;) 

I hope it is somewhat clear what the scenario file defines. In `dependencies` `null` means to check for uninstallation. If you miss something there, feel free to add. Otherwise I hope this PR should be ready to be merged, is it?

blueprints/ember-bootstrap/index.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let bootstrapVersion = option.bootstrapVersion;
+    let bootstrapVersion = option.bootstrapVersion = parseInt(option.bootstrapVersion);
```
srvance: I'm surprised this is necessary unless it's defensive. The docs and the blueprint insert it as a number. We should add the `radix`, as well.

node-tests/blueprints/dependencyScenarios.js
```diff
@@ -0,0 +1,586 @@
+'use strict';
+
+const bs3Regex = '\\^3\\.\\d+\\.\\d+';
+const bs4Regex = 'next';
+const scenarios = [
+  {
+    dependencies: {
+      bower: {
+        bootstrap: null
```
srvance: Let's add `"bootstrap-sass": null` to catch legacy scenarios.

node-tests/blueprints/dependencyScenarios.js
```diff
@@ -0,0 +1,586 @@
+'use strict';
+
+const bs3Regex = '\\^3\\.\\d+\\.\\d+';
+const bs4Regex = 'next';
+const scenarios = [
+  {
+    dependencies: {
```
srvance: Perhaps we could name the scenarios to make them more intent-revealing and use that name in the tests.

node-tests/blueprints/dependencyScenarios.js
```diff
@@ -0,0 +1,586 @@
+'use strict';
+
+const bs3Regex = '\\^3\\.\\d+\\.\\d+';
+const bs4Regex = 'next';
+const scenarios = [
+  {
```
srvance: Nice design! So this first one is equivalent to `installed: {}, options: {}`, correct? If you don't like the name idea below, we should consider being explicit on the empty definition sections.

node-tests/blueprints/dependencyScenarios.js
```diff
@@ -19,9 +15,16 @@ const scenarios = [
...
-        bootstrap: null
-      },
+      npm: {
+        bootstrap: bs3Regex
+      }
+    }
+  },
+  {
+    installed: {
+      npm: ['ember-cli-less', 'bootstrap-sass']
```
srvance: Odd combination, isn't it? Intentionally unusual?

srvance:
I think this is fantastic. We can always improve the tests as we see the need. This gives us great coverage to start. I really like how we've collaborated on this to get a result better than either of us would have probably achieved on our own.

The test failure is strange, complaining about recursive links. Is npm trying to optimize the packages with links and messing itself up?

node-tests/blueprints/dependencyScenarios.js
simonihmig: Yes, good catch. No idea what my thought was, probably just too late in the evening... :)

simonihmig:
> I think this is fantastic. We can always improve the tests as we see the need. This gives us great coverage to start. I really like how we've collaborated on this to get a result better than either of us would have probably achieved on our own.

Confirm! 👍

> The test failure is strange, complaining about recursive links. Is npm trying to optimize the packages with links and messing itself up?

Yeah, that is strange. Had no problems whatsoever on my local machine (also testing with node 4 fwiw). Once cleared the Travis cache, then it did run fine, but now breaking again... 😳

simonihmig:
The weird installation errors in the blueprint test run should be solved by https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/98

srvance:
I tried rerunning the failing job, but same result. Any ideas?

simonihmig:
@srvance Travis had some other issues today, but after these have been sorted out apparently and I cleared its cache, the build is now green!

Should we merge?

jonnii:
Would be awesome to get an alpha cut with this, it would mean we could drop our entire `bower.json` file!

PR #234, opened 2017-02-11
Form element components

simonihmig:
@srvance Have a look at this. One could still add some more component tests for the new (private) components (e.g. controls), but all essential things should be covered by the existing FormElement test, so should hopefully be ok to merge. There will probably be some merge conflicts with the bs4 branch, I can resolve them once merged to master...


simonihmig:
Btw, the failed build comes from FastBoot tests not able to run with node 4, because of https://github.com/ember-fastboot/fastboot/pull/125. So not related to any code changes here...

srvance:
A fair amount of internal complexity but it preserves the external simplicity. I think this is probably the most complex part of the addon. From a quick inspection, it looks like a nice set of abstractions for the problem. I'm going to make a couple of minor code review comments, but otherwise, let's go forward with it. I think it also provides all the control points we'll need for the bs4 adaptations. We're due for a merge to the bs4 branch anyway.

addon/components/bs-form/element/control/checkbox.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+
+ @class FormElementControlCheckbox
+ @namespace Components
+ @extends Components.FormElementControl
+ @private
+ */
+export default Control.extend({
+  attributeBindings: [
```
srvance: This could get ugly with the global attributes, as well. Same issue we discussed with `nav.anchor`. I think we need to live with it in this case. I considered enhancing `attributeBindings` to allow a hash usage that would allow attribute passthrough, but decided not to go there. But my investigations led me to write the missing documentation for it.

addon/templates/components/bs-form/element.hbs
```diff
@@ -1 +1,65 @@
-{{partial formElementTemplate}}
\ No newline at end of file
+{{#component
```
srvance: Just for formatting, can you bring `layoutComponent` up onto the same line to differentiate the variable with the component being invoked from all the arguments being passed to it?

addon/templates/components/bs-form/element.hbs
```diff
@@ -1 +1,65 @@
...
+  {{#if hasBlock}}
+    {{yield
+      (hash
+        value=value
+        id=formElementId
+        validation=validation
+      )
+    }}
+  {{else}}
+    {{component
```
srvance: Same here.

addon/templates/components/bs-form/element.hbs
simonihmig: done

simonihmig:
@srvance I merged this. But another thing: right now al these new component are declared as private, as they are only used by `bs-form/element` internally. However I was asking myself if it would make sense to make the controls (input, textarea, checkbox) public, like `bs-input` (which was existing before, but was just extending the two-way only `{{input}}` of Ember itself). This way one could use these as standalone (one-way) components where the functionality of a form-element might not be needed. Your thoughts? 

On the other hand I would not want to g as gar as one day duplication the functionality of [ember-one-way-controls](https://github.com/DockYard/ember-one-way-controls). 

We can still go that route any day though, when the need arises.

PR #236, opened 2017-02-14
Modal header components

simonihmig:
Allows for easier customization.


simonihmig:
@srvance I guess that needs another manual merge into `bs4`... is this looking ok to you, regarding BS4 integration?

srvance:
I had to move the close button after the title for the bs4 version of the header. If we can do that for both bs3 and bs4, it will simplify things. Keep this in mind as you merge. Otherwise, this looks good.

srvance:
Oh, and there's also a change of the title tag from an `h4` to an `h5`.

simonihmig:
Ok, thanks. should be no problem then...

PR #246, opened 2017-02-21
Bs4 validation feedback

srvance:
I got the validation working and with the exception of some property passing in `element.hbs`, it's completely differentiated between BS3 and BS4.

BS4 makes a distinction between validation feedback text (`form-control-feedback`) and form help text (`form-text`). We don't account for the latter right now. Someone can add it in, but they won't have control over where it's rendered relative to the feedback text.

Also, I notice in both BS3 and BS4, the dummy app fields that show validation don't validate until you leave the first field. Once you do that, they validate on every keystroke.


addon/templates/components/bs4/bs-form/element/layout/horizontal.hbs
```diff
@@ -0,0 +1,12 @@
...
+  <div class="{{horizontalInputGridClass}}">
+    {{yield}}
+    {{component errorsComponent}}
+  </div>
+{{else}}
+  <div class="{{horizontalInputGridClass}} {{horizontalInputOffsetGridClass}}">
+    {{yield}}
+    {{component errorsComponent}}
+  </div>
+{{/if}}
```
simonihmig: If I am not missing anything, the only change here (and also in `vertical.hbs`) is the missing feedback icon component. Instead of duplicating the template one could also set `feetbackIconComponent` to `null` in bs4/bs-form/element.js. I think the `{{component}}` helper will then simply render nothing.

srvance:
I would expect validation feedback from the first keystroke.

addon/templates/components/bs4/bs-form/element/layout/horizontal.hbs
srvance: That works

srvance: Actually it doesn't. Throws an assertion and renders incompletely.

simonihmig: Ok, good to know. Thought it worked like this at some point, but might confuse it...

simonihmig:
> I would expect validation feedback from the first keystroke.

There are different event types configurable, but the default is `focusOut`: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form/element.js#L548

You could use `input` to get what you want...

Issue #251, opened 2017-03-01
Dropdown component doesn't expose isOpen


theroncross:
I'd like to be able to customize the dropdown-toggle content to depend on the `isOpen` state of the menu. This seems like the desired behavior:
```
{{#bs-dropdown as |dd|}}
  {{#dd.toggle}}
    {{icon-component name=(if dd.isOpen "arrow-up" "arrow-down")}}
  {{/dd.toggle}}
{{/bs-dropdown}}
```

PR #252, opened 2017-03-01
Adds isOpen to dropdown template hash

theroncross:
towards #251

### DIFF
--- a/tests/integration/components/bs-dropdown-test.js
+++ b/tests/integration/components/bs-dropdown-test.js
@@ -77,6 +77,16 @@ test('clicking dropdown menu when closeOnMenuClick is false will not close it', 
   assert.equal(this.$(':first-child').hasClass(openClass()), true, 'Dropdown is open');
 });
 
+test('child components can access isOpen property', function(assert) {
+  this.render(hbs`{{#bs-dropdown as |dd|}}{{#dd.toggle}}<span id="toggleText">{{if dd.isOpen "open" "closed"}}</span>{{/dd.toggle}}{{/bs-dropdown}}`);
+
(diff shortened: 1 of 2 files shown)


tests/integration/components/bs-dropdown-test.js
```diff
@@ -77,6 +77,19 @@ test('clicking dropdown menu when closeOnMenuClick is false will not close it',
   assert.equal(this.$(':first-child').hasClass(openClass()), true, 'Dropdown is open');
 });
 
+test('child components can access isOpen property', function(assert) {
+  this.render(hbs`{{#bs-dropdown as |dd|}}{{#dd.toggle}}<span id="toggleText">{{if dd.isOpen "open" "closed"}}</span>{{/dd.toggle}}{{#dd.menu}}<li><a href="#"><span id="linkText">{{if dd.isOpen "open" "closed"}}</span></a></li>{{/dd.menu}}{{/bs-dropdown}}`);
+
+  assert.equal(this.$('#toggleText').text(), 'closed', 'Dropdown is closed');
+  assert.equal(this.$('#linkText').text(), 'closed', 'Dropdown is closed');
```
simonihmig: Looks fine, but the tests for `#linkText` are not really needed, could you please remove these parts. If it works with `#toggleText` it will as well with `#linkText`, otherwise it would be an Ember bug, and I don't want to test for those! ;) 

So a template like this would be enough for testing:

```hbs
{{#bs-dropdown as |dd|}}<span id="toggleText">{{if dd.isOpen "open" "closed"}}</span>{{/bs-dropdown}}
```

simonihmig:
@theroncross Sorry for the delay, I did not notice you updated your PR already. Thanks four your contribution!

PR #256, opened 2017-03-03
BS dropdown initial state

estshy:
Open property is a part of public API but changing it to `true` as initial value will break the component as event for closing it won't be attached to body.

Fix applies only to `v0.11.3` as newest release has made it private for unknown reason ;).


simonihmig:
Your PR has conflicts probably because your PR targets master, which is actually the latest 1.0.0-alpha version, but as you said you are targeting 0.11.3, and for the latest version your code changes don't seem to apply. But actually I don't want to support the older 0.x range with updates, so this is probably not going to be merged, sorry!

PR #260, opened 2017-03-11
Fix BS4 form label class and update tests.

shuriu:
BS4 uses `col-form-label` class for form labels.

### DIFF
--- a/addon/components/bs4/bs-form/element/label.js
+++ b/addon/components/bs4/bs-form/element/label.js
@@ -3,7 +3,7 @@ import FormElementLabel from 'ember-bootstrap/components/base/bs-form/element/la
 export default FormElementLabel.extend({
   tagName: 'label',
 
-  classNames: ['form-control-label'],
+  classNames: ['col-form-label'],
   classNameBindings: ['invisibleLabel:sr-only', 'labelClass'],
   attributeBindings: ['formElementId:for']
 });
--- a/tests/integration/components/bs-form/element/label-test.js
+++ b/tests/integration/components/bs-form/element/label-test.js
@@ -1,4 +1,5 @@
 import { moduleForComponent, test } from 'ember-qunit';
+import { testBS3, testBS4 } from '../../../../helpers/bootstrap-test';
 import hbs from 'htmlbars-inline-precompile';
 
 moduleForComponent('bs-form/element/label', 'Integration | Component | bs form/element/label', {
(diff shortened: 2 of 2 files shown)


addon/components/bs4/bs-form/element/label.js
```diff
@@ -3,7 +3,7 @@ import FormElementLabel from 'ember-bootstrap/components/base/bs-form/element/la
 export default FormElementLabel.extend({
   tagName: 'label',
 
-  classNames: ['form-control-label'],
+  classNames: ['col-form-label'],
```
srvance: From what you told me, we should only have `col-form-label` for horizontal layouts. We still need `form-control-label` for vertical and inline layouts. The [form validation section](https://v4-alpha.getbootstrap.com/components/forms/#validation) seems inconsistent. The "How It Works" section only mentions `col-form-label`. The first example uses `form-control-label` and the second example uses `col-form-label`. Can we either check the BS4 source or ask the maintainers in an issue?

If it can be either, we should create something like a `formLabelClass` in `classNameBindings` that is a computed property returning a string for the proper class name.

srvance:
@shuriu Thank you for noticing and doing this!

I think you've found an area where the BS4 documentation is either inaccurate or incomplete.

addon/components/bs4/bs-form/element/label.js
shuriu: You're right about the BS docs.

My testing and looking into the BS source indicates that `.col-form-label` is the way going forward because it handles both vertical and horizontal forms, for margins, padding and validations. For inline forms, it is not an issue because the `.form-inline` class on the form handles this.

If this isn't the case then I'll update accordingly.

shuriu:
@srvance I've asked on the BS official channel for a confirmation and no answer as of now. If nothing comes out of this I'll open an issue to further clarify, because it is indeed inaccurate or incomplete.

shuriu:
I've opened an issue on the BS project for this: https://github.com/twbs/bootstrap/issues/22173.

srvance:
@shuriu It appears they've resolved the issue, although the changes haven't been released yet.

shuriu:
@srvance on a recent investigation I realised that `form-control-label` is actually needed in the case of checkboxes / radios, but I unfortunately don't have time to update this PR. So I think this PR should be deprecated / closed.

In any case, I saw they overhauled the whole validation system so I guess there might be even more changes once the beta drops.

simonihmig:
In favor of closing this, until BS4 has settled regarding its form markup, and the beta is out. @shuriu thanks anyways for working on this!

PR #263, opened 2017-03-16
Filter trees before calling _super.

rwjblue:
This fixes issues with ember-cli@2.12, and works on prior versions of ember-cli. Simply put, the concept that we are trying to do here is to filter down the set of things in the tree before we process them.

This will also have the positive result of not transpiling all of the compoents and templates for both bootstrap versions, just before removing one of them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,25 +157,27 @@ module.exports = {
     return this.getBootstrapVersion() === 3 ? 4 : 3;
   },
 
-  treeForAddon() {
-    let tree = this._super.treeForAddon.apply(this, arguments);
+  treeForAddon(tree) {
     let bsVersion = this.getBootstrapVersion();
     let otherBsVersion = this.getOtherBootstrapVersion();
-    let componentsPath = 'modules/ember-bootstrap/components/';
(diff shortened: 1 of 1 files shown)


rwjblue:
@simonihmig / @srvance - Please test, but this should resolve the issue being reported around ember-cli@2.12.0.

simonihmig:
@rwjblue Thanks a lot for working on this! I am checking this with the reproduction repo. Has just failed, but with some other error, might be not related. Will check this and report back. Thanks again!

rwjblue:
Awesome, thanks for testing. I also used the demo repo, but noticed I had to run the blueprint to get it to run. Perhaps related to a newer version? I'm unsure that would be related to this particular issue or not. Either way I'll be around to help tonight, please let me know if you run into more issues...

simonihmig:
Yes, exactly. We merged some changes regarding the blueprint and `index.js` to master just some hours ago, so this was affecting the demo repo. Fixed this, and new build is all green! 👍
https://travis-ci.org/simonihmig/ember-boostrap-test-app/builds/211944799

PR #265, opened 2017-03-17
Check for outdated dependencies

simonihmig:
This throws an error when running the build when bootstrap is still a bower dependency. #233 switched the BS3
dependency from bower to npm. The build will fail, if a project just updated ember-bootstrap without rerunning the
blueprint. This will detected this situation and issue an error message.

@srvance I noticed this issue when testing the latest ember-cli fix. After just updating ember-bootstrap, the build failed because it was trying to import the Bootstrap font from `node_modules/bootstrap` which was not present (still in `bower_components`)

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ const stew = require('broccoli-stew');
 const mv = stew.mv;
 // const log = stew.log;
 const rm = stew.rm;
-const chalk = require('chalk');
+const SilentError = require('silent-error'); // From ember-cli
 
 const defaultOptions = {
   importBootstrapTheme: false,
@@ -42,6 +42,7 @@ module.exports = {
     }
     this.bootstrapOptions = options;
 
+    this.validateDependencies();
     this.preprocessor = this.findPreprocessor();
 
     if (!this.hasPreprocessor()) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -66,6 +67,12 @@ module.exports = {
     }
   },
 
+  validateDependencies() {
+    if ('bootstrap' in this.app.project.bowerDependencies()) {
```
srvance: Should we also do a check for `bootstrap-sass`?

index.js
```diff
@@ -68,7 +69,7 @@ module.exports = {
   },
 
   validateDependencies() {
-    if ('bootstrap' in this.app.project.bowerDependencies()) {
+    if (EmberApp.env() !== 'test' && 'bootstrap' in this.app.project.bowerDependencies()) {
```
srvance: Why do we need this?

simonihmig: It would throw the error (well, with the latest commit, it's just a warning) when running inside of our addon itself. Because it still has bootstrap in bower.json. But after think about this again, we actually can remove it from our own bower.json, right? Then I can drop this of course...

index.js
simonihmig: But `bootstrap-sass` is a legit dependency (for bs3/sass)? The new blueprint removes Bootstrap from bower.json unconditionally, so we can use this fact to check if the user has run the new blueprint already. If not (bootstrap still in bower), we can warn him.

index.js
srvance: I'm always wary of code conditional to test environments.

index.js
srvance: We also remove `bootstrap-sass` from bower.json unconditionally and install it in package.json. No dependencies should be in bower.json once they've upgraded.

simonihmig: @srvance you're right, will update!

srvance:
Looks great!

PR #266, opened 2017-03-17
FormElement shows automatic validation state only when validator addon is present

### LINKED ISSUE #240: Enable validations only when validation addon is installed
Currently `hasValidator` is true when having `ember-cp-validations` installed, but `ember-bootstrap-cp-validations` is missing. Maybe `hasValidator` has just to be defined false in `ember-bootstrap` and overriden in the validation addons (`ember-bootstrap-cp-validations` is already doing that, `ember-bootstrap-changeset-validations` and `ember-bootstrap-validations`need to be updated though)

simonihmig:
Fixes #240

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -415,7 +415,7 @@ export default FormGroup.extend({
    *
    * @property errors
    * @type array
-   * @private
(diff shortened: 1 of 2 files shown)


srvance:
Looks good to me.

PR #267, opened 2017-03-17
Fix Dropdown with a block=true toggle button

### LINKED ISSUE #244: Change dropdown class "btn-group" to "dropdown"
I use Dropdown component,  with button block=true.

simonihmig:
Fixes #244

### DIFF
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -154,7 +154,7 @@ export default Ember.Component.extend({
    * @private
    */
   containerClass: computed('toggle.tagName', 'direction', function() {
-    if (this.get('toggle.tagName') === 'button') {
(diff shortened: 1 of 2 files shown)


addon/components/base/bs-dropdown.js
```diff
@@ -154,7 +154,7 @@ export default Ember.Component.extend({
    * @private
    */
   containerClass: computed('toggle.tagName', 'direction', function() {
-    if (this.get('toggle.tagName') === 'button') {
+    if (this.get('toggle.tagName') === 'button' && this.get('toggle.block') !== true) {
```
srvance: Either `!this.get('toggle.block')` or `this.get('toggle.block') === false`? The former is more permissive and would require the property to be initialized reliably.

PR #268, opened 2017-03-18
Add info command & issue template

simonihmig:
The `ember bootstrap:info` command prints something like this, useful for github issues:

```
Bower packages:
bootstrap: n/a
Npm packages:
ember-cli: ^2.10.0 -> 2.10.0
bootstrap: ^4.0.0-alpha.6 -> 4.0.0-alpha.6
bootstrap-sass: ^3.3.7 -> 3.3.7
ember-cli-sass: ^6.1.1 -> 6.1.1
ember-cli-less: n/a
```

@srvance Would be cool if we could also print the e-b config from `ember-cli-build.js`. As this is not just JSON, it is not so easily parseable. Would `ember-cli-build-config-editor` be able to extract that information?

### DIFF
--- a/lib/tasks/info.js
+++ b/lib/tasks/info.js
@@ -9,6 +9,7 @@ const bowerVersion = require('../utils/bower-version');
 const buildConfiguration = require('../utils/build-configuration');
 
 const npmPackages = [
+  'ember-bootstrap',
   'ember-cli',
   'bootstrap',
   'bootstrap-sass',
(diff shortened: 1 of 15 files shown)


srvance:
Should be able to do that pretty easily.

lib/tasks/info.js
```diff
@@ -0,0 +1,50 @@
...
+    npmPackages.forEach((pkg) => {
+      let version = npmVersion(pkg);
+      let installedVersion = version ? npmInstalledVersion(pkg) : undefined;
+      this.printVersion(pkg, version, installedVersion);
+    });
+  },
+
+  printVersion(module, version, installedVersion) {
+    if (version && installedVersion) {
+      this.ui.writeLine(`${module}: ${version} -> ${installedVersion || 'n/a'}`);
```
srvance: I don't think we need the 'n/a' on this line.

lib/utils/bower-version.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const path = require('path');
+const fs = require('fs');
+
+module.exports = function bowerVersion(packageName) {
+  let bowerJsonPath = path.join(process.cwd(), 'bower.json');
```
srvance: There's a way to get the project root from the Project object that may be better than `cwd()`. I was thinking of using it to improve the blueprint. I don't remember where I saw it, though.

lib/utils/npm-version.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const path = require('path');
+
+module.exports = function npmVersion(packageName) {
+  let pkgJson = require(path.join(process.cwd(), 'package.json'));
```
srvance: Same here.

simonihmig:
@srvance Awesome, thanks!

srvance:
This looks good to me. Is there more you want to add to it? The 'n/a' question above is an optimization. Keeping it is certainly safe.

simonihmig:
> The 'n/a' question above is an optimization. Keeping it is certainly safe.

Just forgot about it. I will remove.

Other than that, this should be ok. I just wanted to merge this just before releasing, so users that want to report issues don't see that issue template that would talk about a still non-existing command! ;)

PR #273, opened 2017-03-20
Fix wrong config being added to `ember-cli-build.js` 

simonihmig:
…when no preprocessor was used

@srvance `option.preprocessor` was checked for being `'none'` at some places, while it was just `undefined`. This lead to `importBootstrapCSS` being written as `false` when no preprocessor was used. Unfortunately I broke some blueprint tests when refactoring things, so they did not catch this.

Fixed this, and will merge and release once tests have passed, so hopefully not too many users are affected by this...

### DIFF
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -34,7 +34,7 @@ module.exports = {
   },
 
   beforeInstall(option) {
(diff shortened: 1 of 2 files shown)


srvance:
Good catch, @simonihmig. Thank you.

PR #275, opened 2017-03-20
Added ember-cli-template-lint

simonihmig:
Added template linting. No in-depth review is needed I think, as it is mostly indentation updates (done automatically for the most part)...


.template-lintrc.js
```diff
@@ -0,0 +1,9 @@
+/* jshint node:true */
```
srvance: Should we also have `/* eslint-env node */`?

PR #281, opened 2017-03-22
Use configured `bootstrapVersion` if present

### LINKED ISSUE #276: Can't seem to get bootstrapVersion: 4 to trigger the generator to install version 4
It just keeps switching it back to bootstrapVersion: 3

srvance:
Remove the `--bootstrap-version` option default because there is no way to distinguish between an omitted and defaulted value. Use the configured `bootstrapVersion` as the first fallback if the `--bootstrap-version` option is not supplied and fall back to "3" as a last resort.

Fixes #276

### DIFF
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -24,7 +24,7 @@ module.exports = {
   description: 'Configure ember-bootstrap',
 
   availableOptions: [
-    { name: 'bootstrap-version', type: Number, default: 3, aliases: ['bootstrap', 'bv'] },
+    { name: 'bootstrap-version', type: Number, aliases: ['bootstrap', 'bv'] },
     { name: 'preprocessor', type: String, aliases: ['pp'] }
   ],
 
@@ -34,7 +34,8 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let bootstrapVersion = parseInt(option.bootstrapVersion) || 3;
+    let configuredBootstrapVersion = this.retrieveBootstrapVersion();
+    let bootstrapVersion = parseInt(option.bootstrapVersion) || configuredBootstrapVersion || 3;
     let preprocessor = option.preprocessor;
 
     if (bootstrapVersion !== 3 && bootstrapVersion !== 4) {
@@ -202,5 +203,15 @@ module.exports = {
       let settingsString = JSON.stringify(settings);
       this.ui.writeLine(chalk.red(`Configuration file could not be edited. Manually update your ember-cli-build.js to include '${this.name}': ${settingsString}`));
     }
+  },
+
+  retrieveBootstrapVersion() {
+    let file = 'ember-cli-build.js';
+
+    let source = fs.readFileSync(file);
+    let build = new BuildConfigEditor(source);
+    let config = build.retrieve(this.name);
+
+    return config && config.bootstrapVersion;
   }
 };
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -34,8 +34,7 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let configuredBootstrapVersion = this.retrieveBootstrapVersion();
-    let bootstrapVersion = parseInt(option.bootstrapVersion) || configuredBootstrapVersion || 3;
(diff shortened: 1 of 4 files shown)


blueprints/ember-bootstrap/index.js
```diff
@@ -202,5 +203,15 @@ module.exports = {
...
+  },
+
+  retrieveBootstrapVersion() {
+    let file = 'ember-cli-build.js';
+
+    let source = fs.readFileSync(file);
+    let build = new BuildConfigEditor(source);
+    let config = build.retrieve(this.name);
+
+    return config && config.bootstrapVersion;
```
simonihmig: As we are checking for strict equality of a number, so maybe to increase fault tolerancy this should be `config && parseInt(config.bootstrapVersion)`?

blueprints/ember-bootstrap/index.js
```diff
@@ -34,7 +34,8 @@ module.exports = {
   },
 
   beforeInstall(option) {
-    let bootstrapVersion = parseInt(option.bootstrapVersion) || 3;
+    let configuredBootstrapVersion = this.retrieveBootstrapVersion();
+    let bootstrapVersion = parseInt(option.bootstrapVersion) || configuredBootstrapVersion || 3;
```
simonihmig: You could inline `this.retrieveBootstrapVersion()` here instead of assigning it first to a variable, so it gets lazily evaluated only when needed. When we have `option.bootstrapVersion` set, we don't need to open and parse that file.

node-tests/blueprints/ember-bootstrap-test.js
```diff
@@ -182,8 +184,8 @@ describe('Acceptance: ember generate ember-bootstrap', function() {
 
   describe('install dependencies', function() {
 
-    const Blueprint = require('ember-cli/lib/models/blueprint');
-    const origTaskFor = Blueprint.prototype.taskFor;
+    let Blueprint = require('ember-cli/lib/models/blueprint');
```
simonihmig: Just bikeshedding here, but I would prefer `const` when `require`ing. The second `let` makes sense though

simonihmig:
Great work overall! 

Regarding `this.retrieveBootstrapVersion()`, is it really necessary to use that. There is no way to get the eval-ed config directly somehow from `this.project` or some other ember-cli API? Because as in the other issue, we might not be able to parse that successfully using the BuildConfigEditor...

Absolutely great that you were able to mock the build config in the tests!

Not exactly the same problem, but still related: I just noticed when updating e-b in a app, that it also overrode `importBootstrapFont` from `false` to `true`. Would be great if we could make sure that we set some config items only if they were not defined before!?

node-tests/blueprints/ember-bootstrap-test.js
srvance: Eslint complains about const in a block.

blueprints/ember-bootstrap/index.js
srvance: I wondered why you used `parseInt` on a number. Is that to convert an accidental string?

blueprints/ember-bootstrap/index.js
srvance: Good point.

node-tests/blueprints/ember-bootstrap-test.js
simonihmig: Ok, we could put an ignore comment here, but never mind, not so important...

blueprints/ember-bootstrap/index.js
simonihmig: Yes, there was some issue, but I don't remember what exactly that was 🤔

simonihmig: At least here users could manually set this to a string value like `'4'`...

srvance:
Unfortunately, it doesn't seem that the build configuration is passed to the blueprint.

Conditional override is something I had on the TODO list for e-c-b-c-e (hey, it's a palindrome!). I removed it with the introduction of `retrieve` because we can leave it in the consumer's hands. We could do that pretty easily. I mainly wanted to turn it off for BS4 and figured that the other way around would get caught diff'ing the results, but we should be friendlier than that. I'll create a separate issue so it doesn't block this one.

simonihmig:
Ok, agreed. So this should be ready to be merged once tests pass...

srvance:
Also, I looked at the impact of build config editing on test time and it wasn't discernable over several runs.

PR #285, opened 2017-03-27
Blueprint honor existing config

### LINKED ISSUE #282: Only override configuration in blueprint when necessary
In the #281 discussion we noted that the blueprint sometimes overrides build configuration settings unconditionally when it shouldn't. For example, `importBootstrapFont` needs to be turned off for BS4, but should be preserved as configured for BS3.

srvance:
When running the blueprint, preserve existing configuration settings if they don't need to be overridden by the target configuration.

Fixes #282

### DIFF
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -196,7 +196,7 @@ module.exports = {
     if (preprocessor !== 'none') {
       settings.importBootstrapCSS = false;
     } else if (config.hasOwnProperty('importBootstrapCSS')) {
-      settings.importBootstrapCSS = config. importBootstrapCSS;
+      settings.importBootstrapCSS = config.importBootstrapCSS;
     } else {
       settings.importBootstrapCSS = true;
     }
--- a/blueprints/ember-bootstrap/index.js
+++ b/blueprints/ember-bootstrap/index.js
@@ -33,6 +33,8 @@ module.exports = {
   normalizeEntityName() {
   },
 
+  existingConfiguration: null,
+
   beforeInstall(option) {
     let bootstrapVersion = parseInt(option.bootstrapVersion, 10) || this.retrieveBootstrapVersion() || 3;
     let preprocessor = option.preprocessor;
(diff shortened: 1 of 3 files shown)


srvance:
The blueprint tests passed, but all the scenarios failed saying they don't have ember-source 2.12.0 installed.

blueprints/ember-bootstrap/index.js
```diff
@@ -178,13 +180,24 @@ module.exports = {
...
+    if (bootstrapVersion === 4) {
+      settings.importBootstrapFont = false;
+    } else if (config.hasOwnProperty('importBootstrapFont')) {
+      settings.importBootstrapFont = config.importBootstrapFont;
+    } else {
+      settings.importBootstrapFont = true;
+    }
 
-    settings.importBootstrapCSS = (preprocessor === 'none');
+    if (config.hasOwnProperty('importBootstrapCSS')) {
```
simonihmig: Do we want to have the same logic here as for `importBootstrapFont`, like first check if we need to override because it is the only possible setting. Here `importBootstrapCSS` is only used if there is no preprocessor, so we could make it like
```js
if (preprocessor !== 'none') {
  settings.importBootstrapCSS = false; 
} else if (config.hasOwnProperty('importBootstrapCSS')) {
  settings.importBootstrapCSS = config. importBootstrapCSS;
} else {
  settings.importBootstrapCSS = true;
}
```

blueprints/ember-bootstrap/index.js
```diff
@@ -178,13 +180,24 @@ module.exports = {
...
+      settings.importBootstrapFont = config.importBootstrapFont;
```
simonihmig: Copy paste? ;) (should be `settings.importBootstrapCSS`)

simonihmig:
> The blueprint tests passed, but all the scenarios failed saying they don't have ember-source 2.12.0 installed.

I noticed that with another build, related to the latest ember-cli upgrade. Cleared all Travis caches and rerun build, seems to work now...

blueprints/ember-bootstrap/index.js
srvance: Doh!

blueprints/ember-bootstrap/index.js
srvance: I debated that. The font setting is a strict requirement because the fonts aren't available in BS4. For the CSS, while it seems wasteful to import the CSS when you have a preprocessor, it's not wrong, so I chose to assume they did it for a reason rather than assume they were wrong.

blueprints/ember-bootstrap/index.js
srvance: Now I need to figure out why the test that I thought covered this didn't.

srvance: It passed because the config editor doesn't change or remove properties that aren't in the config being set.

blueprints/ember-bootstrap/index.js
simonihmig: Hm, but sometime (maybe when Sass support was introduced), this changed in `index.js`, so right now even if `importBootstrapCSS` is true it will be ignored if there has been a preprocessor detected. See https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L49-L58.

So we should align the logic, either allow to import the static CSS even with a preprocessor setup (would require changes in `index.js`), or not (would require the proposed change here), right?

simonihmig:
This commit I pushed to master should fix the failing tests: https://github.com/kaliber5/ember-bootstrap/compare/e6f7d8f62ec1...204ff1198c3b

When you rebase to master, the builds of this branch should hopefully run fine again!

blueprints/ember-bootstrap/index.js
srvance: I agree. I was thinking about this more. We're not checking if they changed from 'none' to a preprocessor, which would be the case where we really should force it. Absent that, forcing it for preprocessors makes sense and if they knew enough to have a contrary setting, they can handle the switch back.

PR #291, opened 2017-04-03
Fix nested addon usage by using ember-cli-import-polyfill

### LINKED ISSUE #286: Error thrown when ember-bootstrap is referenced by another addon.
## Symptoms

After installing an addon to my ember app that depends on `ember-bootstrap`, the following error is thrown on app startup:

```
Path must be a string. Received undefined
TypeError: Path must be a string. Received undefined
  at assertPath (path.js:7:11)
  at Object.join (path.js:1211:7)
  at Class.included (/Users/themikejr/Development/repos/eFlex/webApp/node_modules/ember-bootstrap/index.js:27:30)
  at addons.map.addon (/Users/themikejr/Development/repos/eFlex/webApp/node_modules/ember-cli/lib/models/addon.js:346:34)
  at Array.map (native)
```

## Diagnosis

It seems that when one addon references another addon in this way, the arguments to the `included` function in `index.js` change. ([See this related issue](https://github.com/ember-cli/ember-cli/issues/3718))

[These changes](https://github.com/kaliber5/ember-bootstrap/commit/a2e5b29fc51ebb3b5efef4aae35e36e7f6500b6d) were made to deal with the this problem, but in my case they do not fix this issue. The culprit is the condition: `typeof app.import !== 'function' && app.app`. For some reason in my case, `app.import` is indeed a function. When removing this condition (and leaving the truthy check on `app.app`) it appears to work as expected. (See the code below for more context).

```
module.exports = {
  name: 'ember-bootstrap',

  included: function included(app) {
    // workaround for https://github.com/ember-cli/ember-cli/issues/3718
    if (typeof app.import !== 'function' && app.app) {
      app = app.app;
    }
    this.app = app;

    var options = extend(defaultOptions, app.options['ember-bootstrap']);
    var bootstrapPath = path.join(app.bowerDirectory, 'bootstrap/dist');

```

*I'm wondering if anyone else can recreate this, if I should create a PR with the proposed change, or if there is some environmental condition that I'm missing which makes `typeof app.import !== 'function'` necessary.*

simonihmig:
Hopefully fixes #286

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,11 +35,9 @@ module.exports = {
     };
   },
 
-  included(app) {
-    // workaround for https://github.com/ember-cli/ember-cli/issues/3718
-    if (typeof app.import !== 'function' && app.app) {
-      app = app.app;
-    }
+  included() {
+    this._super.included.apply(this, arguments);
+    let app = this._findHost();
     this.app = app;
 
     let options = extend(extend({}, defaultOptions), app.options['ember-bootstrap']);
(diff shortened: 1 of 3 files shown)


simonihmig:
Closing this, as this polyfill seems not to be intended for addons (see https://github.com/ember-cli/ember-cli-import-polyfill/issues/1#issuecomment-237672090), and there have been issues reported when used like this...

Issue #258, opened 2017-03-08
Add interactive mode to default blueprint


srvance:
In #233 we discussed having an interactive mode using `this.ui.prompt()` that could be invoked during package installation. This captures the intent after we decided to defer it.

kolybasov:
Refreshing it!

Example demo
[![asciicast](https://asciinema.org/a/WQ0BR32IU9JKJvk3zE3lYO8Ix.png)](https://asciinema.org/a/WQ0BR32IU9JKJvk3zE3lYO8Ix)

## Questions

### Fast install
`Do you want to customize installation of ember-bootstrap?`
When it will be a part of default blueprint(even behind flag) we need to allow user to proceed with `bootstrap version` and `preprocessor` only or configure entire set of options otherwise.
> However this question could be skipped if user runs task manually because most likely user want to configure the entire set of options.

### Bootstrap version
`Choose desired Bootstrap version:`
It will be shown every time unless bootstrap version was specified by flag(and version is valid).

### Preprocessor
`Choose desired Bootstrap version:`
It will be show every time unless preprocessor was specified by flag(and is valid).
> For BS4 only `['none', 'sass']` will be available.

### Import Bootstrap Theme
`Do you want to include the optional bootstrap-theme.css?`
It will be show only if user chosen custom install in first question.
And BS3 was chosen in the second step.
And Preprocessor is set to none.

### Import Bootstrap CSS

PR #294, opened 2017-04-07
Add importBootstrapJS

Exelord:
Hi,
I've just added an option `importBootstrapJS` to allow users easily migrate from pure bootstrap to ember-bootstrap.

In our case when we are rewriting our components, it is a key feature to use bootstrap js and ember-bootstrap simultaneously till we upgrade them all.

Unfortunately, we can not use bootstrap in bower because of:
```js
validateDependencies() {
    let bowerDependencies = this.app.project.bowerDependencies();
    if ('bootstrap' in bowerDependencies
      || 'bootstrap-sass' in bowerDependencies) {
      throw new SilentError('The dependencies for ember-bootstrap are outdated. Please run `ember generate ember-bootstrap` to install the missing dependencies!');
    }
  },
```

so I made this PR to allow on include.

___
I also fixed some styling issues.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,16 +7,14 @@ const extend = util._extend;
 const mergeTrees = require('broccoli-merge-trees');
 const Funnel = require('broccoli-funnel');
 const stew = require('broccoli-stew');
-const mv = stew.mv;
-// const log = stew.log;
-const rm = stew.rm;
-const chalk = require('chalk');
+const { rm, mv } = stew;
 const SilentError = require('silent-error'); // From ember-cli
 
 const defaultOptions = {
   importBootstrapTheme: false,
   importBootstrapCSS: true,
   importBootstrapFont: true,
+  importBootstrapJS: false,
   insertEmberWormholeElementToDom: true,
   bootstrapVersion: 3
 };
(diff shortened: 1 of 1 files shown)


simonihmig:
Hi @Exelord. Thanks for your effort. Unfortunately I am not so sure if I really want to accept this. The purpose or this addon is to make `bootstrap.js` useless. I do understand your case, so it should be possible to import it from the app. But I don't think I want to actively support that with an import option, as it would give the impression that this is something we support, which I definitely don't want to!

I would prefer a way to allow this, without really supporting it! 😉  Maybe just change the `validateDependencies` implementation to just issue a warning rather than throwing an error? @srvance you have an opinion on this?

Besides that, there are some issues with your code changes:
* you cannot use ES6 destrucuring in node 4, that's why tests are failing
* you import `bootstrap.js` *and* `bootstrap.min.js`, which does not make sense to me. I would always import the unminified version and let ember-cli do the rest...

Exelord:
I agree with `warning` instead of `error`.
But I still need to have a bower dependency for bootstarp which was already added in npm :/

Maybe we can just rename the flag to sth with better meaning that this should only be used in case of migration.

srvance:
Can you avoid the error by moving your bootstrap dependencies to npm instead of bower? I'm fine making it a warning, instead of an error, though.

As for removing the dependency, you can add it back in or reset the change to the file if we change the error to a warning. We're not removing bootstrap completely, just moving the way it's imported.

The blueprint is tailored to the most common cases, figuring if you know enough to need more complicated alternatives, you probably know enough to reconfigure it yourself.

@simonihmig I agree that we shouldn't explicitly support importing the JS files.

simonihmig:
Ok, so I would then go this route:
* we replace the error with a warning
* as you only run the blueprint when installing the addon, or when explicitly calling `ember g ember-bootstrap`, it should still work fine if you re-add bootstrap to `bower.json`. I agree with @srvance we should optimize for the most common use case, which is ember-bootstrap is used without any other bootstrap related addons/imports, but still allow special cases to work. So I would keep the uninstall function in the blueprint, as this makes very much sense when upgrading from an older version. You can still import `bootstrap.js` installed into `bower_components` in your `ember-cli-build.js` using `app.import()`
* maybe when we add #258, we could make the user have to confirm the uninstall

@Exelord do you want to update your PR accordingly? Would require the line where the error is thrown to be replaced with something like https://github.com/kaliber5/ember-bootstrap/blob/master/blueprints/ember-bootstrap/index.js#L205. Otherwise, as this is a pretty trivial change, I can do that as well...

Exelord:
Ok, I will implement this that way, but I still think that the option to import boostrap JS should be a built-in feature. 
It's your package so I respect that decision

Exelord:
We can close this PR in favor of #297

PR #301, opened 2017-04-13
prepare for upcoming FastBoot 1.0 breaking changes

simonihmig:
See https://github.com/ember-fastboot/ember-cli-fastboot/issues/360

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,9 +68,7 @@ module.exports = {
       }
     }
 
-    if (!process.env.EMBER_CLI_FASTBOOT) {
-      app.import('vendor/transition.js');
-    }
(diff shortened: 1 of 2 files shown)


kratiahuja:
@simonihmig this looks good to me! Can you test this in an existing fastboot app to make sure it works? I can test it on a dummy app with my WIP branch.

simonihmig:
@kratiahuja I have FastBoot tests in that addon, and they are all green in Travis. And I manually tested this when setting its own devDependency of ember-cli-fastboot to your fork with running `ember s`, which was working just fine with the dummy app (seeing all content rendered in the page source). So seems to be ok!

PR #302, opened 2017-04-13
right aligned tabs

devinholloway:
Needed the ability to align the tabs to the right side using the pull-right class.

### DIFF
--- a/addon/components/base/bs-nav/item.js
+++ b/addon/components/base/bs-nav/item.js
@@ -17,7 +17,7 @@ const { computed, LinkComponent } = Ember;
 export default
 Ember.Component.extend(ComponentParent, {
   layout,
-  classNameBindings: ['disabled', 'active'],
+  classNameBindings: ['disabled', 'active', 'pullRight'],
   tagName: 'li',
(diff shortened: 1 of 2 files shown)


srvance:
@devinholloway Thank you for the contribution!

1. We have tried to build support for features that can't just be achieved by adding classes directly to the component instance. What do you see as the advantage of having this built in?
1. `pull-*` is part of a family consisting of both `pull-left` and `pull-right`. `pull-right` is probably more commonly used due to the default left alignment, but we should support both if we support this.
1. Bootstrap 4 removes `pull-*` in favor of a larger set of `float-*` classes to both remove redundancy and provide a wider range of responsive options (see [here](https://github.com/twbs/bootstrap/blob/8edfe0ff561eab8b2610db6f161afa913ed8cb7b/docs/migration.md#utilities)). We should support this for BS4, as well.

simonihmig:
Closing this due to lack of feedback. I also agree that these classes can be easily added with `class="pull-right"` without having to explicitly "support" them. Thanks anyway for your contribution, @devinholloway!

PR #304, opened 2017-04-19
Add `helpText` to FormElement

### LINKED ISSUE #169: Proposal add  helpBlock on FormElement
````
<label class="sr-only" for="inputHelpBlock">Input with help text</label>
<input type="text" id="inputHelpBlock" class="form-control" aria-describedby="helpBlock">
...
<span id="helpBlock" class="help-block">A block of help text that breaks onto a new line and may extend beyond one line.</span>
````

### PR DESCRIPTION
**Pull request by simonihmig**
Closes #169

### CODE REVIEW

File: addon/components/base/bs-form/element/control.js [hunk 1]
```diff
@@ -16,6 +16,13 @@ export default Ember.Component.extend({
   value: null,
 
   /**
+   * @property ariaDescribedby
+   * @type {string}
+   * @public
+   */
+  ariaDescribedby: null,
```
srvance: Uppercase "By"

addon/components/base/bs-form/element/control/input.js
```diff
@@ -32,7 +32,8 @@ export default Control.extend({
     'step',
     'form',
     'spellcheck',
-    'title'
+    'title',
+    'ariaDescribedby:aria-describedBy'
```
srvance: Why not put this in Control?

addon/components/bs4/bs-form/element/help-text.js
```diff
@@ -0,0 +1,5 @@
+import FormElementHelpText from 'ember-bootstrap/components/base/bs-form/element/help-text';
+
+export default FormElementHelpText.extend({
+  classNames: ['form-text']
```
srvance: I thought this was elsewhere in the validation feedback. I found it [here](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-form/element.js#L98) in the docs, but I don't know what happened to the code reference to it. It seems this class is used for any feedback presented in the form.

addon/components/base/bs-form/element/control.js
simonihmig: Yeah, I had it like this before, but then realized that the final attribute is actually `aria-describedby`. If the camel case version would be `ariaDescribedBy`, then the dasherized attribute name should be (at least to be consistent with Ember's inflection rules) `aria-described-by`, right?

addon/components/base/bs-form/element/control/input.js
simonihmig: I also thought about this, but this does not apply to only that attribute. In fact you could put most attributes to the parent `control` that are shared among all the current controls. But afaik you cannot remove things from `attributeBindings` in a child class that a parent class has added. So I don't know if putting all into the parent is not assuming too much, like what happens when we add a special control that should not have some of those attributes applied? So I thought even if there is some redundancy involved here, we keep the actual control responsible for defining its allowed attributes. But I'm open to other arguments ;)

addon/components/bs4/bs-form/element/help-text.js
simonihmig: It seems it was part of the `errors` component in BS4, but removed later. See this: https://github.com/kaliber5/ember-bootstrap/commits/master/addon/components/bs4/bs-form/element/errors.js

addon/components/base/bs-form/element/control/input.js
srvance: I would put them in the parent until we know we have a child that doesn't need it. We can break it out if and when we add a special control. I think you can remove things from `attributeBindings` by creating a read-only `null` CP of that name.

addon/components/bs4/bs-form/element/help-text.js
srvance: Thanks. Looks like I missed the documentation reference. Would you mind cleaning up the docs in that area?

addon/components/base/bs-form/element/control.js
srvance: But you're overriding the emitted attribute name anyway, so I'd stick with convention in the attribute naming.

addon/components/base/bs-form/element/control/input.js
simonihmig: That's assuming we have control of when and what kind of controls are added. But I actually worked the past week on a future `ember-bootstrap-power-select` addon, that just adds some more convenient support for the `power-select` component to our forms. It works by adding a `bs-form/element/control/power-select` component that glues them together. This works because of #299, where it automatically picks up this new component when you use `{{form.element controlType="power-select"}}`. Have not pushed this yet, but I can do so later.

So this is basically an external component, that `ember-bootstrap`  is not really aware of. And when we make this way of extending form control "public", we cannot assume what controls we have...

addon/components/base/bs-form/element/control.js
simonihmig: Ok, makes sense

addon/components/base/bs-form/element/control/input.js
srvance: Okay. Should we refactor this to a mixin so it's in one place and can optionally be pulled into controls?

simonihmig: Did that...

addon/components/base/bs-form/element/control.js
simonihmig: done

addon/components/bs4/bs-form/element/help-text.js
simonihmig: done

Issue #283, opened 2017-03-23
bs-accordion with sub-component as header


danidr:
Hey there,
First of all thanks for your add-on, it's been a life saver for me in a few projects.
So, right now I am building a quite complex interface, and I am a little limited by the fact that you can only pass a `title` parameter to `{{acc.item}}`; it would be way more flexible if we had a sub-component to handle the content. The structure could be:
```htmlbars
{{#acc.item as |aitem|}}
  {{#aitem.title}}Whatever{{/aitem.title}}
  . . .
{{/acc.item}}
```

Of course, I could try and implement a pull request; I'll see if I have time.

simonihmig:
I understand the need for this, will have a look.

Thanks for the nice words, btw! :)

bluscreen:
i agree this would be a nice feature. here is some code on how i used it in another project:

template of acc item:
```
{{yield headerPart}}
<div role="tab" class="panel-heading {{if collapsed "collapsed"}}">
    <h4 class="panel-title">
        {{#if isMeldung}}
            <span {{action 'meldungClicked' meldungen.firstObject}}
                    class="{{unless showWhite 'red'}} pointer margin-right--10 glyphicon glyphicon-hand-right"></span>
        {{/if}}
        <a href="#" class="dummy" {{action "toggleActive"}}>
            {{yield titlePart}}
        </a>
        {{yield actionPart}}
    </h4>
</div>
{{yield additionalPart}}
{{#bs-collapse collapsed=collapsed class="panel-collapse"}}
    <div class="panel-body flat">
        {{yield bodyPart}}
    </div>
{{/bs-collapse}}
{{yield footerPart}}
```

component of acc item (excerpt)
```
    headerPart: {isHeader: true},
    titlePart: {isTitle: true},
    actionPart: {isAction: true},
    bodyPart: {isBody: true},
    additionalPart: {isAdditional: true},
    footerPart: {isFooter: true},
```

usage:

```
        {{#accordion-item value="xy" as |section| }}
            {{#if section.isTitle}}
                <h7> bla
                    <small class="pull-right hidden-xs">blubb</small>
                </h7>
            {{else if section.isBody}}
                {{partial "xy"}}
            {{/if}}
        {{/accordion-item}}
```

cheers

Daniel

simonihmig:
I would go for an API like this:

```hbs
{{#bs-accordion as |acc|}}
  {{#acc.item value="1" title="First item"}}
[...]

PR #307, opened 2017-04-22
Add the contextual component to accordion's title

dajk:
Being able to create specific template to accordion's title (issue #283)


srvance:
@dajk The tests seem to be failing because there is no `item.body` contextual.

simonihmig:
@dajk the idea is interesting, that you `{{yield}}` twice, but each with a different component (title/body). Never used this pattern, although it seems to work with the latest ember versions, but apparently not with the older ones.

I thought to do it like this: `yield` once, but yield a hash of both title and body components. Then all the markup (like `.panel-heading`, the `bs-collapse` component etc.) goes into those components, so `bs-accordion/item.hbs` is just an empty template with this `{{yield}}`. Similar to things like the modal, that yields header, body, footer...

dajk:
Ok @simonihmig, thanks for clear and detail instructions. Just let me know if you have any additional instruction.

addon/templates/components/bs4/bs-accordion/title.hbs
```diff
@@ -0,0 +1,8 @@
+{{! template-lint-disable nested-interactive }}
+<div role="tab" class="card-header {{if collapsed "collapsed" "expanded"}}">
```
simonihmig: As Ember components (for now, newer glimmer components are better in this regard IMHO) always have their own root DOM element (unless you set `tagName: ''`), this will result in this `<div  class="card-header">` being wrapped in another (useless) `<div>` from the component itself. Better remove this `<div>` form the template and reuse the components element, by using the special component properties like `classNames`, `classNameBindings`, `role` to customize the root element.

addon/components/base/bs-accordion/item/body.js
```diff
@@ -0,0 +1,24 @@
...
+ See [Components.Accordion](Components.Accordion.html) for examples.
+
+ @class AccordionBody
+ @namespace Components
+ @extends Ember.Component
+ @public
+ */
+export default Ember.Component.extend({
+  layout,
+
```
simonihmig: for the same reason as the other comment, you can set `tagName: ''` here to make it a tagless component, removing the unnecessary wrapping `<div>`

addon/templates/components/bs4/bs-accordion/title.hbs
simonihmig: Btw, you cannot use the `tagName: ''` approach here, as that would make the `click` handler on the component not work anymore (cannot have an event listener on a tagless component obviously).

tests/integration/components/bs-accordion-test.js
```diff
@@ -328,3 +328,103 @@ testBS4('yields change action to add custom behaviour', function(assert) {
...
+    {{#acc.item value=1 as |aitem|}}
+      {{#aitem.title}}TITLE1{{/aitem.title}}
+      {{#aitem.body}}CONTENT1{{/aitem.body}}
+    {{/acc.item}}
+    {{#acc.item value=2}}CONTENT2{{/acc.item}}
+  {{/bs-accordion}}`);
+  let item = this.$('.panel:eq(0)');
+  let done = assert.async();
+
+  item.find('.panel-heading').click();
```
simonihmig: We could unify the separate tests for BS3/BS4 here, by using a helper like this to get theBS version depending class name: https://github.com/kaliber5/ember-bootstrap/blob/master/tests/helpers/bootstrap-test.js#L62-L64. It seems the logic is identical, the difference are just class names. So by not hardcoding class names instead using some helper, we can have a single test implementations for both B3 and BS4.

simonihmig:
@dajk Looks pretty good, but somehow the BS4 tests are failing, not sure why. You can switch to BS4 locally with `BOOTSTRAPVERSION=4 ember s`.

simonihmig:
Good job, thanks! :)

PR #308, opened 2017-04-22
Use Chrome in CI

simonihmig:


srvance:
Check out [this for Firefox](https://docs.travis-ci.com/user/firefox/). It doesn't look like Chrome supported.

simonihmig:
@srvance thanks. Yes, FF was preinstalled, but pretty old version it seems. Nevertheless it seems it is running now fine with Chrome on Travis! Had to use the `trusty` image...

PR #310, opened 2017-04-27
fix(Modal): Sending wrong event by clicking on close button

dajk:


simonihmig:
Actually the `onClick` evebnt was sent, otherwise it would not have worked (which it does), but it would break if the component was used without setting `onClick` to a function. So this is still needed. Good catch, thanks!

dajk:
Yes, right. Just wrongly explained message. 
Anytime 😉

PR #323, opened 2017-05-15
Soften the build-time errors to warnings for non-standard configurations

### LINKED ISSUE #322: Having a ember-cli-sass installed without boostrap-sass breaks build.
There is no way to not use bootstrapp-sass (or bootstrap-less (via `importBootstrapTheme` and `importBootstrapCSS`) if you have a supported preprocessor installed.

Therefore, since I have ember-cli-sass installed (for other things), but am including the boostrap css elsewhere, I get an error when trying to build ([index.js:91](https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L91)):

> Npm package "bootstrap-sass" is missing, but required for SASS support. Please run `ember generate ember-bootstrap` to install the missing dependencies!

I tried running `ember generate ember-bootstrap --preprocessor=none` but all that accomplished was it uninstalling my ember-cli-sass, which wasn't precisely the solution I was looking for.

(For now, I'm working around this by installing ember-boostrap but removing the import from app.scss)
---

> Output from `ember bootstrap:info` 
> ```
> Npm packages:
> ember-bootstrap: ^1.0.0-alpha.12 -> 1.0.0-alpha.12
> ember-cli: 2.12.1 -> 2.12.1
> bootstrap: ^3.3.7 -> 3.3.7
> bootstrap-sass: n/a
> ember-cli-sass: ^5.6.0 -> 5.6.0
> ember-cli-less: n/a
> 
> Bower packages:
> bootstrap: n/a
> bootstrap-sass: n/a
> 
> ember-bootstrap configuration:
> importBootstrapTheme: false
> importBootstrapCSS: false
> bootstrapVersion: 3
> importBootstrapFont: true
> ```

srvance:
Fixes #322.

Should we soften the less+BS4 error? While BS4 doesn't support less, that doesn't mean someone couldn't use less in their project with BS4.

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,7 +75,7 @@ module.exports = {
     let bowerDependencies = this.app.project.bowerDependencies();
 
     if ('bootstrap' in bowerDependencies || 'bootstrap-sass' in bowerDependencies) {
-      this.ui.writeLine(chalk.yellow('The dependencies for ember-bootstrap can be outdated. Please run `ember generate ember-bootstrap` to install appropriate dependencies!'));
+      this.ui.writeLine(chalk.yellow('The dependencies for ember-bootstrap may be outdated. Please run `ember generate ember-bootstrap` to install appropriate dependencies!'));
     }
   },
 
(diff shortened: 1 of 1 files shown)


simonihmig:
Thanks @srvance! Will merge once tests pass...

srvance:
@simonihmig What do you think about the less+BS4 case above? Wait for someone to ask for it or soften it also?

simonihmig:
@srvance I think to soften that also makes sense. At least the message is now misleading: "Falling back..." is not true when we throw...

PR #324, opened 2017-05-15
Updated ModalSimple FormElement documentation examples for v1.0

HodofHod:
bs-form-element is no longer an independent component.


simonihmig:
Good catch, thanks!

HodofHod:
Wow, that was fast! Kudos for the near-immediate response!

PR #326, opened 2017-05-16
No jquery

### LINKED ISSUE #300: Drop jQuery
* Remove the usage of jQuery in component code, to allow building jQuery free apps.
* would require refactoring tests to use `ember-native-dom-helpers`

See http://miguelcamba.com/blog/2017/04/11/the-future-of-embers-testing-and-the-beheading-of-jquery/

simonihmig:
Refactored all components to use native DOM instead of jQuery. Allows to build jQuery free apps. Closes #300 

Some tests will still fail, until https://github.com/rwjblue/ember-native-dom-event-dispatcher/pull/10 is merged and integrated...

cc @srvance don't know if you really want to look through all this, but at least be aware of it! ;)


addon/components/base/bs-contextual-help.js
```diff
@@ -213,20 +211,20 @@ export default Component.extend(TransitionSupport, {
...
+   * The DOM element of the viewport element.
    *
    * @property viewportElement
    * @type object
    * @readonly
    * @private
    */
   viewportElement: computed('viewportSelector', function() {
-    return $(this.get('viewportSelector'));
+    return document.getElementById(this.get('viewportSelector'));
```
srvance: This really restricts the selector that can be passed in.

addon/components/base/bs-contextual-help.js
```diff
@@ -155,7 +153,7 @@ export default Component.extend(TransitionSupport, {
   transitionDuration: 150,
 
   /**
-   * Keeps the tooltip/popover within the bounds of this element when `autoPlacement` is true. Can be any valid jQuery selector.
+   * Keeps the tooltip/popover within the bounds of this element when `autoPlacement` is true. Can be any valid CSS selector.
```
simonihmig: you are right. I hope in most cases it should not matter, but when people use some non-standard jQuery selectors like `: eq` and a bunch of others, this would be a breaking change indeed...

addon/components/base/bs-contextual-help.js
simonihmig: Oh, I think that was by mistake. Should really use `querySelector` as well. Good catch!

- [ ] Fix https://github.com/kaliber5/ember-bootstrap/pull/326#discussion-diff-116896205R222

PR #333, opened 2017-05-30
Enhance addon support

srvance:
Update the version of `ember-cli-build-config-editor` which supports config in `EmberAddon`s.


simonihmig:
Thanks!

Issue #332, opened 2017-05-29
Don't include glyphicon css


Techn1x:
So I've had this option set for a while now, which stops the glyphicon font from being included
`'importBootstrapFont': false`

Rather than glyphicons, I'm using fontawesome. But I've recently started using a bootstrap datepicker
https://github.com/dangrossman/bootstrap-daterangepicker

Or more specifically, an ember implementation of it
https://github.com/josemarluedke/ember-cli-daterangepicker

It seems the way the datepicker works with icons is it supports both fontawesome & glyphicon, but it does this by adding both the glyphicon & fontawesome css classes, for example:
```handlebars
<i class="fa fa-calendar glyphicon glyphicon-calendar"></i>
```

When I go to use the datepicker, it attempts to load & use the glyphicons, which it can't find font files for. It seems as though the glyphicon css files are still being included, despite the setting above.

What is the best way to not include the glyphicon css? Or if I'm completely wrong, what's the best way to fix the issue I'm having?

Thanks

simonihmig:
If you include the static bootstrap.css, there is no way to remove the glyphicons portion, as it's just one big file that is imported. If you use a preprocessor (Less or Sass), then you can pick and choose which .less/.scss you want to import (so skip the glyphicons). See http://www.ember-bootstrap.com/#/getting-started/assets

Techn1x:
Thanks Simon. I am using a sass preprocessor, and my app.scss previously had
`@import "ember-bootstrap/bootstrap"`

I've tried replacing that with statements like (for example, the variables stylesheet)
`@import "ember-bootstrap/variables"`
and
`@import "ember-bootstrap/variables.scss"`

But when building my app it doesn't seem to be able to find the files... to get it to work I actually ended up having to do the sass import statements of the form
`@import "ember-bootstrap/bootstrap/variables"`

Should this be made clearer in the doc you linked? Or is it just something specific about my setup?

simonihmig:
It is specific to Sass, the Less files of v3 bootstrap don't have this `bootstrap` subfolder. And the example was based on Less. But adding an example for Sass would make sense then. Want to create a PR? 😛  (*The docs are part of the tests/dummy app)

Techn1x:
Sure! Pull request submitted

PR #334, opened 2017-05-30
Updated importing assets guide to include sass examples

Techn1x:
As mentioned here;
https://github.com/kaliber5/ember-bootstrap/issues/332#issuecomment-304817706


simonihmig:
Thanks!

Issue #336, opened 2017-06-01
bs4 checkbox and radio syntax missing


gijsbotje:
the bs4 checkbox uses the markup from common, which is the bs3 way.

bs3: http://getbootstrap.com/css/#checkboxes-and-radios
bs4: https://v4-alpha.getbootstrap.com/components/forms/#checkboxes-and-radios

this results in non styled checkboxes that don't align properly
i don't have enough knowledge of ember to build them myself, so if someone would be so kind to create these that would be awesome

simonihmig:
That seems to be wrong indeed. Thanks for reporting!

gijsbotje:
@simonihmig if i have more elements that are not complient to the bs4 docs, should i comment them here or make new issues?

simonihmig:
@gijsbotje Feel free to create additional issues, at least for all elements that are not really related! Hope there aren't so many! ;)

gijsbotje:
@simonihmig prepare yourself.... :)

PR #341, opened 2017-06-02
Change the form element components conform BS4 standards

Rapid0o:
Change the form component and form element components (especially the checkbox component) to be conform the standards for BS4, while keeping the BS3 components conform BS3 standards (related to issue [#336](https://github.com/kaliber5/ember-bootstrap/issues/336)).


addon/templates/components/bs4/bs-form/element/errors.hbs
```diff
@@ -0,0 +1,3 @@
+{{#if show}}
+  <div class={{feedbackClass}}>{{messages.firstObject}}</div>
+{{/if}}
```
simonihmig: It seems the error-templates for `bs3` and `bs4` are still the same, so it could still be a single template in ? common`, right?

addon/components/base/bs-form/element.js
```diff
@@ -330,7 +330,7 @@ const nonDefaultLayouts = A([
  */
 export default FormGroup.extend({
   layout,
-  classNameBindings: ['disabled:is-disabled', 'required:is-required', 'isValidating'],
+  classNameBindings: ['disabled:disabled', 'required:is-required', 'isValidating'],
```
simonihmig: These classes were never part of the official ones defined by Bootstrap, so is there a good reason to rename `is-disabled` to `disabled`? That would be a breaking change, which I would like to prevent if possible...

addon/components/bs4/bs-form/element/label.js
```diff
@@ -1,9 +1,65 @@
...
+} = Ember;
+
 export default FormElementLabel.extend({
   tagName: 'label',
 
-  classNames: ['form-control-label'],
-  classNameBindings: ['invisibleLabel:sr-only', 'labelClass'],
-  attributeBindings: ['formElementId:for']
+  classNames: [],
+  classNameBindings: ['invisibleLabel:sr-only', 'isHorizontal:col-form-label', 'isCheckbox:form-check-label', 'labelClass'],
```
simonihmig: The change to `col-form-label` seems to overlap https://github.com/kaliber5/ember-bootstrap/pull/260, and is subject to https://github.com/twbs/bootstrap/issues/22173 getting clarified. But this might be ok for now?

addon/components/bs4/bs-form/element/label.js
```diff
@@ -1,9 +1,65 @@
...
+   * * 'textarea'
+   * * any other type will use an input tag with the `controlType` value as the type attribute (for e.g. HTML5 input
+   * types like 'email'), and the same layout as the 'text' type
+   *
+   * @property controlType
+   * @type string
+   * @default 'text'
+   * @public
+   */
+  controlType: 'text',
```
simonihmig: I would prefer these two additional properties (`controlType` and `formLayout` to live in `base/bs-form/element/label`, so we can keep using only `common/bs-form/element.hbs` for bs3 and bs4. This would hide the fact that the bs4 label renders differently based on these properties as an implementation detail of that component, and we don't have to care otherwise.

addon/components/bs4/bs-form/element/label.js
```diff
@@ -1,9 +1,65 @@
...
+
+  /**
+   * Indicates whether the type of the control widget equals `checkbox`
+   *
+   * @property isCheckbox
+   * @type boolean
+   * @public
```
simonihmig: Should be private, we don't set these from outside...

addon/components/bs4/bs-form/element/label.js
```diff
@@ -1,9 +1,65 @@
...
+   */
+  isCheckbox: computed.equal('controlType', 'checkbox').readOnly(),
+
+  /**
+   * Indicates whether the form type equals `horizontal`
+   *
+   * @property isHorizontal
+   * @type boolean
+   * @public
```
simonihmig: Same

tests/integration/components/bs-form/element-test.js
```diff
@@ -589,4 +589,4 @@ test('shows help text if available', function(assert) {
 
   assert.ok(find('input').getAttribute('aria-describedby'), 'control has aria-describedby attribute');
   assert.equal(find('input').getAttribute('aria-describedby'), find(helpTextClass).getAttribute('id'), 'aria-describedby matches id');
-});
\ No newline at end of file
+});
```
simonihmig: There should be a couple of new tests to cover the changes here, specifically the newly introduced classes (`col-form-label`, `form-check`, `form-check-label`, `form-check-input`). @Rapid0o can you add those? See the other tests for examples, sometimes there are separate tests for bs3/bs4, sometimes they are just diffentiated by using a test helper function to get the expected class name...

addon/templates/components/bs4/bs-form/element/errors.hbs
Rapid0o: The only difference is the element tag. BS4 uses div tags to render validation messages, while BS3 uses span tags. As far as i have seen, it doesn't seem to affect the styling in any way. I changed it to keep it closer to the BS4 standard. But i guess for keeping maintainability we could keep the errors template in common and ignore the tag difference. What is your opinion on this @simonihmig ?

addon/components/base/bs-form/element.js
Rapid0o: I changed it to `disabled` instead of `is-disabled`, because BS4 and BS3 both add styling (like a `cursor` etc.) to the `label` when using the `disabled` class (see for example https://v4-alpha.getbootstrap.com/components/forms/#checkboxes-and-radios).

But you're right it isn't worth a breaking change. I guess it would be better to keep it `is-disabled` and let the users add the styling themselves?

addon/components/bs4/bs-form/element/label.js
Rapid0o: You're right. I will change it :+1:

addon/components/bs4/bs-form/element/label.js
Rapid0o: :+1:

addon/components/bs4/bs-form/element/label.js
Rapid0o: :+1:

addon/components/bs4/bs-form/element/label.js
Rapid0o: My bad, you're right [#260](https://github.com/kaliber5/ember-bootstrap/pull/260) does overlap. The only difference is the use of the `isHorizontal` check. 

I have seen the same issue that is mentioned, when i was looking at the BS4 documentation. I think the use of `form-control-label` is a (copy from BS3?) mistake in the BS4 documentation. At least 90% of the BS4 form documentation uses `form-control-label`, i think it is ok to use it.

tests/integration/components/bs-form/element-test.js
Rapid0o: I will try to add some tests :+1:

addon/templates/components/bs4/bs-form/element/errors.hbs
srvance: It's more a preference than a rule. Either works. I didn't feel it was worth differentiating.

addon/components/bs4/bs-form/element/label.js
srvance: I'd wait on this until BS4 releases their beta.

addon/templates/components/bs4/bs-form/element/errors.hbs
simonihmig: Agree, let's keep a single template!

addon/components/base/bs-form/element.js
simonihmig: Actually, if BS4 supports the `disabled` class as it seems (I have seen `disabled` for BS3 only for `input` and `fieldset`), we should support it! The `is-desiabled` class was never documented, so should be ok to change that, at least while still in beta. So let's keep `disabled`!

addon/templates/components/bs4/bs-form/element/errors.hbs
gijsbotje: well personally i would prefer a div, imagine you have 2 error rules.
span will place the elements in a single line until it has to break
div are displayed as a block so they will display on 2 lines

i'm guessing that's why they changed it

simonihmig: @gijsbotje I agree

Rapid0o:
Thnx for reviewing! I added some changes based on the reviews :smiley:

simonihmig:
@Rapid0o Awesome! I think the only missing piece is reverting the separate error-templates to a single (`common`) again, using `<div>`s!
(see https://github.com/kaliber5/ember-bootstrap/pull/341/files/c15e285b3ba9be80b0aec18dd1c4534843b9b462#diff-21b014979425000a1b48f85632f284a5)

Rapid0o:
@simonihmig I agree, i have reverted it back to common :+1:

simonihmig:
Merged. Thanks a lot @Rapid0o!

PR #342, opened 2017-06-03
Add customWarning property to show warnings independently of validation

elgordino:
Adds a customWarning property with similar behavior to customError.

The customWarning will be displayed immediately. If showValidationOn has been triggered and the form has errors then the errors will be displayed instead of the warning.

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -507,23 +507,23 @@ export default FormGroup.extend({
   hasCustomWarning: computed.notEmpty('customWarning'),
 
   /**
-   * The array of validation messages (either errors or warnings) from the `model`'s validation.
+   * The array of validation messages (either errors or warnings) from either custom error/warnings or , if we are showing model validation messages, the model's validation
    *
    * @property validationMessages
    * @type array
    * @private
    */
-  validationMessages: computed('hasCustomError', 'customError', 'hasErrors', 'hasCustomWarning', 'customWarning', 'hasWarnings', 'errors.[]', 'warnings.[]', function() {
+  validationMessages: computed('hasCustomError', 'customError', 'hasErrors', 'hasCustomWarning', 'customWarning', 'hasWarnings', 'errors.[]', 'warnings.[]', 'showModelValidation', function() {
     if (this.get('hasCustomError')) {
       return A([this.get('customError')]);
     }
-    if (this.get('hasErrors')) {
+    if (this.get('hasErrors') && this.get('showModelValidation')) {
       return A(this.get('errors'));
     }
     if (this.get('hasCustomWarning')) {
       return A([this.get('customWarning')]);
     }
-    if (this.get('hasWarnings')) {
+    if (this.get('hasWarnings') && this.get('showModelValidation')) {
       return A(this.get('warnings'));
     }
     return null;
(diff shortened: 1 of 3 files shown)


addon/components/base/bs-form/element.js
```diff
@@ -617,11 +639,16 @@ export default FormGroup.extend({
    * @type string
    * @private
    */
-  validation: computed('hasCustomError', 'hasErrors', 'hasWarnings', 'hasValidator', 'showValidation', 'isValidating', 'disabled', function() {
+  validation: computed('hasCustomError', 'hasErrors', 'hasCustomWarning', 'hasWarnings', 'hasValidator', 'showValidation', 'isValidating', 'disabled', function() {
     if (!this.get('showValidation') || !this.get('hasValidator') || this.get('isValidating') || this.get('disabled')) {
       return null;
+    } else if (this.get('showOwnValidation') || this.get('showAllValidations')) {
```
simonihmig: Why do we need this added if-statement, compared to the previous implementation? 
And `showOwnValidation` and `showAllValidations` are not declared as dependent keys of the computed property, if we really need this...

addon/components/base/bs-form/element.js
```diff
@@ -617,11 +639,16 @@ export default FormGroup.extend({
...
+      // The display of validation messages has been enabled
+      return this.get('hasErrors') || this.get('hasCustomError') ? 'error' : (this.get('hasWarnings') || this.get('hasCustomWarning') ? 'warning' : 'success');
+    } else {
+      // If there are custom errors or warnings these should always be shown
+      return this.get('hasCustomError') ? 'error' : (this.get('hasWarnings') || this.get('hasCustomWarning') ? 'warning' : null);
```
simonihmig: Related to the previous comment, the `null` return value here can never be reached, I think? If this statement is reached, either `hasCustomError` or `hasCustomWarning` has to be true.

addon/components/base/bs-form/element.js
elgordino: The intent is to distinguish between the two possible causes for validation to be displayed

* User has triggered a `showValidationOn` state, or the component consumer has set `showOwnValidation`.
or
* Model has `customError` or `customWarning`

If `showOwnValidation` is true then I think the order of displaying messages is:

* `customError`
* The model's validation error
* `customWarning`
* The model's validation warning
* The success state

However if `showOwnValidation` is false then I think the order of validation messages is
* `customError`
* `customWarning`
* no message displayed

This isn't very explicit behaviour in the code though... perhaps if the property were called `showModelValidations` or something similar the intent would be clearer. This could be a CP observing `showOwnValidation` and `showAllValidations`, ie the same as `showValidations` but with the two `has*` properties removed.

What do you think?

addon/components/base/bs-form/element.js
elgordino: Yep you're right. Thanks. Will tweak.

addon/components/base/bs-form/element.js
simonihmig: @elgordino thanks for the clarification! I agree on the general logic, and like adding `showModelValidations` to make this more readable!

elgordino:
Requested changes made. I made one further change to fix an issue where the component had `customWarning` and model errors, but not `showOwnValidation` then a warning level style would be applied but the error from the model validation.

Now it shows the customWarning as expected. I updated the test plan to catch this too.

simonihmig:
@elgordino Thanks a lot for working on this! 😀

elgordino:
@simonihmig Excellent! Thanks for merging it 👍

Issue #317, opened 2017-05-04
RFE: Auto collapse Navbar during navigation


wireframe:
Recollapsing the nav-bar when a user clicks on a link would be a great feature enhancement.  Hooking into the ember router `didTransition` event should be possible to accomplish this.

simonihmig:
Makes sense! 

Will have to figure out how the API should look like. Not sure if making this depend on the router is the best way. What if a link just triggers an action without a route change? Might be better to trigger this on any click on a `nav.item`?

Techn1x:
I found that using didTransition() caused stutters in the collapse animation when I clicked on a slower loading page (specifically, with a map view)

Instead, I used willTransition() and the collapse animation is much smoother (because it is triggered before loading the slow page)

I agree that it would be better if it wasn't dependent on the router in the first place, but here is my basic implementation that seems to work well for me;

template
```
{{#bs-navbar collapsed=collapseValue as |navbar|}}
    ....
{{/bs-navbar}}
```

controller
`collapseValue: true, // Start navbar as collapsed`

route
```
actions: {
    willTransition() {
        // Collapse navbar on every transition
        this.controllerFor('application').set('collapseValue', true);
        return true; // Bubble
    }
}
```

Techn1x:
Would something like this be possible? (perhaps there might be nav items that you don't want to collapse the navbar on..)
{{nav.item collapseNavbar=true}}

danidr:
I have proposed an interface in the referenced issue above, #321, where the collapseNavbar property @Techn1x mentioned is handled in the `{{nav.link-to}}` instead.
Also, the workaround above is fine when using the navbar directly in a "standard" template (not a component), but when putting it inside a component, that would need a lot of variables passing around. Yep, I do believe this would be a very nice enhancement 😊

wireframe:
@Techn1x and @danidr I *really* like the proposal to have the setting on the link vs a global mixin!

PR #348, opened 2017-06-06
Expose Navbar collapse property and actions

simonihmig:
Makes `collapsed` property public, call `onCollapse`/`onCollapsed` and `onExpand`/`onExpanded` actions. Allows for more flexible customization, while the existing default behavior stays in place. Provides some primitives and a first step for a solution for #317


simonihmig:
@srvance Thanks for reviewing!

PR #358, opened 2017-06-14
Fix documentation for base/bs-button

nilsding:
I just noticed that a wrong property is used in the code examples in the documentation.  Changed it to `onClick`, because `action` will have no effect


simonihmig:
Thanks @nilsding!

PR #359, opened 2017-06-14
fix(Popover): Fix scrollposition for FF

### LINKED ISSUE #357: Wrong position of pop-over in Firefox
pop-overs do not respect scrolled offset in Firefox.
Tested with latest ember-bootstrap 1.0.0-beta.1 on Mac (macOS Sierra) with FF 53.0.2. 
On Chrome and Safari, position is correct.

See this twiddle for reproduction:
https://ember-twiddle.com/54d3b05d8c6f88b2ba530f45a158e864

chvonrohr:
fixes #357

### DIFF
--- a/addon/utils/get-position.js
+++ b/addon/utils/get-position.js
@@ -37,11 +37,10 @@ export default function getPosition(el) {
   return assign({}, clonedRect, scroll, outerDims, elOffset);
 }
 
-
 function getScrollPosition() {
   return {
     //    IE, Firefox                            Chrome
-    top:  document.documentElement.scrollTop  || document.body.scrollTop  || 0,
+    top: document.documentElement.scrollTop || document.body.scrollTop || 0,
     left: document.documentElement.scrollLeft || document.body.scrollLeft || 0
   };
 }
--- a/addon/utils/get-position.js
+++ b/addon/utils/get-position.js
@@ -17,8 +17,8 @@ export default function getPosition(el) {
   // See https://github.com/twbs/bootstrap/issues/20280
   let rect = el.getBoundingClientRect();
   let elOffset = isBody ? { top: 0, left: 0 } : (isSvg ? {} : {
-    top: rect.top + document.body.scrollTop,
-    left: rect.left + document.body.scrollLeft
+    top: rect.top + (window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0),
+    left: rect.left + (window.pageXOffset || document.documentElement.scrollLeft || document.body.scrollLeft || 0)
   });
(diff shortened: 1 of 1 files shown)


addon/utils/get-position.js
```diff
@@ -17,8 +17,8 @@ export default function getPosition(el) {
   // See https://github.com/twbs/bootstrap/issues/20280
   let rect = el.getBoundingClientRect();
   let elOffset = isBody ? { top: 0, left: 0 } : (isSvg ? {} : {
-    top: rect.top + document.body.scrollTop,
-    left: rect.left + document.body.scrollLeft
+    top: rect.top + (window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0),
```
srvance: Might be worth extracting these into functions and documenting the browser dependencies that contribute to it.

chvonrohr: might be a good idea... two lines below scroll-position is also evaluated. But I don't know where it's used...

addon/utils/get-position.js
```diff
@@ -35,3 +36,12 @@ export default function getPosition(el) {
 
   return assign({}, clonedRect, scroll, outerDims, elOffset);
 }
+
+
+function getScrollPosition() {
+  return {
+    //    IE, Firefox                            Chrome
+    top:  document.documentElement.scrollTop  || document.body.scrollTop  || 0,
```
srvance: What happened to `window.page[XY]Offset`?

srvance:
Looks like ESLint issues, so should be simple to fix.

simonihmig:
@chvonrohr Thanks for your contribution! Happy to merge this once the ESLint issues and https://github.com/kaliber5/ember-bootstrap/pull/359#discussion_r121978671 have been resolved!

addon/utils/get-position.js
chvonrohr: I was checking which browsers are targeted by ```window.page[XY]Offset``` but couldn't find a clear answer. As most seem to work with the above two, I added them for now. Tested in FF, Safari, Chrome

simonihmig:
@chvonrohr Thanks for fixing the linting errors! What about `window.pageYOffset`, is this not needed anymore?

chvonrohr:
@simonihmig Seems that ```document.documentElement.scrollTop``` and ```document.body.scrollTop``` cover all major browsers. I couldn't find why its needed, so I removed this one. I added it in the first commit, as it was mentioned in https://stackoverflow.com/questions/28633221/document-body-scrolltop-firefox-returns-0-only-js (see #357).

simonihmig:
Ok, LGTM. @srvance your review seems to have been addressed. Ready to go?

PR #360, opened 2017-06-14
Ensure component is not destroyed before setting textState

elgordino:
I have been getting this error when removing a button immediately after it is clicked

`Assertion Failed: calling set on destroyed object: <lineup-ninja@component:bs-button::ember2411>.textState = pending`

This PR adds an additional check that the component is not destroyed before setting textState

### DIFF
--- a/addon/components/base/bs-button.js
+++ b/addon/components/base/bs-button.js
@@ -259,7 +259,7 @@ export default Ember.Component.extend(TypeClass, SizeClass, {
     let action = this.get('onClick');
     if (action !== null) {
       let promise = (action)(this.get('value'));
-      if (promise && typeof promise.then === 'function') {
+      if (promise && typeof promise.then === 'function' && !this.get('isDestroyed')) {
         this.set('textState', 'pending');
         promise.then(() => {
           if (!this.get('isDestroyed')) {


simonihmig:
The changes certainly do no harm, but I am a bit puzzled how that can happen. 

@elgordino Are you sure it is this [line](https://github.com/kaliber5/ember-bootstrap/pull/360/files#diff-9106abe854aad72249ee0542b0562d99L263) that is causing the error? Because that is executed in the same event loop as the click event, there is no asynchrony in between, so wondering how the component can be already destroyed at that point?

I guess the component gets removed within the `onClick` action? How do you remove it, anything special happening in that action?

elgordino:
@simonihmig it's weird isn't it. I'm sure that's the line that's causing the issue, I'm a lot less sure why!

If I do this

```
      if (promise && typeof promise.then === 'function') {
        if (this.get('isDestroyed')) {
          console.error('about to set textState on destroyed object');
        }
        this.set('textState', 'pending');
```
Then it will log the console error just before throwing the `calling set on destroyed object` error.

The button that is experiencing this problem has this behaviour:

* The button has `onClick=clickHandler` where `clickHandler=(perform clicked)`
* `clicked` (an ember concurrency task) will yield to another task `editRow`, itself supplied by a the perform helper
* `editRow` yields to an async function `startEditing`
* `startEditing` awaits on an async function`makeChangeset` 
* `makeChangeset` awaits on an ember data query
* If it is the first time the ember data query has been executed everything is fine. On the second run the exception is seen.

I use the same button elsewhere, with the only difference being that `makeChangeset` does not wait on an ember data query. It works fine every time.

I attempted to repro the issue in ember-twiddle but wasn't able to :(

At this point I'm thinking there is some weird thing going on between ember data and the run loop, but I'm far from sure!

simonihmig:
Not sure what the actually circumstances are which trigger that error for you, but as I said it does no harm, so merging that for now. Thanks!

elgordino:
@simonihmig Thanks v much!

PR #365, opened 2017-06-17
Automatically collapse navbar menu when clicking nav item

### LINKED ISSUE #317: RFE: Auto collapse Navbar during navigation

simonihmig:
Closes #317

### DIFF
--- a/addon/components/base/bs-navbar.js
+++ b/addon/components/base/bs-navbar.js
@@ -45,15 +45,29 @@ const { observer } = Ember;
  ### Responsive Design
 
  For the mobile breakpoint the Bootstrap styles will hide the navbar content (`{{navbar.content}}`). Clicking on the
- navbar toggle button (`{{navbar.toggle}}`) will expand the menu. By default all NavItems (`{{nav.item}}`) are already
+ navbar toggle button (`{{navbar.toggle}}`) will expand the menu. By default all nav links (`{{nav.link-to}}`) are already
  wired up to call the navbar's `collapse` action, so clicking any of them will collapse the navbar. To selectively
- prevent that, you can override the NavItem's `onClick` action, with a custom action if needed, or just be setting it
- to false:
+ prevent that, you can set its `collapseNavbar` property to false:
 
  ```hbs
- {{#nav.item onClick=false}}
-   {{#nav.link-to "index"}}Don't collapse{{/nav.link-to}}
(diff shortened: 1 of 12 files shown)


simonihmig:
The Navbar's behavior is changed here, as clicking a nav item will now automatically collapse the navbar's content (on mobile). I don't know if that should be regarded a breaking change, or rather a bug fix, as this is how the original Bootstrap Navbar works when used in a non-SPA?

addon/components/base/bs-navbar.js
```diff
@@ -42,6 +42,20 @@ const { observer } = Ember;
  * `collapse`: triggers the `onCollapse` action and collapses the navbar (mobile)
  * `expand`: triggers the `onExpand` action and expands the navbar (mobile)
 
+ ### Responsive Design
+
+ For the mobile breakpoint the Bootstrap styles will hide the navbar content (`{{navbar.content}}`). Clicking on the
+ navbar toggle button (`{{navbar.toggle}}`) will expand the menu. By default all NavItems (`{{nav.item}}`) are already
+ wired up to call the navbar's `collapse` action, so clicking any of them will collapse the navbar. To selectively
+ prevent that, you can override the NavItem's `onClick` action, with a custom action if needed, or just be setting it
```
srvance: Edit: "or just set it"

danidr:
I would like to point out two things:

1. Nav item could be used for other elements than links; therefore, there are many realistic cases where the onClick should be set to false manually. This would also break backwards compatibility.
2. Nav Item in bootstrap has a padding to avoid having link text overlapping the border, for example in case of dropdown menus. This would bring a case where clicking on an item would collapse the menu (since the item registers the onClick action) but the actual link wouldn't trigger because the user has clicked just a wee bit outside of it, therefore confusing the user as to what has happened.

simonihmig:
@danidr I took your feedback into account, and changed the API accordingly, see https://github.com/kaliber5/ember-bootstrap/pull/365/files#diff-a0ee6ab773729a34ca7901616a846cbe. 

@danidr @srvance Mind having another look?

danidr:
@simonihmig If nobody has any comment about this implementation, I would say it looks good to me!

Issue #248, opened 2017-02-24
Fix BS4 `active` state in tabs and dropdown items


srvance:
This is waiting for the BS4 beta to see how they resolve the known `active` issues.

From #206,

* Why aren't default tabs initial state outlined properly? Diagnosis: The `active` class needs to go on the element with the `nav-link` rather than the `nav-item`. See [this](https://github.com/twbs/bootstrap/issues/21223) and [this](https://github.com/twbs/bootstrap/issues/20156).
  * Change 1caa0ebb2c9aabfb9c0f9896301b1a00a532ee82 has a little leakage of BS4 classes into common templates with `nav-link`. To fix the `active` state issue, there's even more. However, it's a pretty complex template to duplicate and i'm not sure refactoring out a new component is worth it.
  * The same kind of issue exists with dropdown items within a nav.

simonihmig:
With all the latest changes to BS4 beta support, I *think* this should be fixed by now, so closing this...

PR #366, opened 2017-06-18
Make dropdowns BS4 compliant

### LINKED ISSUE #337: bs4 dropdown still uses ul element
the bs4 dropdown does not use a ul element anymore
the dropdown menu just contains the a every menu item, title or divider

bs4: https://v4-alpha.getbootstrap.com/components/dropdowns/#single-button-dropdowns
bs3: http://getbootstrap.com/components/#dropdowns-example

_side note_
would be nice if alignment is configurable with a prop

srvance:
Fixes #337. Continuing the conversation here.

I don't see a way to do this without a breaking change. That's not ideal since we've gone to beta, but I think it's necessary.

On rereading the docs, the `<a>` or `<button>` decision is about the dropdown button, not the item. For items, they only mention links or dividers. We have dividers covered, leaving us with two types of links, internal and external. I'm not worried separately about dropdown items that are plain text as that can be covered with a disabled item.

I'm going to create `bs-dropdown.menu.link-to` for the internal links.

External links need more thought as their tag structure between the two versions is significantly different.

### DIFF
--- a/tests/integration/components/bs-dropdown/menu/link-to-test.js
+++ b/tests/integration/components/bs-dropdown/menu/link-to-test.js
@@ -3,32 +3,30 @@ import { moduleForComponent } from 'ember-qunit';
 import { testBS3, testBS4 } from '../../../../helpers/bootstrap-test';
 import hbs from 'htmlbars-inline-precompile';
 
-moduleForComponent('bs-dropdown/menu/item', 'Integration | Component | bs-dropdown/menu/item', {
+moduleForComponent('bs-dropdown/menu/link-to', 'Integration | Component | bs dropdown/menu/link to', {
   integration: true
 });
 
 testBS3('it has correct markup', function(assert) {
   // Template block usage:
   this.render(hbs`
-    {{#bs-dropdown/menu/item}}
+    {{#bs-dropdown/menu/link-to}}
       template block text
-    {{/bs-dropdown/menu/item}}
+    {{/bs-dropdown/menu/link-to}}
   `);
 
-  assert.equal(findAll('li').length, 1, 'renders as <li> element');
   assert.equal(find('*').textContent.trim(), 'template block text');
+  assert.equal(findAll('a.dropdown-item').length, 0, 'renders as plain element with no dropdown item class');
 });
 
 testBS4('it has correct markup', function(assert) {
   // Template block usage:
   this.render(hbs`
-    <span>
-    {{#bs-dropdown/menu/item}}
+    {{#bs-dropdown/menu/link-to}}
       template block text
-    {{/bs-dropdown/menu/item}}
-    </span>
+    {{/bs-dropdown/menu/link-to}}
   `);
 
-  assert.equal(findAll('div').length, 0, 'renders as no element');
   assert.equal(find('*').textContent.trim(), 'template block text');
+  assert.equal(findAll('a.dropdown-item').length, 1, 'renders as plain element with dropdown item class');
 });
--- a/tests/integration/components/bs-tab-test.js
+++ b/tests/integration/components/bs-tab-test.js
@@ -142,12 +142,12 @@ test('tab navigation is groupable', function(assert) {
   assert.equal(find('ul.nav.nav-tabs > li:nth-child(2) > a').textContent.trim(), 'Tab 2', 'navigation item shows pane title');
   assert.equal(find('ul.nav.nav-tabs > li:nth-child(3)').classList.contains('dropdown'), true, 'adds dropdown for grouped items');
   assert.equal(find('ul.nav.nav-tabs > li:nth-child(3) > a').textContent.trim(), 'Dropdown', 'drop down item shows pane groupTitle');
-  assert.equal(findAll('ul.nav.nav-tabs > li:nth-child(3) > ul.dropdown-menu > li').length, 2, 'puts items with groupTitle under dropdown menu');
+  assert.equal(find('ul.nav.nav-tabs > li:nth-child(3) > .dropdown-menu').children.length, 2, 'puts items with groupTitle under dropdown menu');
   assert.equal(find(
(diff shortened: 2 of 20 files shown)


simonihmig:
@srvance if I understand your intention correctly, you want to change this...

```hbs
{{#menu.item}}
  {{#link-to "index"}}Home{{/link-to}}
{{/menu.item}}
```

... to this...

```hbs
{{#menu.item}}
  {{#menu.link-to "index"}}Home{{/menu.link-to}}
{{/menu.item}}
```

Correct?

That does not need to be a breaking change, does it? For BS3 the first example would still work, but maybe not be the prefered way to do this, as it would prevent an easy upgrade path to BS4. And as BS4 support is still considered experimental, that would not be considered a breaking change, would it?

srvance:
I was thinking of changing

```handlebars
{{#menu.item}}
  {{#link-to "index"}}Home{{/link-to}}
{{/menu.item}}
```

to

```handlebars
{{#menu.link-to "index"}}Home{{/menu.link-to}}
```

but I like your suggestion better. It simplifies the change.

It's still breaking in that it requires a change of usage to be mutually compatible, but it's backwards compatible.

srvance:
@simonihmig Thank you for the insight. That simplified things tremendously. I was trying to wrap it all into `menu.item`.

@gijsbotje I can't explicitly request a review from you, but you should be able to contribute one. I would love it if you could point your app at this branch version of e-b and see if it addresses your concerns properly.

tests/integration/components/bs-tab-test.js
```diff
@@ -142,12 +142,45 @@ test('tab navigation is groupable', function(assert) {
...
   assert.equal(find('ul.nav.nav-tabs > li:nth-child(3) > a').textContent.trim(), 'Dropdown', 'drop down item shows pane groupTitle');
-  assert.equal(findAll('ul.nav.nav-tabs > li:nth-child(3) > ul.dropdown-menu > li').length, 2, 'puts items with groupTitle under dropdown menu');
+  assert.equal(find('ul.nav.nav-tabs > li:nth-child(3) > .dropdown-menu').children.length, 2, 'puts items with groupTitle under dropdown menu');
   assert.equal(find(
-    'ul.nav.nav-tabs > li:nth-child(3) > ul.dropdown-menu > li:nth-child(1) > a'
+    'ul.nav.nav-tabs > li:nth-child(3) > .dropdown-menu > :nth-child(1) > a'
   ).textContent.trim(), 'Tab 3', 'dropdown menu item shows pane title');
   assert.equal(find(
-    'ul.nav.nav-tabs > li:nth-child(3) > ul.dropdown-menu > li:nth-child(2) > a'
+    'ul.nav.nav-tabs > li:nth-child(3) > .dropdown-menu > :nth-child(2) > a'
```
simonihmig: If you omit the last part of the selector (`> a`) here and on the statement before, this should still work, doesn't it? This is the only difference between the BS3 and BS4 tests, or did I miss something else? This way we could maybe unify the tests to avoid too much duplication?

tests/integration/components/bs-dropdown/menu/link-to-test.js
```diff
@@ -0,0 +1,32 @@
...
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('bs-dropdown/menu/link-to', 'Integration | Component | bs dropdown/menu/link to', {
+  integration: true
+});
+
+testBS3('it has correct markup', function(assert) {
+  // Template block usage:
+  this.render(hbs`
+    {{#bs-dropdown/menu/link-to}}
```
simonihmig: Seems tests are failing because of the missing route parameter ("index" should work) for the link component!

simonihmig:
Oh btw, how do the changes here relate to #248, do they fix that issue as well?

gijsbotje:
my colleague @Rapid0o and were wondering why you wouldn't make a seperate v4 template
most of the bs4 elements have different structure from bs3
So wouldn't it be obvious to make seperate templates?

srvance:
@gijsbotje The differences have to do with component tags and classes, which don't require a template. I considered a template approach, which is what I was struggling with until @simonihmig's suggestion. The challenge there is coming up with a component definition that works for both BS3 and BS4 but still has all the flexibility of `{{link-to}}` and `<a>`. This would require accepting all attributes and passing them down appropriately, whereas this solution just requires slight changes to the component definitions and the new contextual definition.

srvance:
@simonihmig re #248, it might fix it. I'll have to check.

gijsbotje:
@srvance we looked at it again you're right :)
the changes to the tags and classnames is exactly what we where hoping for
@Rapid0o is pulling it to local as I type

Rapid0o:
@srvance These changes are exactly what we meant. It is according to the (alpha/beta) bs4 specs and addresses our concerns perfectly. You're right this is better than creating different templates for bs3 and bs4 :+1:

simonihmig:
@srvance Seems this is ready to go then! 🚀

What about this comment here: https://github.com/kaliber5/ember-bootstrap/pull/366#discussion_r122792275? This is not a major concern to me, but you might want to take a look? Otherwise changes are approved, so feel free to merge once you are happy with it! 😀

simonihmig:
@srvance Oh btw, when you merge, please squash the commits and create a commit message according to https://conventionalcommits.org/, (like `fix(Dropdown): some message`), so the changelog is populated automatically. I should write a `CONTRIBUTING.md` I guess...

srvance:
@simonihmig I'll take a look at the tab test comment this evening.

tests/integration/components/bs-tab-test.js
srvance: You're correct. The relaxed selector necessary to satisfy BS4 also works for BS3. The only disadvantage is that it's less specific about the DOM that's expected.

PR #368, opened 2017-06-19
Fix(Form): model is passed to isInvalid action as first argument

jelhan:
Model is passed to isInvalid since e1397a1a58690ccfbc291d5f98d7923091d40ec0 but documentation was not changed.


simonihmig:
Thanks!

PR #369, opened 2017-06-21
Add static CSS fix to enable clicking on whole header

### LINKED ISSUE #355: Fix for opening accordion item on iOS devices
Thanks to @sukima for the fix

Before the ember-bootstrap beta.1 release, on iOS devices (or at least, iOS 10 anyway), clicking the accordion item anywhere didn't open the accordion.

On beta.1, it was now possible to click the title to open the accordion item (but clicking the accordion item itself still did not work.) This was likely due to this fix https://github.com/kaliber5/ember-bootstrap/issues/311

It turns out, that iOS devices refuse to click on elements that aren't `<button>` or `<a>`. To allow clicking on other elements (such as `<div>`) need to apply the style `cursor: pointer;` to the element - doing this fixes the issue on iOS.

Should this style/class be added by default to accordion items so that they work on iOS? I can probably to a PR if needed

simonihmig:
Fixes #355

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,11 +51,10 @@ module.exports = {
     this.validateDependencies();
     this.preprocessor = this.findPreprocessor();
 
+    // static Bootstrap CSS is mapped to vendor tree, independent of BS version, so import from there
+    let cssPath = path.join('vendor', 'ember-bootstrap');
+
     if (!this.hasPreprocessor()) {
-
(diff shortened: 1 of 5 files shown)


srvance:
@simonihmig Failure was a `curl` failure installing `yarn`. Can you just kick that single build job to run again?

simonihmig:
done!

PR #370, opened 2017-06-22
insertAdjacentElement doesn't exist in Firefox <= 47

HodofHod:
This was added when the jQuery dependency was removed. It breaks in any Firefox version before 48, which is when support for insertAdjacentElement was added.

Not sure if you care about supporting a year-old version with negligible market share, but it's two lines, so I  figured I'd fix first and ask later.


simonihmig:
Sure, thanks @HodofHod!

PR #375, opened 2017-07-08
Add bs-carousel

### LINKED ISSUE #138: Carousel coming soon?

c410-f3r:
Fixes #138.

Does everything the original bootstrap carousel did but API is different, take a look into the docs.
I tried really hard to include the `render` and `afterRender` queues into the sliding process but I gave up when I started to pull my hairs off, instead, a manual reflow with `Ember.run.next` was used.
To support page visibility the addon `ember-in-viewport` is a nice approach but I guess it doesn't work without jquery.


simonihmig:
Hey @c410-f3r, that's a massive PR. Absolutely awesome that you found the time to contribute this! 👏

I just had a quick first glimpse at it, looks like really a high quality contribution. I will need some more time however to thoroughly review it. Just wanted to say thanks in advance! 😃

c410-f3r:
Thanks! The structure was already laid out so I didn't have much trouble implementation this feature.
Feel free to indicate any objections or errors to be fixed so we can merge this ASAP.

addon/components/base/bs-carousel.js
```diff
@@ -0,0 +1,467 @@
...
+   * Applies class names to slides, sets the current index as the following index and
+   * indicates that the transition has finished.
+   * 
+   * @method didTransition
+   * @private
+   */
+  didTransition(currSlide, followingSlide) {
+    currSlide.set('active', false);
+    currSlide.set(this.get('directionalClassName'), false);
+    followingSlide.set('active', true);
```
c410-f3r: Thanks for the review, I hope that I will be able to fix everything ASAP.
Regarding this comment, the slides should change its other properties (next, left, right, prev) as well?

c410-f3r: Well, kept the same strategy since the slides instances play nicely with the component-(child|parent) mixins and I added a real time iterative carousel that allows dynamic inserting and removing.
The code is not ready to merge yet but I think we are getting close.

addon/templates/components/common/bs-carousel.hbs
```diff
@@ -1,17 +1,11 @@
-{{#if showIndicators}}
-  <ol class='carousel-indicators'>
-    {{#each indicators as |indicator|}}
-      <li class='{{if indicator.active 'active'}}' onclick={{action 'toSlide' indicator.index}} role='button'></li>
-    {{/each}}
-  </ol>
-{{/if}}
+<ol class='carousel-indicators'>
```
simonihmig: I think you still need to wrap this in the previous `{{#if}}` block, to not have a `<ol>` without any `<li>` if `showIndicators` is false.

addon/components/base/bs-carousel.js
```diff
@@ -58,13 +138,28 @@ export default Ember.Component.extend({
...
+      indicators.push({
+        active: false,
+        index: a
+      });
+    }
+    if (indicators.length > 0) {
+      indicators[this.get('_index')].active = true;
+    }
+    return indicators;
+  }),
```
simonihmig: I think this could be refactored a bit, to make it simpler. And because it depends on `_index`, it will create a new array every time the index changes, so causing a full rerender of all indicators. How about this:

```
indicators: computed('childSlides.length', function() {
  return new Array(this.get('childSlides.length'));
}
```

This would need some changes in the template as well:

```hbs
{{#if showIndicators}}
  <ol class='carousel-indicators'>
    {{#each indicators as |indicator index|}}
      <li class='{{if (bs-eq _index index) 'active'}}' onclick={{action 'toSlide' index}} role='button'></li>
    {{/each}}
  </ol>
{{/if}}
```

This makes use of the existing `bs-eq` helper to set the active class directly in the template, and `#each` has a built-in index, so no need to put that into the array...

c410-f3r: Oh man... this solution is so much better and cleaner, thaaaankkssss!

c410-f3r:
Hey @simonihmig, thanks for your support reviewing this code.
Things are now working as expected but there are a few topics remaining:

- [ ] Add more tests
- [ ] Double check async stuff
- [ ] Improve performance (Rerenders, begin()/end() and queues)
- [ ] Support page visibility

I guess it is going to take a while :(

simonihmig:
> I guess it is going to take a while :(

@c410-f3r no problem. Feel free to get in touch if I can help with anything!

c410-f3r:
Most tests need to wait a transition time duration to verify correct state, therefore, the whole testing time is significantly increased.

c410-f3r:
Branch is messed up

PR #385, opened 2017-07-24
Tree shaking

### LINKED ISSUE #140: Add tree shaking support

simonihmig:
Closes #140

To Do:

- [x] support dependencies between components
- [x] add docs
- [x] add e2e test

### DIFF
--- a/fastboot-tests/fixtures/treeshake/ember-cli-build.js
+++ b/fastboot-tests/fixtures/treeshake/ember-cli-build.js
@@ -1,0 +1,11 @@
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, {
+    'ember-bootstrap': {
+      whitelist: ['bs-button-group']
+    }
+  });
+
+  return app.toTree();
+};
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ const mv = stew.mv;
 const rm = stew.rm;
 const map = stew.map;
 const rename = stew.rename;
+const BroccoliDebug = require('broccoli-debug');
 const chalk = require('chalk');
 const SilentError = require('silent-error'); // From ember-cli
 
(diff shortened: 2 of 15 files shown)


simonihmig:
@srvance added you as a reviewer, as this was touching some build hooks, and also I refactored the docs a bit (including assets guide). Going to merge this now, as I want to cut a new release. But should you have time to look at this and find something, feel free to nevertheless leave a comment! ;)

index.js
```diff
@@ -28,9 +29,25 @@ const supportedPreprocessors = [
   'sass'
 ];
 
+const componentDependencies = {
```
srvance: Mixins and helpers are automatically included?

fastboot-tests/fixtures/treeshake/ember-cli-build.js
```diff
@@ -0,0 +1,11 @@
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, {
+    'ember-bootstrap': {
+      whitelist: ['bs-button-group']
```
srvance: Seems like we should add to the blueprint, as well, at least to create empty arrays. Is there a `blacklist` config, as well? It would require adding array value support to ember-cli-build-config-editor.

index.js
simonihmig: Yes, somewhere is restricts the filter to just components and its templates. So when excluding everything, those things will still make it to the build, but checking if/which of these are needed based on the selection of whitelisted components would be a bit complicated, probably not worth the effort, and very brittle (we would need extensive tests to not suddenly break things)

fastboot-tests/fixtures/treeshake/ember-cli-build.js
simonihmig: Yes, there is a `blacklist`, but the current implementation relies on having either a `whitelist` or a `blacklist`, not both. And adding `whitelist: []` would exclude all components form the build, so probably not a good default! 😝

I don't know if it would be possible to add those, but commented out? I don't think the is a way to do that with AST, is there? But I personally am ok with how it is now. Users can still add those options if they need them...

srvance: Sounds good. Maybe we want to have an `all` alias that we could default to. I'm fine going without the blueprint. I'll create an issue in both projects to address it.

PR #386, opened 2017-07-26
Remove transitionEnd eventlistener

lodrantl:
Because the transitionEnd event isn't always triggered on the root component node (if modal contains some other transitions), it wasn't always removed.


simonihmig:
Thanks again!

PR #387, opened 2017-07-28
Use better approach to find app object in included hook of index.js

synaptiko:
I have `ember-addon` which depends on `ember-bootstrap` directly. I want to use `ember-addon` in my `ember-app`. When I run `ember build` in this `ember-app` it ends with the following error:
```
Cannot read property 'ember-bootstrap' of undefined
TypeError: Cannot read property 'ember-bootstrap' of undefined
    at Class.included (/home/jprokop/work/sandbox/ember-addon-common/node_modules/ember-bootstrap/index.js:27:53)
    at /home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/models/addon.js:254:34
    at Array.map (native)
    at Class.eachAddonInvoke (/home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/models/addon.js:252:24)
    at Class.included (/home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/models/addon.js:429:10)
    at Class.superWrapper [as included] (/home/jprokop/work/git/ember-app-sme/node_modules/core-object/lib/assign-properties.js:32:18)
    at EmberApp.<anonymous> (/home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/broccoli/ember-app.js:470:15)
    at Array.filter (native)
    at EmberApp._notifyAddonIncluded (/home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/broccoli/ember-app.js:465:45)
    at new EmberApp (/home/jprokop/work/git/ember-app-sme/node_modules/ember-cli/lib/broccoli/ember-app.js:138:8)
```

I did a little research and found that there is already some workaround in `ember-bootstrap` code: https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L58

I checked few other addons and found out that for example `ember-power-select` contains following: https://github.com/cibernox/ember-power-select/blob/master/index.js#L20-L21

I used it as an inspiration and created this pull request. It worked in my case. Please consider to use this way instead of the current workaround.


synaptiko:
This PR has conflict because I used v0.11.2 which I currently depend on. But I'll resolve the conflict in case you decide to incorporate my change.

simonihmig:
@synaptiko Thanks for working on this. I would be ok with the changes, as the previous "solution" was trying to allow using e-b as a nested addon. However I am not sure if that is actually the best way to do it. Nested addons seem to be not entirely supported in ember-cli, as for example you cannot run the blueprints of a nested addon, AFAIR. So e.g. `ember g ember-bootstrap --bootstrap-version=4` would not work for switching the BS version, as ember-bootstrap is not a direct dependency of the project.

An alternative would be to install ember-bootstrap automatically using the other addons' default blueprint, using https://ember-cli.com/api/classes/Blueprint.html#method_addAddonToProject. That way the app could also specify the version of ember-bootstrap required.

synaptiko:
@simonihmig I also already found out that support for nested addons is in some ways uncomplete. The alternative you propose is something we are considering to do in our addon too.

Now I've checked again my change and I'm not sure if it covers the same cases as the previous code. What do you think? Maybe it will break something. It would be good to have some example or test case to verify it.

simonihmig:
@synaptiko I think that should be ok. There are FastBoot tests that test building and serving an app end 2 end, so if it would not compile the app correctly I *think* those would uncover it. There are no tests for different ember-cli versions though, but I am still comfortable that this should not break anything.

There are however two little ESLint errors, see the Travis log. :)

synaptiko:
@simonihmig fixed

simonihmig:
Merged, thanks @synaptiko!

PR #390, opened 2017-08-02
Button orientation on Windows.

lodrantl:
As a lifelong Windows user it bothered me really hard that Submit and Close buttons were oriented differently. What do you think guys?

### DIFF
--- a/addon/templates/components/common/bs-modal/footer.hbs
+++ b/addon/templates/components/common/bs-modal/footer.hbs
@@ -2,8 +2,13 @@
   {{yield}}
 {{else}}
   {{#if hasSubmitButton}}
-    {{#bs-button onClick=(action onClose)}}{{closeTitle}}{{/bs-button}}
-    {{#bs-button type="primary" onClick=(action onSubmit) disabled=submitDisabled}}{{submitTitle}}{{/bs-button}}
+    {{#if _isPlatformWindows}}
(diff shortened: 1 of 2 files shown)


srvance:
Does ember-bootstrap behave differently from Bootstrap in this regard?

addon/templates/components/common/bs-modal/footer.hbs
```diff
@@ -2,8 +2,13 @@
   {{yield}}
 {{else}}
   {{#if hasSubmitButton}}
-    {{#bs-button onClick=(action onClose)}}{{closeTitle}}{{/bs-button}}
-    {{#bs-button type="primary" onClick=(action onSubmit) disabled=submitDisabled}}{{submitTitle}}{{/bs-button}}
+    {{#if _isPlatformWindows}}
```
srvance: The change in ordering fixes the styling issue you bring up? What browsers?

simonihmig:
@lodrantl Thanks for your effort, however I don't think I want to merge this. This is very specific to your needs. I don't think everybody expects to have their modals look a bit different based on the users platform, people might have their own styleguide rules about that. Right now the default is what is also the documented markup in Bootstrap: http://getbootstrap.com/javascript/#static-example

If users want different behavior, this can be easily customized in app-space, by extending `bs-modal-simple` and/or overriding its template.

Issue #391, opened 2017-08-02
BS-Modal Form Submission Breaks in Old Browsers


zmnatz:
The following error is produced when submitting a modal form in Firefox 45 and previous versions:

forms.forEach is not a function

This seems to be caused by the following line of code:
https://github.com/kaliber5/ember-bootstrap/blob/b2b32b8ba7d4ded6d77dbf882b2da0d07825b699/addon/components/base/bs-modal.js#L345

PR #392, opened 2017-08-02
#391 Fix NodeList.forEach Browser Compatibility

zmnatz:
Using for...of Loop instead of


srvance:
Looks like some lint issues to clean up to pass.

simonihmig:
Yeah, ESLint is complaining about a missing space after `for`... @zmnatz could you please clean that up, thanks!

PR #400, opened 2017-08-04
Fix failing Safari tests

### LINKED ISSUE #397: Failing tests for Safari

simonihmig:
Fixes #397

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,19 +20,19 @@ env:
     - secure: QKF4gJdgWzwTk6eMMVZvP+FBs7Ab3fFYnXfLHHNbQOkIVrwO0D7ns3Umtea9lMIt1bQOq6VHmcsf/BUhi4L+oLEXAZ827GShDKhojqGIDg4poaA9kECtUzsIUnKB+bYkvadRsuUP0KjxNbSM7nVchUDTgRvt0OZZ0rP3q4/oQC8=
 
   matrix:
-#    - EMBER_TRY_SCENARIO=ember-2.3
-#    - EMBER_TRY_SCENARIO=ember-lts-2.4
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -18,26 +18,26 @@ env:
...
-    - EMBER_TRY_SCENARIO=ember-beta
-    - EMBER_TRY_SCENARIO=ember-canary
-    - EMBER_TRY_SCENARIO=ember-lts-2.4-bs4
-    - EMBER_TRY_SCENARIO=ember-lts-2.8-bs4
-    - EMBER_TRY_SCENARIO=ember-lts-2.12-bs4
-    - EMBER_TRY_SCENARIO=ember-release-bs4
-    - EMBER_TRY_SCENARIO=ember-beta-bs4
-    - EMBER_TRY_SCENARIO=fastboot-addon-tests
-    - EMBER_TRY_SCENARIO=node-tests
+#    - EMBER_TRY_SCENARIO=ember-2.3
```
srvance: Is this WIP? Why leave the commented configs?

simonihmig: yes, WIP. Wanted to have quick feedback from Travis...

simonihmig:
@srvance This should be ready to go now. 

For the record: The issues with testing Safari w/ Saucelabs came from the fact, that those WebDriver controlled browsers in those VMs seem to not always have "focus", that way functionality relying on focusing things (like modals putting focus on themselves to catch ESC-key events) lead to failing tests. Some background: https://shanetomlinson.com/2014/test-element-focus-javascript/

For now I took the same approach as suggested there, and disabled certain tests that need `focus` to work, when `document.hasFocus()` is false. Rather a workaround than a real fix, but what to do else? I guess somethings things are a little bit messy... 🙄

Issue #249, opened 2017-02-24
Explore contextualizing form element pieces


srvance:
Adding form help text per #169 and the discussion in #206 [here](https://github.com/kaliber5/ember-bootstrap/pull/206#issuecomment-282051230) suggest we should look into contextualizing some of the form element pieces for use directly. This issue is to capture the intent to explore that.

srvance:
I don't see that I'll have time to dive into this in the foreseeable future. Closing.

PR #410, opened 2017-08-13
Update to `bootstrap@4.0.0-beta`

srvance:
Updating to BS4 beta, all tests pass.

The following issues are observed in the components in the demo app:

- [x] Accordion:
  - [x] Change `card-block` to `card-body`
  - [x] fix failing tests 
- [x] Alert: none in the functionality
  - [x] The spacing on the forms between the checkboxes is too tight
  - [x] The dropdown resizes in a way that crops the longer text when the shorter options are selected (This seems to be an `ember-power-select` issue and unrelated to BS4)
- [x] Button: none
- [x] Button group: none
- [x] Collapse: none
- [ ] Dropdown:
  - [x] Neither regular nor button dropdowns work (Reason: The `show` class moved to the menu)
  - [x] The regular dropdown spacing seems too close together (See Forms)
  - [x] Dropup drops down (Reason: I think this has to do with the use of Popper and how BS4 defers several elements of styling to it)
  - [x] fix flicker of dropup
  - [ ] TODO: Specialize dropdown toggle so that the caret span is unnecessary
- [ ] Forms:
  - [ ] The inline spacing in the first example is too tight (Reason: BS4 requires the addition of the [spacing](https://getbootstrap.com/docs/4.0/utilities/spacing/) classes. We need to apply defaults and a means to override.)
  - [x] verify validation markup
- [x] Modal: none
- [x] Navbar:
  - [x] Shows as responsive and collapsed by default (Reason: Class was changed from `navbar-toggleable-*` to `navbar-expand-*`
  - [x] Menu icon empty (Reason: This is actually an issue with the way the toggle passes the color scheme into the navbar in the demo app. It could also be seen as a bug in how `bs-navbar` handles `null` or `undefined` values as its defaults, so that has been fixed.)
  - [x] Checkbox spacing is too tight
- [x] Nav: none in the functionality
  - [x] Form spacing is too tight (checkboxes are okay now) (See Forms)
- [x] Popover:
  - [x] Styling looks wrong, but content and location are fine
- [x] Progress: none in the functionality
  - [x] No way to restore default color from the `Type` dropdown (This seems to be an `ember-power-select` issue)
  - [x] TODO: Dependencies between options should be enforced on the form
- [x] Tabs
  - [x] Dropdown for grouped tabs doesn't work
  - [x] Not sure if custom tabs are working correctly
- [x] Tooltip
  - [x] Tooltips have no arrows
  - [x] Clicking on one of the first four tooltips makes it do you can't dismiss the tooltip without toggling the clickable tooltip

Additional:
- [ ] Move BS4-specific component docs to `base`
- [ ] Add tests for responsive breakpoints


simonihmig:
Hey @srvance, thanks for tackling this!

I am just wondering if looking at the demo app is the best way to approach this. I think it is still to coupled to BS3, as it is just meant to drive the docs site. Given that, I think that we might get too many "false positives" when looking at it for visual issues?

Shouldn't we just compare the rendered output of the (BS4) component and compare that to the markup the BS docs site suggests, and see if there is some mismatch?

srvance:
I've had good results in the past running the demo app with `BOOTSTRAPVERSION=4 ember s` to try things out. I'll take the approach you suggest, as well.

So far, I'm wondering if the docs have completely caught up with the implementation and how much of the implementation was moved into the js. For example, see the first bullet [here](https://getbootstrap.com/docs/4.0/migration/#responsive-utilities) and note that there is a lot of raw styling appearing now when you toggle a dropdown in the BS docs.

Anyway, this may take awhile. The changes are significant. I'm just starting the investigation. Any findings at this point should be taken as tentative. 😄

srvance:
@simonihmig It appears that things like dropups, popovers, and tooltips are now being handled by Popper.js, which injects styling directly onto the element. There's an ember-popper, still in alpha, that we could use for the same effect. It seems BS4 has delegated that part of the styling, though, only using Popper to manage it. I filed twbs/bootstrap#23399 to see if I'm missing something.

simonihmig:
@srvance What kind of styling do you see managed by popper.js? I just had a look, and it seems to me this is just limited to positioning things. So e.g. for popovers and tooltips this are similar CSS props like `top` or `left` but now managed by popper.js instead of an own implementation as it was for BS3 (Btw, for 4-alpha seems this was managed by tether.js). For Dropdowns it seems to be indeed substantially different, as it required now some inline style attribute, whereas before this was just limited to applying classes.

srvance:
Definitely positioning styling, but also something complicated to replicate. What do you think of bringing in ember-popper to bridge the gap and helping to bring it to release? Or would you rather just bring popper.js in directly?

joshuairl:
Anybody get this error:

```
Error: Undefined operation: "calc(2.25rem + 2px) lt 0".
```

Getting this after updating to this branch.

joshuairl:
Accordion seems to be wrong, the checklist says there are no changes but `card-block` is not the class for accordion body.  It's `card-body` now.  Otherwise, padding is lost on the left and right.

This is true for a lot of class names across the system, it's pretty significant from my knowledge from alpha6.

srvance:
Thanks for the sharp eye, @joshuairl! Let me know if you see anything else.

srvance:
Fixes #421

srvance:
@simonihmig Two things:

1. Would you mind looking at the form spacing issue mentioned above? See the comment for some analysis.
1. How do you feel about a dependency on `ember-popper` to have the same behavior as BS4?

addon/components/bs4/bs-navbar.js
```diff
@@ -4,7 +4,19 @@ import Navbar from 'ember-bootstrap/components/base/bs-navbar';
...
+    },
+
+    set(key, value) { // eslint-disable-line no-unused
+      let newValue = (!value || value === 'default') ? 'light' : value;
+      this.set('appliedType', newValue);
+      return newValue;
+    }
+  }),
+
+  appliedType: 'light',
```
simonihmig: Why is this separate property needed?

simonihmig:
> Would you mind looking at the form spacing issue mentioned above? See the comment for some analysis.

Sorry, but which comment are you referring to? Might be missing something...

> How do you feel about a dependency on ember-popper to have the same behavior as BS4?

Yeah, sorry for the late feedback on that topic... I think that is acceptable. popper.js itself and the ember addon seem solid, tested, small. The addon also supports older Ember versions (even 1.11+), popper is tested for all browsers, including IE10+ (don't know about IE9).

Some questions though:
* should this be used only for BS4 components, while the BS3 still relies on our own proprietary code? In that case we should make sure to not add bloat, like adding popper for BS3, or shipping our BS3-positioning code for BS4.
* Or we could refactor also the BS3 implementations to drop our custom positioning code and delegate to popper. I guess that would be ok, as long as it positions things the same way it is expected for BS3

addon/components/bs4/bs-navbar.js
srvance: `type` is used by the `TypeClass` mixin and I needed somewhere to store the value while giving it a default value. I got tired of wonder why the hamburger menu didn't show when running the demo app in BS4. It's bitten me three times already. It's because it sets `type` to `undefined` if the checkbox is unchecked. Perhaps I shouldn't override the use of `default`, though. What do you think?

srvance:
> which comment are you referring to?

Yeah, I was a little vague. In the todo list above, the comment about spacing next to the "Forms" sub-item.

srvance:
re popper, let me see if we can use it for both and reduce the code.

addon/components/bs4/bs-navbar.js
simonihmig: I might be missing something, but can't you just return a literal 'light' in the CP's getter, and remove the `this.set` call in the setter?

simonihmig:
Towards inline form spacing: that's hard with our current API. I wouldn't like to have to introduce some very explicit properties like `inlineLabelSpacing` or `inlineControlSpacing`. On the other hand we don't expose the things like labels and control elements as contextual components with our current API. Would that be the case users could just apply simple assignments like `{{el.label class="mb-2 mr-sm-2 mb-sm-0"}}` which seems much better to me than introducing just another layer of complexity.

I would tend to "postpone" this for now. I don't know but I actually never used inline forms, and if you do need them, you can still implement them with your own custom markup, nobody forces you to use those `form.element`. One of their stronger points is the markup generation for validation, and I have so far not seen any example of custom validation (not the HTML5 built-in) in an inline form. So maybe it makes more sense to as I said postpone this for now and come up with some different/additional form API that gives you more control over the markup? Somewhat related to https://github.com/kaliber5/ember-bootstrap/issues/249 I guess?

What do you think?

Regarding the popper thing, If you want me to help or take this over (like exploring using that as the replacement for our existing BS3-based custom positioning), just say so. You should not think that you are obliged to work on all this upcoming BS4 stuff exclusively! 😉

srvance:
I'm fine deferring the spacing. Inline forms are mostly what's used in the demo app for the meta-controls, which is where this becomes apparent. We should probably just add some spacing there if we want to support it in BS4.

I won't have much time to look into popper until next weekend. Feel free to look into it if you want. That will cover a large amount of what's left.

How do you feel about a compatible API change for the dropdown toggle?

addon/components/bs4/bs-navbar.js
srvance: That was my first approach, but I needed the setter to redirect the set anyway. I'll look at it again and see if it can be reduced or simplified.

basz:
hio, i don't think it is possible at the moment (bs4-beta branch) to use a `navbar-expand` class on navbars. So no extension to avoid breaking al together


from https://getbootstrap.com/docs/4.0/components/navbar/
> "Navbars require a wrapping .navbar with .navbar-expand{-sm|-md|-lg|-xl} for responsive collapsing and color scheme classes."

simonihmig:
@srvance I merged #437, so this can be rebased to master to bring in the new popper integration.

Do you want to continue work on this, or should I tackle the remaining tooltip, popover and dropdown issues, given they are related to this popper integration? (*I'm ok with both*)

sdhull:
Not sure if this is being tracked and I just didn't see an item for it, but BS4 significantly [changes how form validation errors are displayed](https://getbootstrap.com/docs/4.0/components/forms/#validation). For instance, `.has-danger` class does nothing now.

And on a ease-of-use note, it would be awesome if errors added to a model (eg `model.errors.attributeName`) by ember-data were simply displayed by default.

simonihmig:
I rebased this to current master (includes ember-popper), and fixed the markup of tooltips and popovers. They should be fine now! cc @srvance

drewcovi:
@simonihmig any updates on this one? looks like it's failing.... thanks for all the contributions here!!

simonihmig:
@drewcovi still WIP...

basz:
popup with direction up now works correctly. 

but: https://www.dropbox.com/s/6j51lybqnpttk85/popper-up.mov?dl=0 Any idea what this could be?

simonihmig:
> Any idea what this could be?

@basz you are refering to this strange "flicker" effect, right?

basz:
indeed

simonihmig:
@basz should be fixed with the latest commits!

basz:
yes, beautiful  thank you!

adambedford:
+1 for automatic error display based on `ember-data` model errors. Is this something you'd consider, @simonihmig ?

sdhull:
@adambedford thanks for asking that, I was planning on submitting a PR for it, but won't bother if it's not going to be considered...

simonihmig:
> +1 for automatic error display based on ember-data model errors. Is this something you'd consider

@sdhull @adambedford Sure, that would be nice. I don't know however what makes most sense, to integrate that into `ember-bootstrap` itself, or provide the "glue" in a "meta" addon, like the other existing addons that provide validation integration (`ember-bootstrap-cp-validations`, `ember-bootstrap-changeset-validations`)?

As it is a bit off-topic here, please create a separate issue for this, where we can further discuss, thanks!

gijsbotje:
heads up, beta-2 has been released
https://github.com/twbs/bootstrap/releases/tag/v4.0.0-beta.2

simonihmig:
> Not sure if this is being tracked and I just didn't see an item for it, but BS4 significantly changes how form validation errors are displayed. For instance, .has-danger class does nothing now.

@sdhull this should have been fixed with the latest commit! 🎉

package.json
```diff
@@ -21,7 +21,7 @@
   "author": "Simon Ihmig <ihmig@kaliber5.de>",
   "license": "MIT",
   "devDependencies": {
-    "bootstrap": "4.0.0-alpha.6",
+    "bootstrap": "^4.0.0-beta",
```
bgentry: FYI you'll also need to tweak index.js the same way: https://github.com/kaliber5/ember-bootstrap/blob/da0b49cda770cda80c3e0066601bf1fedba638e9/blueprints/ember-bootstrap/index.js

simonihmig: @bgentry thanks for the reminder, fixed this!

simonihmig:
Seems all the important things are done here.

@srvance do you see any blockers left that prevent this from merging?

srvance:
Other than the test failure, no. The remaining checklist items can be follow on. I don't think the dropdown toggle is an API change.

bgentry:
I'm getting errors on this branch when I hover over an item that has a tooltip. I suspect it's related to ember-popper changes?

srvance:
@simonihmig I'm finding my time is more limited than anticipated right now. Unless you see a showstopper, let's track outstanding issues separately and merge this. I'll let you make the final call on the right timing for this, but I think it's ready for broader use.

simonihmig:
@srvance no problem at all! Agree, will merge this soon. There will be some more `-rc.x` releases I guess before hitting 1.0, so we will be able to iterate an BS4 support when problems arise. 

@bgentry can you please elaborate, what errors exactly do you see? (there was an ember-popper issue with IE11 fixed recently, but I guess that's not what you are referring to?)

bgentry:
@simonihmig just tried this again, looks like another ember-popper update fixed the errors (e06ee9a) 👍

simonihmig:
@bgentry Oh, that's great! Thanks for the super quick feedback!!!

srvance:
🎉 ❗️

PR #412, opened 2017-08-15
Test for IE 10 + 9

simonihmig:


simonihmig:
This should be ready to be merged now. Created this (https://github.com/kaliber5/ember-cli-classlist-polyfill) to fix IE9 support.

*Actually I was a bit hesitant to actively support IE9+10, as they are not even supported by MS anymore. But as Ember + Bootstrap do support them, and the only issue was this classList thing, it seemed ok to fix this and declare support for them. However that implies that we must continue to support them for the lifetime of e-b 1.x I guess. So any different opinions on that? cc @srvance*

srvance:
Looking [here](https://www.netmarketshare.com/browser-market-share.aspx?qprid=2&qpcustomd=0&qpcustom=) in the "Browser Version" section, IE 8, 9, and 10 have market share of 1.35%, 1.10%, and .80%, respectively. This is well below the latest versions of Safari and Firefox, several versions of Chrome, IE11, and Edge14.

Your statement about longevity of support gives me pause. I don't think we can do a market share cutoff. I would rather say something like "Latest version of the major browsers at the time of 1.0 release. Some previous versions tested and working and we'll make reasonable efforts to continue to maintain compatibility, but ongoing support not guaranteed." How long do you think it will be until EB2.0? Or do you want to have to rev to 2.0 just to drop browser support?

simonihmig:
Ok, I think it makes sense to try to cover older IEs as long as this seems reasonable and affordable (in terms of dev costs), but to exclude them specifically from semver guarantees as you were suggesting. Would add some notes then, so basically I would support all latest versions of evergreen browsers (Chrome, FF, Safari, Edge) and IE11, so IE9+10 would fall out of guarantees. Sounds ok?

PR #416, opened 2017-08-17
Addl. Minor Updates to BS4-BETA: Mainly comments, but a couple of actual code changes.

joshuairl:
Specifically one of the concerns I had was with `classTypePrefix: 'card'` within `addon/components/bs4/bs-accordion/item.js`.  If this determines the `card-(color)` then those should change to `bg-(color)` for the `item`.

### DIFF
--- a/addon/components/base/bs-navbar.js
+++ b/addon/components/base/bs-navbar.js
@@ -82,7 +82,7 @@ import listenTo from 'ember-bootstrap/utils/listen-to-cp';
 
  Bootstrap 4 changed the default navbar styling option from `navbar-default` to `navbar-light`.
  If you explicitly specified "default" in Bootstrap 3 and are migrating, you will need to change
- this in your code. Bootstrap 4 changes navbar-inverse to navbar-dark.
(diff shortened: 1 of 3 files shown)


addon/components/bs4/bs-accordion/item.js
```diff
@@ -6,8 +6,8 @@ export default AccordionItem.extend({
   /**
    * @property classTypePrefix
    * @type String
-   * @default 'card'
+   * @default 'bg'
```
srvance: This reminds me that we made a decision to move all docs to the `base` version, even for BS4-specific stuff.

addon/components/base/bs-navbar.js
```diff
@@ -258,7 +258,7 @@ export default Component.extend(TypeClass, {
    *
    * @property backgroundColor
    * @type String
-   * @default 'faded'
+   * @default 'light'
```
srvance: We should probably elaborate this to express the BS3/4 differences.

srvance: I'll take care of this.

srvance: I didn't look far enough up in the context.

Issue #418, opened 2017-08-18
Failing layout when `formLayout` property is set to undefined


donaldwasserman:
@simonihmig Per our conversation on slack, here's the bug report for the failing `formLayout` property.

When using the `{{bs-form}}` component, if you provide an undefined `formLayout` property, the form fails to render, without errors or exceptions.

See PR #419

PR #419, opened 2017-08-18
Provides fallback for undefined `formLayout`

donaldwasserman:
@simonihmig  Provided a fix based on your suggestion for #418

### DIFF
--- a/addon/components/bs3/bs-form.js
+++ b/addon/components/bs3/bs-form.js
@@ -3,8 +3,10 @@ import Form from 'ember-bootstrap/components/base/bs-form';
 
 export default Form.extend({
   layoutClass: computed('formLayout', function() {
-    assert('must provide a `formLayout` attribute.', this.get('formLayout'));
     let layout = this.get('formLayout');
+    let supportedTypes = ['vertical', 'horizontal', 'inline'];
+    assert('must provide a valid `formLayout` attribute.',
+      supportedTypes.includes(this.get('formLayout')));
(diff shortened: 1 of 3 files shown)


addon/components/bs3/bs-form.js
```diff
@@ -1,9 +1,12 @@
-import { computed } from '@ember/object';
+import { computed, assert } from '@ember/object';
```
simonihmig: should be `import { assert } from '@ember/debug';`
See https://github.com/ember-cli/ember-rfc176-data

addon/components/bs3/bs-form.js
```diff
@@ -1,9 +1,12 @@
...
 import Form from 'ember-bootstrap/components/base/bs-form';
 
 export default Form.extend({
   layoutClass: computed('formLayout', function() {
     let layout = this.get('formLayout');
+    let supportedTypes = ['vertical', 'horizontal', 'inline'];
+    assert('must provide a valid `formLayout` attribute.',
+      supportedTypes.includes(this.get('formLayout')));
```
simonihmig: I think older browsers don't have `.includes`. You should either wrap the array in a `Ember.A()` or use plain `indexOf(...) !== -1`. Also you can reuse `layout` instead of getting this again.

donaldwasserman:
This fancy module loading is too much for me!

I had to update some of the tests. One of the tests I has some particularly ugly changes, because I don't know how to add dynamic variables into that hbs'' string

simonihmig:
Thanks!

Issue #404, opened 2017-08-08
Accordion class binding on item.title and item.body


Techn1x:
I've just been playing around with accordions and noticed that a supplied class name doesn't seem to get binded to item.body (but it does for item.title)

For example, the following code;
```handlebars
{{#bs-accordion class="list-accordion" as |acc|}}
	{{#acc.item class="list-accordion-item" type=(if (lt i 5) 'danger' 'warning') as |accItem|}}
		{{#accItem.title class="list-accordion-item-heading"}}
			...
		{{/accItem.title}}
		{{#accItem.body class="list-accordion-item-body"}}
			...
		{{/accItem.body}}
	{{/acc.item}}
{{/bs-accordion}}
```

Results in the something similar to the following;
```html
<div role="tablist" class="list-accordion panel-group ember-view">
	<div id="ember1174" class="list-accordion-item panel panel-danger ember-view">
		<div id="ember1175" role="tab" class="list-accordion-item-heading panel-heading collapsed ember-view">
			...
		</div>
		<div id="ember1191" class="panel-collapse collapse ember-view">
			...
		</div>
	</div>
</div>
```

I tried to pinpoint in the e-b code what was causing it but some of the accordion code goes a little over my head unfortunately

simonihmig:
The item.body is a tagless component (`tagName: ''`), so cannot receive a class automatically. We would have to assign it explicitly [here](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/templates/components/bs3/bs-accordion/body.hbs#L2) (same for bs4). Want to do a PR?

Nevertheless, you should be able to appply your custom styling without having to set new classes everywhere, with a selector like `ist-accordion-item .panel-body`, right?

Techn1x:
Ah ok! I'll see if I can do a PR for it.

Definitely only a minor issue because like you say, I can just use a different selector, I was more just curious about why the class binding worked for one component and not another :)

PR #420, opened 2017-08-19
Accordion item body can now accept custom classes

Techn1x:
As discussed in https://github.com/kaliber5/ember-bootstrap/issues/404


Techn1x:
I think I did it correctly, let me know if there's a better way.
Only downside to this approach is that if no custom class is defined you end up with a space at the end of the class attribute like this
`<div class="panel-body ">`

simonihmig:
Thanks! Will merge once I figured out what's up with Travis...

PR #431, opened 2017-08-27
Update ember-cli-build-config-editor to v5.1

srvance:


srvance:
I released it as a patch version upgrade, but it appears to have a problem in e-b with esprima version conflicts and I changed the exact text of how the properties are written to use identifiers over literal strings as keys where possible.

simonihmig:
@srvance what's the status with this? Should this be updated, or rather closed?

srvance:
Your choice. It will expand the configs the blueprint can handle, but I won’t be able to work on it for awhile.

srvance:
This has since been replaced.

jelhan:
@srvance What do you mean by "has been replaced"? `ember-cli-build-config-editor` is still locked at `0.5.0`.

srvance:
I thought it had been replaced with the interactive configuration tool. We can re-open if we want to keep enhancing it.

Issue #364, opened 2017-06-15
Make bs-form showAllValidations property public?


jelhan:
I've sometimes uses cases where I have to show all validation errors independently of form submission. `showAllValidations` property of `{{bs-form}}` is very helpful therefore. But it's currently private. Is there any reason why it shouldn't be public? Or is there a better way to achieve that?

Moxide:
Hi,
If you set it, though it's private, it seems to work (I'm using it in one of my forms).

srvance:
`private` means you can't rely on the stability of that property, so you'll need some extra testing when you upgrade ember-bootstrap.

simonihmig:
Another downside that the we set `showAllValidations` to true when submitting the form. If you have bound that property to some property of your app's state, it will implicitly change that (after all it's always two-way bindings, unfortunately), thus violating the DDAU paradigm. So should we make this public, this should be implemented in a way to prevent this...

jelhan:
After all it seems to be a valid use case to show validations independently of user interaction but just changing visibility of `showAllValidations` does not seem to be sufficient. So perhaps we should introduce a new public property `showValidations` which controls visibility of validation errors. If we make it boolean and nullable we support three use cases: 

1. `null` (default): Show validation errors after user interactions like form submission, focus out event of input fields etc. So just the same behaviour as currently implemented.
2. `true`: Show all validation errors immediately independently of user interactions.
3. `false`: Do not show validation errors in any case (but prevent form submission if form is invalid).

@simonihmig: Would you accept a pull request implementing that feature?

Please not that this was supported in ember-bootstrap 0.11 by public `showValidation` property of `FormElement`. This one had the DDAU paradigm violation issue simonihmig described in last commit.

simonihmig:
> Would you accept a pull request implementing that feature?

@jelhan Certainly! 

> prevent form submission if form is invalid

Not so sure about this part though. After all what happens after the `onSubmit` action is triggered is up to the user, so he could implement ignoring the submission easily!? Otherwise this behavior would be hardcoded and not adjustable by the user...

jelhan:
> > prevent form submission if form is invalid
> 
> Not so sure about this part though. After all what happens after the onSubmit action is triggered is up to the user, so he could implement ignoring the submission easily!? Otherwise this behavior would be hardcoded and not adjustable by the user...

[...]

PR #436, opened 2017-09-02
showValidations public property to control visibility of validations errors in bs-form

brisc:
Add showValidations public property to control the visibility of validations on errors in bs-form

### DIFF
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -1,6 +1,6 @@
 import Component from '@ember/component';
 import RSVP from 'rsvp';
-import { set, computed } from '@ember/object';
+import { set, computed, observer } from '@ember/object';
 import { assert } from '@ember/debug';
 import { isPresent } from '@ember/utils';
 import layout from 'ember-bootstrap/templates/components/bs-form';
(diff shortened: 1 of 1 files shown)


simonihmig:
Thanks @brisc for working on this. I guess this will implement #364, right?

Some issues though with the current state of this PR:
* some minor ESLint issues (see tests)
* I would appreciate to not have to use observers. I think the logic can be easily implemented with a computed property and a setter function
* the new functionality would need to be covered by tests. In particular the three cases (`null`, `true`, `false`)
* after form submission and invalid validation, [this line](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-form.js#L238) will still make the validations visible, even though `showValidations` might be `false`, right? 

cc @jelhan

jelhan:
@brisc Thanks for starting that one. Didn't found any time for it the last weeks.

I think more refactoring needs to be done to get this working. Not only a form submission will still set `showAllValidations` to true but also form elements will show validations after user interaction even if form `showValidations` property is set to `false`.

Also we should consider the following edge case:
* `showValidation` is false.
* User interacts with form.
* `showValidations` gets set to `true`.

Therefore we need to store information if form element had a user interaction even if we should not show any validation.

simonihmig:
Closing this as it has gone stale. Feel free to open a new PR when the remaining issues have been addressed!

PR #437, opened 2017-09-02
Introduce ember-popper for tooltip positioning

simonihmig:
@srvance initial implementation is done for tooltips. Works so far, but had to work around some issues. Still WIP...

To Do:

- [x] Fix missing 3px offset for top/left tooltips (failing tests)
- [x] Fix repositioning (wrong position class when popper.js uses preventOverflow modifier, add test!)
- [x] Update popovers
- [x] arrow positioning
- [x] ARIA role
- [x] Cleanup, missing YUIDoc comments
- [x] Check BS4


srvance:
Nice! And so much deleted code!

Do you want to merge into the `bs4-beta` branch instead of `master`?

simonihmig:
> Do you want to merge into the bs4-beta branch instead of master?

I thought about merging this to master once ready, so we can ship early another RC with this internal refactoring. Then rebase `bs4-beta`. Hopefully tooltips and popovers should then "just work". For BS4 dropdowns, they would need another update, as they should also use popper.js like original BS. This can then be done in `bs4-beta` so we can merge it when BS4-beta is fully supported. Sounds ok?

srvance:
@simonihmig Looks like you're done from the punch list. Ready for final review?

simonihmig:
There is one failing test for Ember beta and canary, probably due to the `-in-element` Glimmer API used in ember-popper. Will need to investigate/fix this. Otherwise this should be ready to go! cc @srvance

simonihmig:
For the record: the failing tests are related to https://github.com/kybishop/ember-popper/issues/42, so unfortunately not ready yet...

chickeaterbanana:
You said in issue kybishop/ember-popper#42 
> So currently I am still trying to support IE9+ as long as possible (or at least as long Ember and Bootstrap themselves do support IE9).
bootstrap4 has dropped the Support for ie9+. [Bootstrap4](https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/)

simonihmig:
> Bootstrap4 has dropped the Support for ie9+.

Well, this is used also for BS3, which supports IE9, and BS4 supports IE10, which is also affected by the mentioned issue. So unfortunately it is still something that needs to be resolved...

simonihmig:
Rebased. As we have dropped support for IE9/10, tests should go green now...

PR #443, opened 2017-09-13
Bubble the dropdown button toggle click event

smottt:
Hey!

So, currently if I have multiple `bs-dropdown` dropdowns, when I click on two of the toggle dropdown buttons, both dropdowns remain open. What I would expect is for the first dropdown to be closed when a second toggle is clicked.

This PR fixes that. If this can be done in a better way, I'm open to suggestions.

Thanks!


jeloba:
PR #445 should also address this issue.

smottt:
@jeloba Yes, it does! Thanks for a better fix. :)

PR #445, opened 2017-09-13
Close dropdowns during event capture phase

### LINKED ISSUE #255: BS Dropdown should use capture phase in events
BS Dropdown adds events on `document` and use `bubble` phase of events. I don't like two things about this approach. First of all some ember application are not rendered into body but inside custom container (`appRoot` div) and the second thing is the fact that this handler will be never called in case event will be cancelled beforehand which happens quite often I must say.

It should be at least refactored to `capture` phase of event.

https://developer.mozilla.org/en-US/docs/Web/API/Event/eventPhase

jeloba:
Fixes #255 (the event phase part, not the location where the listener is placed)


simonihmig:
@jeloba Nice, thanks!

PR #448, opened 2017-09-14
fixes #441, providing fallback for undefined formLayout prop

### LINKED ISSUE #441: form not working in production build
```
ember new bs-test
ember install ember-bootstrap
// copy paste form http://www.ember-bootstrap.com/api/classes/Components.Form.html
ember build --prod
```
Then serve the files in dist/, only buttons in #bs-form would show up, `form.element` would not be rendered.

Both bootstrap3/4 fail.

```
Npm packages:
ember-bootstrap: ^1.0.0-rc.2 -> 1.0.0-rc.2
ember-cli: ~2.15.1 -> 2.15.1
bootstrap: ^3.3.7 -> 3.3.7
bootstrap-sass: n/a
ember-cli-sass: n/a
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 3
importBootstrapFont: false
importBootstrapCSS: true
```

donaldwasserman:
Fixes #441 

Since BS4 has only two options, I didn't use assert, and just used false as fallback.

My previous PR used the local branch.

### DIFF
--- a/addon/components/bs4/bs-form.js
+++ b/addon/components/bs4/bs-form.js
@@ -1,9 +1,10 @@
-import { computed } from '@ember/object';
+import { computed, get } from '@ember/object';
+
 import Form from 'ember-bootstrap/components/base/bs-form';
 
 export default Form.extend({
   layoutClass: computed('formLayout', function() {
-    let layout = this.get('formLayout');
-    return layout === 'inline' ? 'form-inline' : null;
+    let layout = get(this, 'formLayout');
+    return layout === 'inline' ? 'form-inline' : false;
   }).readOnly()
 });


simonihmig:
@donaldwasserman Thanks for working on this! Can you briefly explain to me what the issue really was, what is different for returning `false` vs `null`? And how did that influence the production build issues mentioned in #441?

donaldwasserman:
Well, I had a problem with undefined/null values with `classNameBindings` previously, so I didn't test it, to be honest. I could probably write a test to actually programmatically check for this. 

I guess it would probably be better to add an `assert` to make the apis the same at least.

simonihmig:
> I guess it would probably be better to add an assert to make the apis the same at least.

Yes, I think actually that adding that assertion into a `didReceiveAttrs` method makes more sense, for bs3 and 4. The rendering error is pretty sure not related to the form's layout class, but to the non-existing layout sub components when `layout` is invalid!

> Do you know why Integration | Component | bs-modal-simple: size property adds size class is failing?

Strange, tests seem to be a mbit flakey. I restarted, now all green!

donaldwasserman:
Sounds good.

As a side note, Do you know why this happens? I couldn't realy figure it out.

PR #451, opened 2017-09-22
allow for empty navbar toggle breakpoint [bs4]

basz:


basz:
I wouldn’t mind, but the local tests didn’t run.

I’m a testing noob (for ember anyway) but i’ll give it a blind shot.

srvance:
Why didn't they run? Doesn't `BOOTSTRAPVERSION=4 ember test` work to run the tests under bs4? I don't have the right computer with me to give you exactly what I use. You can use `--filter=navbar` to reduce the set of tests during development.

basz:
that runs. thanks

i peeked at the travis cone and tried `ember try:one ember-release-bs4`, which fails with broccoli-tree could not be found, or something… nvm

basz:
done

srvance:
Thank you very much, @basz!

PR #463, opened 2017-10-05
lock in bootstrap in package.json to alpha 6

bgentry:
#415 was not sufficient. Without these changes, yarn still installs the v4 beta.


simonihmig:
Seems you are right, thanks for catching that!

> Maybe this regex needs updating too?

Yes! That causes the failing node tests.

bgentry:
Updated the regex too.

simonihmig:
LGTM, thanks!

*Failing test is unrelated, flakey Saucelabs...*

PR #467, opened 2017-10-11
Injecting custom CSS classes to the bs-tab component

lordkada:
Hey,
in order to use a template I purchased (built on top of Bootstrap), I needed to change a little bit the bs-tab/bs-nav components.

Basically I wanted to push a custom class to the bs-nav component.. something like:

`{{#bs-tab class="tab-nav" as |tab|}} ...`
-->
`<ul class="tab-nav ember-view">`

I don't know if what I did is a desirable implementation (maybe you just want to keep your "ember-bootstrap" 100% standard bootstrap). 

Just let me know

### DIFF
--- a/addon/components/base/bs-nav.js
+++ b/addon/components/base/bs-nav.js
@@ -75,9 +75,18 @@ export default Component.extend({
   layout,
 
   tagName: 'ul',
-  classNames: ['nav'],
 
-  classNameBindings: ['typeClass', 'justified:nav-justified'],
+  classNameBindings: 'classes',
+
+  classes: computed('customClass', 'typeClass', function() {
+    let customClasses = this.get('customClass');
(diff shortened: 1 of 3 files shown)


simonihmig:
@lordkada I somehow overlooked this one, sorry about that!

You can always add classes to any components here by setting them in the template (`{{#bs-nav class="custom"}}`). You PR seems to address setting the class of a nav inside of a tab. But, besides the failing test, this seems to be very specific to your Bootstrap template, and should be solvable by some tweaked CSS selectors. If that does not work, you can still override components/templates of this addon in you app, for your specific use cases.

So I would prefer to not incorporate the changes here, and stick to standard Bootstrap as you assumed. Thanks nevertheless for contributing!

lordkada:
Yep, I perfectly understand. Thanks anyway ;-)

PR #473, opened 2017-10-26
remove twitter association

gijsbotje:
according to the brand page on the docs introduced with bootstrap v4-beta2, bootstrap should not be associated with twitter


simonihmig:
Thanks!

PR #481, opened 2017-11-08
warn if Bootstrap 4 version is not at least 4.0.0-beta

simonihmig:
When using any version of ember-bootstrap that includes #410 users have to use at least 4.0.0-beta of the bootstrap npm package, otherwise the CSS will not match the markup our components generate.

cc @srvance

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,7 @@ const rename = stew.rename;
 const BroccoliDebug = require('broccoli-debug');
 const chalk = require('chalk');
 const SilentError = require('silent-error'); // From ember-cli
+const VersionChecker = require('ember-cli-version-checker');
 
 const defaultOptions = {
   importBootstrapTheme: false,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -110,6 +113,15 @@ module.exports = {
   validateDependencies() {
     let bowerDependencies = this.app.project.bowerDependencies();
 
+    if (this.getBootstrapVersion() === 4) {
+      let checker = new VersionChecker(this);
+      let dep = checker.for('bootstrap');
+
+      if (!dep.gte(minimumBS4Version)) {
```
srvance: How well does it handle comparisons between versions with "alpha" and "beta" tags?

simonihmig: Yeah, I was also unsure, so I checked this using the underlying `semver` package. And it works as we need it, for alpha `gte` will return false! (seems to compare alphabetically I assume)

Issue #477, opened 2017-11-06
calling set on destroyed object .showBackdrop


happydude:
I am seeing almost the same message as [#450](https://github.com/kaliber5/ember-bootstrap/issues/450), but it only happens intermittently.
```
Assertion Failed: calling set on destroyed object: <travel@component:bs-modal::ember941>.showBackdrop = false
```
We have a few routes where we check to see if a model is dirty in willTransition. If so we set the controller property to true for bs-modal open. 

If the user clicks yes then in the action tied to onSubmit we rollback the model and flip the open property back to false then we retry the previous transition. 

I have not seen the error if I comment out setting the open property to false. The reason we set the property back to false is so the modal does not reopen if they come back.

---

Output from `ember bootstrap:info` (available since 1.0.0-alpha.9):
```
Npm packages:
ember-bootstrap: ^1.0.0-rc.3 -> 1.0.0-rc.3
ember-cli: ~2.16.2 -> 2.16.2
bootstrap: ^3.3.7 -> 3.3.7
bootstrap-sass: n/a
ember-cli-sass: n/a
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 3
importBootstrapFont: false
importBootstrapCSS: true

```

PR #484, opened 2017-11-11
fix sporadic errors with bs-modal in tests

larsklevan:
If I create and close a modal in a test I get two sporadic errors. Investigated this because our app's CI has been failing sporadically with the `scrollTop` error.

Here is the test I used to reproduce the issues:

```javascript
let i;
for (i = 0; i < 5000; i++) {
  test(`bs-modal remove markup ${i}`, function(assert) {
    assert.expect(0);
    this.set('showModal', true);
    this.render(hbs`{{#bs-modal open=showModal onHide=(action (mut showModal) false) as |modal|}}
        {{#modal.body}}Hello world!{{/modal.body}}
      {{/bs-modal}}`);
    this.set('showModal', false);
  });
}
```

error w/ "scrollTop"
screen shot 2017-11-11 at 1 40 20 pm

error w/ "set on destroyed object"
screen shot 2017-11-11 at 3 05 05 pm

### DIFF
--- a/addon/components/base/bs-modal.js
+++ b/addon/components/base/bs-modal.js
@@ -418,7 +418,9 @@ export default Component.extend(TransitionSupport, {
       this.set('inDom', true);
       schedule('afterRender', () => {
         let modalEl = this.get('modalElement');
-        modalEl.scrollTop = 0;
+        if (!modalEl) {
+          return;
+        }
 
         this.handleUpdate();
         this.set('showModal', true);
@@ -512,6 +514,9 @@ export default Component.extend(TransitionSupport, {
       assert('Backdrop element should be in DOM', backdrop);
 
       let callbackRemove = function() {
+        if (this.get('isDestroyed')) {
+          return;
+        }
         this.set('showBackdrop', false);
         if (callback) {
           callback.call(this);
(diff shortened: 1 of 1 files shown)


addon/components/base/bs-modal.js
```diff
@@ -418,7 +418,9 @@ export default Component.extend(TransitionSupport, {
       this.set('inDom', true);
       schedule('afterRender', () => {
         let modalEl = this.get('modalElement');
-        modalEl.scrollTop = 0;
```
simonihmig: Why did you remove this line? Shouldn't it just be moved after the guard you added?

simonihmig:
Thanks for working on this! Seems to fix #477.

addon/components/base/bs-modal.js
larsklevan: Oops - that was an accident. Updated.

simonihmig:
Thanks again! 👍

bryanhickerson:
@simonihmig Can you cut a new release with this fix? I'm hitting this error pretty regularly.

simonihmig:
Sure, will do at the latest on Monday!

PR #486, opened 2017-11-14
customize ModalSimple submit button type

### LINKED ISSUE #232: Customizable primary button type on ModalSimple?
One feature I'd like is the ability to control which button `type` is used as the submit button in the simple modal, such as `primary`, `danger`, etc., depending on the action being performed. Of course I could just use the full modal and completely control its contents that way, but that's quite a bit messier 😃 

Do you have thoughts on if that's a good idea and if so, what the option should be called? `submitType` seems like the obvious choice.

The other issue is that if I were to contribute this as a PR, I'd want to be able to use it on the BS4 branch (#206) because that's where I'm currently running. Maybe if I PR it to master it can be cherry-picked over to `bs4`?

bgentry:
Took me a long while to actually do this, but I re-encountered the desire for this feature today.

Fixes #232.

### DIFF
--- a/addon/components/base/bs-modal-simple.js
+++ b/addon/components/base/bs-modal-simple.js
@@ -111,6 +111,16 @@ export default Modal.extend({
   closeTitle: 'Ok',
 
   /**
+   * The type of the submit button (primary button).
+   *
+   * @property submitButtonType
+   * @type string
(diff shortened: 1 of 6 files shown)


bgentry:
test failure looks like a flake

PR #493, opened 2017-11-27
fix conflicting usage of the `size` property

### LINKED ISSUE #492: Conflicting size attribute results in IndexSizeError for size="sm" on form.element
```
Npm packages:
ember-bootstrap: ^1.0.0-rc.4 -> 1.0.0-rc.4
ember-cli: ^2.16.2 -> 2.16.2
bootstrap: 4.0.0-beta.2 -> 4.0.0-beta.2
bootstrap-sass: n/a
ember-cli-sass: ^7.0.0 -> 7.0.0
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapCSS: false
importBootstrapTheme: false
importBootstrapFont: false
rattletrap:hf-app-lab bas$ 
```
IndexSizeError (DOM Exception 1): The index is not in the allowed range.

A form-group supports `size="sm/md/lg"` to render form-group-* classes. However the size attribute is also use for the native size attribute of the input element. Strings aren't allowed there. So when I use an integer output renders as follows.

```
{{#bs-form submitOnEnter=true formLayout="inline" as |form|}}
    {{form.element size="50" controlType="input"}}
{{/bs-form}}

```
renders as
```
<form id="ember531" role="form" class="form-inline ember-view">
    <div id="ember550" class="form-control-50 form-group ember-view">
<!----> 
     <input size="50" type="input" id="ember550-field" class="form-control ember-view">
<!---->
<!---->
</div>
</form>
```

simonihmig:
Fixes #492

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -282,7 +282,7 @@ const nonDefaultLayouts = A([
  <td></td>
  </tr>
  <tr>
- <td>size</td>
+ <td>size<br>via <code>controlSize</code> property</td>
  <td></td>
  <td></td>
  <td>✔︎</td>
--- a/addon/templates/components/common/bs-form/element.hbs
+++ b/addon/templates/components/common/bs-form/element.hbs
@@ -36,7 +36,7 @@
       disabled=disabled
       readonly=readonly
       required=required
-      size=size
+      size=controlSize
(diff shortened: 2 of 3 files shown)


basz:
thanks!

PR #508, opened 2017-12-07
Fix accordion items wrongly registering as childs on …

simonihmig:
…the (wrong) parent, e.g. on tab component.

Fixes the weird tab/accordion issue demonstrated in https://github.com/mqje/ember-bootstrap-accordion-example: accordion items were being registered as child to the tab containing them, that's why the tab nav got rerendered whenever a new accordion item was added... 🙄

@mqje this should fix it! Thanks for the awesome reproduction!


mqje:
Who thanks, thqt was quick, Ill check it out monday on the project haha

Issue #498, opened 2017-11-30
BS4 form-control-plaintext


basz:
I don't think the correct BS4 master supports add a readonly plain input type with a .form-control-plaintext. When I do that the .form-control is also added.

Should we add a controlType='input-plain'?

```
<form>
  <div class="form-group row">
    <label for="staticEmail" class="col-sm-2 col-form-label">Email</label>
    <div class="col-sm-10">
      <input type="text" readonly class="form-control-plaintext" id="staticEmail" value="email@example.com">
    </div>
  </div>
</form>
```

basz:
hi @simonihmig what would be a suggested solution for this use case? Willing to pick this up when not too complicated...

basz:
this modification seems to do the trick, but I'm not sure if this approach is best practice

```

  classNameBindings: ['_classNames'],
  _classNames: computed('_type', '_isPlain', function() {
    if (this.get('_isPlain')) {
      return ['form-control-plaintext'];
    }

      return ['form-control'];

  }),

  /**
   * @property _isPlain
   * @type {Boolean}
   * @private
   */
  _isPlain: false,

  /**
   * @property _type
   * @type {String}
   * @private
   */
  _type: 'text',

  /**
   * @property type
   * @type {String}
   * @public
   */
  type: computed({
    get(key) {
      return this.get('_key');
    },
    set(key, value) {
      if (value.split('-').pop() === 'plain') {
        this.set('isPlain', true);
        this.set('_type', value.split('-').shift());
      } else {
        this.set('_type', value);
      }
    }
  }),
```

PR #510, opened 2017-12-14
Form control plaintext

basz:
https://github.com/kaliber5/ember-bootstrap/issues/498

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -36,6 +36,10 @@ const nonDefaultLayouts = A([
  {{/bs-form}}
  ```
 
+ Suffixing the controlType with `-plain` will render an input type with the `form-control-plaintext` added to the
+ input. This will only work on text inputs.
+
+
  ### Form validation
 
  In the following example the control elements of the three form elements value will be bound to the properties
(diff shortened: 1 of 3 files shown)


basz:
first problem: BS3 apparently uses 

```
<p class="form-control-static">email@example.com</p>
```

simonihmig:
@basz thanks for your effort so far! Let's see how to do this...

Indeed BS3 seems to expect completely different markup. That could be a reason to create a completely separate `plaintext` form control, so we create a component at `bs-form/element/control/plaintext.js` and have different implementations for BS3 and BS4 (by having separate components in the respective `bs3`/`bs4` folders).

So the API from the user's PoV would be something like `{{form.element type="plaintext" label=".." ...}}`.

Would that make sense? This would not allow for different input types in the case of BS4 (like `email`), but would that even be useful given they are `readonly` anyways?

basz:
Ok agreed. I’ll go ahead and refactor this a bit.

Read only email tel; etc, makes sense on a implementations level but functionality wise is seems indeed a bit off.

basz:
On second thought; who says readonly or disabled needs to be enabled for a plaintext input of types other then text?

basz:
found that this renders (almost) what I need without EB changes


```
{{#form.element controlType="text" size='sm' readonly=true disabled=true value=(t 'table.showingOutOfTotal' this.showingOutOfTotal) as |el|}}
            {{component el.control classNames="form-control-plaintext bg-white"}}
          {{/form.element}}
```

simonihmig:
@basz closing this as the PR seems stale and out-dated. Should you want to continue working on this, feel free to reopen! Thanks

PR #513, opened 2017-12-20
Issue #466

### LINKED ISSUE #466: Importing variables in v4
I think the way to import variables has changed in v4. Here's what ember-bootstrap's docs say to do:

```scss
@import "ember-bootstrap/bootstrap/variables.scss";
```

However, that gives an import error. I tried a bunch of different combos, and you actually need to drop the `/bootstrap` part of the path. That also doesn't work on its own, though, because that file now uses functions which have to be imported too. Here's what seems to work for me:

```scss
@import "ember-bootstrap/functions.scss";
@import "ember-bootstrap/variables.scss";
```

Admittedly I'm using the bootstrap v4 beta version right now. I realize lots of stuff is broken there but I don't think this is related or at least not likely to change in future betas.

To be clear, the reason I'm trying to import these files separately is that I want to customize variables before bootstrap itself gets imported (otherwise they have no effect) and I also want to use some of bootstrap's variables in my own definitions. Specifically, things like the grayscale and color functions.

---

```
Npm packages:
ember-bootstrap: ^1.0.0-rc.2 -> 1.0.0-rc.2
ember-cli: ~2.15.1 -> 2.15.1
bootstrap: ^4.0.0-alpha.6 -> 4.0.0-beta
bootstrap-sass: n/a
ember-cli-sass: ^7.0.0 -> 7.0.0
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapFont: false
importBootstrapCSS: false
```

c410-f3r:
Fixes #466.

### DIFF
--- a/tests/dummy/app/templates/getting-started/assets.hbs
+++ b/tests/dummy/app/templates/getting-started/assets.hbs
@@ -87,52 +87,52 @@ module.exports = function(defaults) {
 <pre><code class="language-less">// app.less
 
 // Required
-@import "ember-bootstrap/variables.less";
-@import "ember-bootstrap/mixins.less";
+@import "ember-bootstrap/variables";
+@import "ember-bootstrap/mixins";
 
 // Optional - Bootstrapping/Resetting
-@import "ember-bootstrap/bootstrap/normalize.less";
-@import "ember-bootstrap/bootstrap/print.less";
+@import "ember-bootstrap/normalize";
+@import "ember-bootstrap/print";
 
 // Optional - Everything else
-@import "ember-bootstrap/scaffolding.less";
-@import "ember-bootstrap/type.less";
-@import "ember-bootstrap/buttons.less";
+@import "ember-bootstrap/scaffolding";
+@import "ember-bootstrap/type";
+@import "ember-bootstrap/buttons";
 </code></pre>
 
 <p><b>BS3 - Sass</b></p>
 <pre><code class="language-sass">// app.scss
 
 // Required
-@import "ember-bootstrap/bootstrap/variables.scss";
-@import "ember-bootstrap/bootstrap/mixins.scss";
+@import "ember-bootstrap/bootstrap/variables";
+@import "ember-bootstrap/bootstrap/mixins";
 
 // Optional - Bootstrapping/Resetting
-@import "ember-bootstrap/bootstrap/normalize.scss";
-@import "ember-bootstrap/bootstrap/print.scss";
+@import "ember-bootstrap/bootstrap/normalize";
+@import "ember-bootstrap/bootstrap/print";
 
 // Optional - Everything else
-@import "ember-bootstrap/bootstrap/scaffolding.scss";
-@import "ember-bootstrap/bootstrap/type.scss";
-@import "ember-bootstrap/bootstrap/buttons.scss";
+@import "ember-bootstrap/bootstrap/scaffolding";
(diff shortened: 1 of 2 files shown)


simonihmig:
@c410-f3r Thanks again for your efforts here!

I fear this big table looks a bit overwhelming. And it does not deliver that much new information. I think basically what users have to know is that for the official bootstrap packages (3 + 4), you do a `@import "ember-bootstrap/<filename.ext>;`. 

Only for the Sass port of BS3 you have this additional `bootstrap` folder you have to take into account:
```sass
@import "ember-bootstrap/bootstrap.scss"; // the main bootstrap file does not require the folder
@import "ember-bootstrap/bootstrap.scss/forms.scss"; // all specific files are within bootstrap/
```

Also just importing a single file won't probably work, you always have to import `variables` and I think `mixins` too. And there is a special thing for BS4: it requires you to import `functions.scss` before anything else. There is an issue for this already: https://github.com/kaliber5/ember-bootstrap/issues/466

So maybe when you create three blocks (for the three cases you have now in table columns), with some example imports, it is better understandable this way? Each should include the mandatory (`functions` [only BS4], `variables`, `mixins`) and some additional imports. We can mention that they just have to look into their `node_modules/bootstrap` (`node_modules/bootstrap-sass` for BS3 Sass) folder to see what files are available.

Does that make sense?

c410-f3r:
Seems fair to me.

tests/dummy/app/templates/getting-started/assets.hbs
```diff
@@ -79,69 +79,66 @@ module.exports = function(defaults) {
...
+@import "ember-bootstrap/bootstrap/buttons.scss";
+</code></pre>
+
+<p><b>BS4 - Sass only</b></p>
+<pre><code class="language-sass">// app.scss
+
+// Required
+@import "ember-bootstrap/variables.scss";
+@import "ember-bootstrap/mixins.scss";
+@import "ember-bootstrap/functions.scss";
```
simonihmig: IIRC,`functions` has to be the first import, as `mixins` relies on it! At least that's what they do: https://github.com/twbs/bootstrap/blob/v4-dev/scss/bootstrap.scss#L8-L10

tests/dummy/app/templates/getting-started/assets.hbs
```diff
@@ -79,69 +79,66 @@ module.exports = function(defaults) {
...
-</table>
+<p>
+  It is possible to import only the assets your app needs but the definitions varies according to the chosen configuration.
+</p>
+
+<p><b>BS3 - Less</b></p>
+<pre><code class="language-less">// app.less
+
+// Required
+@import "ember-bootstrap/variables.less";
```
simonihmig: I wonder if we should remove the file extension here and everywhere lese, `.less` as well as `.scss`?

I now this works for both. Always did it like this myself. BS4 also: https://github.com/twbs/bootstrap/blob/v4-dev/scss/bootstrap.scss

c410-f3r: Nice!

tests/dummy/app/templates/getting-started/assets.hbs
c410-f3r: Done

simonihmig:
Thanks again!

PR #514, opened 2017-12-20
Allow multiple form error messages

### LINKED ISSUE #253: Multiple error messages in form fields
Is there a reason why we only display the first error on an invalid model attribute? (cf: https://github.com/kaliber5/ember-bootstrap/blob/536036adad90e70a4ac9add59bea881a99eff631/addon/templates/components/common/bs-form/element/errors.hbs#L2)

Wouldn't we want to loop through all items to show them?

c410-f3r:
Fixes #253.

### DIFF
--- a/addon/components/base/bs-form/element/errors.js
+++ b/addon/components/base/bs-form/element/errors.js
@@ -26,9 +26,12 @@ export default Component.extend({
   messages: null,
 
   /**
+   * Whether or not should display several errors at the same time.
+   *
+   * @default false
    * @property showMultipleErrors
-   * @type {Boolean}
    * @public
+   * @type {Boolean}
    */
   showMultipleErrors: false
 });
--- a/addon/components/base/bs-form/element/errors.js
+++ b/addon/components/base/bs-form/element/errors.js
@@ -2,7 +2,6 @@ import Component from '@ember/component';
 import layout from 'ember-bootstrap/templates/components/bs-form/element/errors';
 
 /**
-
  @class FormElementErrors
  @namespace Components
  @extends Ember.Component
@@ -24,5 +23,12 @@ export default Component.extend({
    * @type {Ember.Array}
    * @public
    */
-  messages: null
+  messages: null,
+
+  /**
+   * @property showMultipleErrors
+   * @type {Boolean}
+   * @public
+   */
+  showMultipleErrors: false
 });
--- a/addon/templates/components/common/bs-form/element/errors.hbs
+++ b/addon/templates/components/common/bs-form/element/errors.hbs
@@ -1,3 +1,11 @@
 {{#if show}}
-  <div class={{feedbackClass}}>{{messages.firstObject}}</div>
+  {{#if showMultipleErrors}}
+    <div class="{{feedbackClass}} pre-scrollable">
+      {{#each messages as |message|}}
+        <p>{{message}}</p>
+      {{/each}}
+    </div>
+  {{else}}
+    <div class={{feedbackClass}}>{{messages.firstObject}}</div>
+  {{/if}}
 {{/if}}
--- a/addon/templates/components/common/bs-form/element/errors.hbs
+++ b/addon/templates/components/common/bs-form/element/errors.hbs
@@ -2,7 +2,7 @@
   {{#if showMultipleErrors}}
     <div class="{{feedbackClass}} pre-scrollable">
       {{#each messages as |message|}}
-        <p>{{message}}</p>
+        <div>{{message}}</div>
       {{/each}}
     </div>
   {{else}}
--- a/addon/templates/components/common/bs-form/element/errors.hbs
+++ b/addon/templates/components/common/bs-form/element/errors.hbs
@@ -1,8 +1,8 @@
 {{#if show}}
   {{#if showMultipleErrors}}
-    <div class="{{feedbackClass}} pre-scrollable">
+    <div class="pre-scrollable">
       {{#each messages as |message|}}
-        <div>{{message}}</div>
+        <div class="{{feedbackClass}}">{{message}}</div>
       {{/each}}
     </div>
   {{else}}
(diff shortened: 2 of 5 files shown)


c410-f3r:
Failing test in not related: `not ok 1736 IE 11.0 - Integration | Component | bs-popover: should adjust popover arrow`

addon/templates/components/common/bs-form/element/errors.hbs
```diff
@@ -1,3 +1,11 @@
 {{#if show}}
-  <div class={{feedbackClass}}>{{messages.firstObject}}</div>
+  {{#if showMultipleErrors}}
+    <div class="{{feedbackClass}} pre-scrollable">
```
simonihmig: What is `pre-scrollable` here? This one described here: https://getbootstrap.com/docs/4.0/content/code/#code-blocks?

If so, it should be a class, right? And I am not sure we should take the decision to apply that on behalf of the user. Maybe add some own class like `multiple-errors` or so, so the user can apply its own styling he chooses?

addon/templates/components/common/bs-form/element/errors.hbs
```diff
@@ -1,3 +1,11 @@
...
+      {{#each messages as |message|}}
+        <p>{{message}}</p>
```
simonihmig: What about the `<p>` here, I guess this will add some additional margin to the bottom, i.e. increase the distance from the error message to the following form element? I would think that a single message with `showMultipleErrors=true` should look the same as the first message of many messages with `showMultipleErrors=false`. Because of the slightly different markup here, this would not be the case, right?

Maybe wrap each message in its own `<div class={{feedbackClass}}>`

addon/templates/components/common/bs-form/element/errors.hbs
c410-f3r: I think it is much better to have a `scrollable` container by default, otherwise, the page height could increased by several orders of magnitude.
[`pre-scrollable`](https://github.com/twbs/bootstrap/blob/34cd2038d2b16f305919e1fc36ff45913e3d1dfd/dist/css/bootstrap.css#L564-L567) is very handy and uses `max-height` which helps to mimics the `show me only one error` use case.

addon/templates/components/common/bs-form/element/errors.hbs
c410-f3r: Done

tests/integration/components/bs-form/element/errors-test.js
```diff
@@ -26,4 +26,11 @@ module('Integration | Component | bs form/element/errors', function(hooks) {
     assert.dom(`.${formFeedbackClass()}`).exists({ count: 1 });
     assert.dom(`.${formFeedbackClass()}`).hasText('foo');
   });
+
+  test('shows multiple errors', async function(assert) {
+    this.set('messages', A(['foo', 'bar']));
+    await render(hbs`{{bs-form/element/errors messages=messages show=true showMultipleErrors=true}}`);
+    assert.dom(`.${formFeedbackClass()}`).exists({ count: 1 });
```
simonihmig: This is failing, should be `{ count: 2 }` now! See https://travis-ci.org/kaliber5/ember-bootstrap/jobs/325103870#L1027

tests/integration/components/bs-form/element/errors-test.js
```diff
@@ -26,4 +26,11 @@ module('Integration | Component | bs form/element/errors', function(hooks) {
...
+    assert.dom(`.${formFeedbackClass()}`).hasText('foo bar');
```
simonihmig: Also this will have to be updated. `dom(`.${formFeedbackClass()}`)` will pick the first `div`, maybe make it two assertions with `:first-child` and `:last-child` added?

c410-f3r: Thanks!

c410-f3r:
Failing test is not related.

`Error: [get("http://localhost:7000/-1/tests/index.html?hidepassed")] Error response status: 23, NoSuchWindow - A request to switch to a different window could not be satisfied because the window could not be found. Selenium error: The driver is not focused on a window. You must switch to a window before proceeding. (WARNING: The server did not provide any stacktrace information)`

simonihmig:
> Failing test is not related

Yes, restarted it. Saucelabs tests tend to be quite flaky unfortunately...

addon/components/base/bs-form/element/errors.js
```diff
@@ -24,5 +23,12 @@ export default Component.extend({
    * @type {Ember.Array}
    * @public
    */
-  messages: null
+  messages: null,
+
+  /**
+   * @property showMultipleErrors
+   * @type {Boolean}
+   * @public
```
simonihmig: Oh, sorry, overlooked this: could you please add a short description sentence, what this property does, and add `@default false`, so the automatically generated API docs are complete.

c410-f3r: Done.
Although the variable name is self explanatory and a commentary makes me feels like I am doing something like this:

```
class Clazz {
  /**
   * Constructs a new Clazz object.
   */
  Clazz() {
  }
}
```

simonihmig:
After reviewing this again, I think an essential piece is actually missing here. You added `showMultipleErrors` to the `bs-form/element/errors` component. But this is a private component, users should not use this directly. They will only interact with `bs-form` and its yielded `bs-form/element`. 

So this property must also live at least on the `bs-form/element` component, and the value be passed to `bs-form/element/errors` here: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/templates/components/common/bs-form/element.hbs#L5-L8

c410-f3r:
D'oh! More of the `Know the whole to better understand the specific` use case.

simonihmig:
Merged this, thanks again!

PR #417, opened 2017-08-18
Add bootstrap carousel

c410-f3r:
I guess this is it.

PR #522, opened 2018-01-04
Add carousel

c410-f3r:
Followed by #375 and #417.
As you said @simonihmig, I would love your help as this will probably speed up things.


simonihmig:
@c410-f3r I rebased this and migrated to the newer testing APIs. Had to change the tests a little bit, as the new async/await based tests and the new test-helpers were working a bit differently, especially when concurrency-tasks (autoPlay) were involved. 

Also an issue with fixed `autoPlay`: previously it only worked when it was initially set, but did not react to changes.

You can have a look at the commit I added, to see what has changed. Still want to investigate this mouse events issue, hopefully over the weekend...

Now let's see what Travis thinks of this! ;)

c410-f3r:
Awesome, thanks for your kindly help, @simonihmig.

I am against changing the `autoPlay` (data-ride) behaviour because as stated in the  [bootstrap documentation](https://getbootstrap.com/docs/4.0/components/carousel/#via-data-attributes), `data-ride="carousel` (autoPlay=true) has a specific responsibility and is used only at page load, therefore, it shouldn't be dynamic. If the user wants to manipulate the sliding process, he should change the `interval` attribute.

Good news! I found out that `mouseEnter` and `mouseLeave` doesn't work because of `ember-native-dom-event-dispatcher`. Is it possible to merge this PR "as is" and solve this event matter in a later time?

simonihmig:
> I am against changing the autoPlay (data-ride) behaviour

Ok, I started to play with this because tests with `autoPlay=true` were failing, because the new `render()` would wait for all run loops to have finished, which is never the case with autoPlay as the concurrency task will run forever... will se how I can revert the `autoPlay` behavior and make tests still pass...

> Is it possible to merge this PR "as is" and solve this event matter in a later time?

I'm ok with that. Will see if I can find out what exactly the problem with `ember-native-dom-event-dispatcher` is...

addon/components/base/bs-carousel.js
```diff
@@ -0,0 +1,580 @@
...
+  /**
+   * Should have hard stop on corners, i.e., first slide won't make a transition to the
+   * last slide and vice versa.
+   *
+   * @default true
+   * @property continuouslyCycle
+   * @public
+   * @type boolean
+   */
+  continuouslyCycle: true,
```
simonihmig: Is there actually a reason this is not named `wrap` as in the original Bootstrap?

c410-f3r: Many properties aren't named as the originals because when I came across these names for the first time it was **really** hard to figure out what they were actually doing. That's why I've renamed them to something more self-explanatory.

I don't mind at all about changing these names to something else, here are the nomenclature tables defined in the documentation:

 | Description | Original | Component |
 | ------ | ------ | ------ |
 | Autoplays after first user event or on page load. | ride='carousel'\|false | autoPlay=false\|true |
 | Disable automatic cycle. | interval=false | interval=0 |
 | If first slide should follow last slide on "previous" event, the  opposite will also be true for "next" event. | wrap=false\|true | continuouslyCicle=false\|true |
 | Jumps into specific slide index | data-slide-to=n | index=n |
 | Keyboard events. | keyboard=false\|true | keyboard=false\|true |
 | Left-to-right or right-to-left sliding. | N/A |  ltr=false\|true |
 | Pause current cycle on mouse enter. | pause='hover'\|null | pauseOnMouseEnter=false\|true |
 | Show or hide controls  | Tag manipulation. | showControls=false\|true |
 | Show or hide indicators  | Tag manipulation. | showIndicators=false\|true |
 | Waiting time of slides in a automatic cycle. | interval=n | interval=n |

| Operation | Description |
| ------ | ------ |
| Transition | Swaps two slides. |
| Interval | Waiting time after a transition. |
| Presentation | Represents a single transition, or a single interval, or the union of both. |
| Cycle | Presents all slides until it reaches first or last slide. |
| Continuously cycle | Infinite cycling without stops. |

c410-f3r:
@simonihmig, this run loop issue is indeed not trivial.
It is possible to break the infinite run loop by simulating the case when the user hovers the mouse on the carousel by setting `isMouseHovering=true`, but it is an ugly KOP (Kludge Oriented Programming) workaround.
Despite all this, I am still inclined to avoid using `shouldRunAutomatically: and('autoPlay', 'hasInterval')` because it doesn't match the original boostrap carousel behavior.

simonihmig:
@c410-f3r add a few more commits:
* reverted the `autoPlay` as you suggested. You are right, we should match the original API here. Got tests working, by setting `interval=0` to stop the infinite run loops...
* speed up tests by using shorter transition duration
* I renamed `continuouslyCycle` to the original `wrap`: I agree with you that we don't have to match the API 1:1, like using a boolean flag instead of setting something to a certain string value (like "carousel"). But in this special case, I think "wrap" still makes sense (at least to me), and I feel it is better to match the naming here. Or do you have strong feelings against it?

c410-f3r:
Wonderful! Coudn't look better, @simonihmig. 
Thank you for these 6 months of collaborations.

simonihmig:
> Thank you for these 6 months of collaborations.

Oh my, has it **really** been that long already? 😨
Sorry for that actually, I at times was delaying this because of late feedback... but this was quite a big PR. So thanks again a lot!

Will merge later, and release as 1.1 when I fixed some other little issues. Do you have a twitter account I can mention when announcing this? You certainly deserve the credit for this! ;)

c410-f3r:
I humbly appreciate your kindness but there is no need for that. Besides, I am the one who must thank you and @srvance for all the hard work you guys put into this awesome project

PR #534, opened 2018-01-19
fix(Modal): submitting form in nested bs-modal-simple throws

### LINKED ISSUE #533: Submitting form in nested bs-modal-simple throws
Having a template like that one:

```
{{#bs-modal-simple
  onSubmit=(action (mut outerFormSubmitted) true)
  submitTitle='submit outer form'
}}
  {{#bs-modal-simple
    onSubmit=(action (mut innerFormSubmitted) true)
    submitTitle='submit inner form'
  }}
  {{/bs-modal-simple}}
{{/bs-modal-simple}}
```

If child modal is submitted an `Uncaught RangeError: Maximum call stack size exceeded.` error is thrown. It seems like [`submit` action of `bs-modal`](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-modal.js#L391-L402) and [`submit` event of `bs-modal/footer`](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-modal/footer.js#L78-L82) are called in an infinite loop.

Here is an ember-twiddle demonstrating the issue: https://ember-twiddle.com/d358d7b3b591941dd64ffa69e6699676?openFiles=templates.application.hbs%2C Error is thrown after clicking "submit inner form" button.

```
Npm packages:
ember-bootstrap: ^1.0.0-alpha.1 -> 1.0.0
ember-cli: ~2.16.2 -> 2.16.2
bootstrap: ^3.3.7 -> 3.3.7
bootstrap-sass: n/a
ember-cli-sass: n/a
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 3
importBootstrapFont: true
importBootstrapCSS: true
```

jelhan:
`querySelectorAll()` is not scoped. `:scope` pseudo-class is not supported in all target browsers so far. So using `elementId` as a work-a-round.

Some background information on scope and `querySelectorAll()` could be found on MDN:
https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll

Fixes #533

### DIFF
--- a/addon/components/base/bs-modal.js
+++ b/addon/components/base/bs-modal.js
@@ -389,7 +389,9 @@ export default Component.extend(TransitionSupport, {
       }
     },
     submit() {
-      let forms = this.get('modalElement').querySelectorAll('.modal-body form');
+      // replace modalId by :scope selector if supported by all target browsers
+      let modalId = this.get('modalId');
+      let forms = this.get('modalElement').querySelectorAll(`#${modalId} .modal-body form`);
       if (forms.length > 0) {
         // trigger submit event on body forms
         let event = document.createEvent('Events');


simonihmig:
Oh and btw, the twiddle seems to render in-place, as the wormhole container div seems to not be added there somehow. If they are properly "wormholed" (i.e. they are rendered as children of the container, thus not nested, in terms of DOM structure), I think this problem would not occur, right?

jelhan:
> Oh and btw, the twiddle seems to render in-place, as the wormhole container div seems to not be added there somehow. If they are properly "wormholed" (i.e. they are rendered as children of the container, thus not nested, in terms of DOM structure), I think this problem would not occur, right?

Yes. This one be limited to modals rendered in-place. I faced this issue in tests only. It took me some time to figure out why the tests are failing. After this small fix it seems to work just fine.

PR #537, opened 2018-01-19
toggle button does not inherit from `bs-button` anymore,…

### LINKED ISSUE #531: NavbarToggle delete button class btn btn-secondary
When use {{navbar.toggle}}
Returns:
```
<button type="button" id="ember692" class="btn navbar-toggler btn-secondary collapsed ember-view">  
<span class="navbar-toggler-icon"></span>
</button>
```

simonihmig:
… removing the extraneous `.btn`, `.btn-*` classes

Fixed #531

### DIFF
--- a/addon/components/base/bs-navbar/toggle.js
+++ b/addon/components/base/bs-navbar/toggle.js
@@ -1,4 +1,4 @@
-import BsButtonComponent from 'ember-bootstrap/components/bs-button';
+import Component from '@ember/component';
 import layout from 'ember-bootstrap/templates/components/bs-navbar/toggle';
 
 /**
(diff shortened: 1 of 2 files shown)


simonihmig:
`navbar.toggle` was extending from `bs-button`, thus inheriting `.btn` and `.btn-{default|secondary}` classes (which it should not have). Only way to prevent this (due to the way the magic `mergedProperties` in Ember work) is AFAIK to not inherit from it. Which makes sense anyway IMHO, as the toggle does not need any of the more advanced features of `bs-button` (e.g. `type`, `icon` etc.), it should just be a plain `<button>`.

It was previously documented as inheriting from `bs-button`, thus changing this *could* be regarded a breaking change. But as this never was intended to be used as a generic button, and I hope nobody was doing this, I *think* we can regard this as a bug fix (so no need to release as `2.0.0` 🙄)

@srvance as you were working on this, I hope you agree?

PR #539, opened 2018-01-23
WIP: Yield validate and submit actions as part of the form hash

### LINKED ISSUE #476: Yield submit/validate actions from bs-form

acorncom:
Fixes #476 

@simonihmig Here's the promised PR. This is still a WIP but thought I'd put it up to get your input on a few things (and make sure I'm taking this the direction you want). This has been tested against the client app I'm working on (and it greatly simplifies the use case I'm dealing with).

### DIFF
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -44,6 +44,7 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
     {{form.element controlType="email" label="Email" placeholder="Email" property="email"}}
     {{form.element controlType="password" label="Password" placeholder="Password" property="password"}}
     {{form.element controlType="checkbox" label="Remember me" property="rememberMe"}}
+    {{bs-button defaultText="Validate" type="info" onClick=(action form.validate)}}
     {{bs-button defaultText="Submit" type="primary" buttonType="submit"}}
   {{/bs-form}}
   ```
@@ -242,6 +243,24 @@ export default Component.extend({
     }
   },
 
+  _validate() {
+    let model = this.get('model');
+
+    let validationPromise = this.validate(model);
+    if (validationPromise && validationPromise instanceof RSVP.Promise) {
+      let validatedPromise = new RSVP.Promise((resolve, reject) => {
+        validationPromise
+          .then(() => { resolve() })
+          .catch(() => {
+            this.set('showAllValidations', true);
+            return reject();
+          });
+      });
+
+      return validatedPromise;
+    }
+  },
+
   keyPress(e) {
     let code = e.keyCode || e.which;
     if (code === 13 && this.get('submitOnEnter')) {
@@ -261,6 +280,38 @@ export default Component.extend({
         isPresent(model) && isPresent(property)
       );
       set(model, property, value);
+    },
+
+    /**
+     * Submit action yielded in the bs-form template and accessible as `form.submit`
+     *
+     * @method actions.submit
+     * @param {Object} [model] Optional, falls back to model passed to bs-form
+     * @return {Promise} (not yet implemented)
+     * @public
+     */
+    submit(model=null) {
+      if(model) {
+        this.set('model', model);
+      }
+      return this.submit();
+    },
+
+    /**
+     * Validate action yielded in the bs-form template and accessible as `form.validate`
+     * This action returns a promise that will either resolve if the model is valid
+     * or reject if it's not.
+     *
+     * @method actions.validate
+     * @param {Object} [model] Optional, falls back to model passed to bs-form
+     * @return {Promise}
+     * @public
+     */
+    validate(model=null) {
+      if(model) {
+        this.set('model', model);
+      }
+      return this._validate();
     }
   }
 });
--- a/tests/integration/components/bs-form-test.js
+++ b/tests/integration/components/bs-form-test.js
@@ -97,6 +97,17 @@ module('Integration | Component | bs-form', function(hooks) {
     assert.ok(submit.called, 'onSubmit action has been called');
   });
 
+  test('Clicking a normal button that triggers the yielded submit action submits the form', async function(assert) {
+    let submit = this.spy();
+    this.actions.submit = submit;
+    await render(
+      hbs`{{#bs-form onSubmit=(action "submit") as |form|}}{{#bs-button onClick=form.submit}}Submit{{/bs-button}}{{/bs-form}}`
+    );
+
+    await click('button');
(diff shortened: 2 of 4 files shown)


addon/components/base/bs-form.js
```diff
@@ -261,6 +280,38 @@ export default Component.extend({
...
+     * @method actions.submit
+     * @param {Object} [model] Optional, falls back to model passed to bs-form
+     * @return {Promise} (not yet implemented)
+     * @public
+     */
+    submit(model=null) {
+      if(model) {
+        this.set('model', model);
+      }
+      return this.submit();
```
acorncom: Do we want to return a promise here that callers call use? Doing so could allow for some nice UI approaches (or wrapping these calls in ember-concurrency tasks)

addon/components/base/bs-form.js
```diff
@@ -242,6 +243,24 @@ export default Component.extend({
...
+    let model = this.get('model');
+
+    let validationPromise = this.validate(model);
+    if (validationPromise && validationPromise instanceof RSVP.Promise) {
+      let validatedPromise = new RSVP.Promise((resolve, reject) => {
+        validationPromise
+          .then(() => { resolve() })
+          .catch(() => {
+            this.set('showAllValidations', true);
+            return reject();
```
acorncom: I didn't add the `onInvalid` call here that exists in `submit`, which might not match what we're after. It seemed like the caller might want more control than that, but happy to put that in if you'd like

tests/integration/components/bs-form-test.js
```diff
@@ -97,6 +97,17 @@ module('Integration | Component | bs-form', function(hooks) {
...
+    let submit = this.spy();
+    this.actions.submit = submit;
+    await render(
+      hbs`{{#bs-form onSubmit=(action "submit") as |form|}}{{#bs-button onClick=form.submit}}Submit{{/bs-button}}{{/bs-form}}`
+    );
+
+    await click('button');
+    assert.ok(submit.called, 'onSubmit action has been called');
+  });
+
```
acorncom: Just realized I didn't add a test of the `form.validate` here, will put that in as part of the next pass

tests/integration/components/bs-form-test.js
```diff
@@ -171,6 +182,35 @@ module('Integration | Component | bs-form', function(hooks) {
...
+    setTimeout(() => {
+      assert.dom(formFeedbackElement()).hasClass(
+        validationErrorClass(),
+        'validation errors are shown after form validation'
+      );
+      assert.dom(`.${formFeedbackClass()}`).hasText('There is an error');
+      done();
+    }, 1);
+
+  });
```
acorncom: Do we want additional tests here?

tests/unit/components/common/bs-form-test.js
```diff
@@ -0,0 +1,59 @@
...
+module('Unit | Component | bs-form', function(hooks) {
+  setupTest(hooks);
+
+  test('calling form.validate returns success if validations pass', function(assert) {
+    assert.expect(1);
+
+    let Factory = this.owner.factoryFor('component:bs-form');
+    let subject = Factory.create({ hasValidator: true, model: {}, validate: function() { return resolve(); } });
+
+    let validatedResult = subject.actions.validate.apply(subject);
```
acorncom: Doing this got a bit ugly, but I wanted to test the promises we were returning to ensure I did them right 😉

addon/components/base/bs-form.js
```diff
@@ -261,6 +280,38 @@ export default Component.extend({
...
+     * Submit action yielded in the bs-form template and accessible as `form.submit`
+     *
+     * @method actions.submit
+     * @param {Object} [model] Optional, falls back to model passed to bs-form
+     * @return {Promise} (not yet implemented)
+     * @public
+     */
+    submit(model=null) {
+      if(model) {
+        this.set('model', model);
```
simonihmig: Hm, I am nor sure I like this one. Setting `this.model` would be kind of a violation of the DDAU principle, as it has been set by the user, so we would make this a two way binding if we set it here, right? 

Is there a specific use case you want to support by having the model be passed as an argument here?

addon/components/base/bs-form.js
```diff
@@ -261,6 +280,38 @@ export default Component.extend({
...
+     * or reject if it's not.
+     *
+     * @method actions.validate
+     * @param {Object} [model] Optional, falls back to model passed to bs-form
+     * @return {Promise}
+     * @public
+     */
+    validate(model=null) {
+      if(model) {
+        this.set('model', model);
```
simonihmig: Same concerns as in `submit`

tests/unit/components/common/bs-form-test.js
simonihmig: LGTM! :) Maybe just tweak the name of the test, to express that it returns a resolving promise?

tests/unit/components/common/bs-form-test.js
```diff
@@ -0,0 +1,59 @@
...
+import { module } from 'qunit';
+
+module('Unit | Component | bs-form', function(hooks) {
+  setupTest(hooks);
+
+  test('calling form.validate returns success if validations pass', function(assert) {
+    assert.expect(1);
+
+    let Factory = this.owner.factoryFor('component:bs-form');
+    let subject = Factory.create({ hasValidator: true, model: {}, validate: function() { return resolve(); } });
```
simonihmig: No need actually to wrap it in another function I think... `validate: resolve` would be enough. But I have been doing this in other places as well, so no big deal 😉

addon/components/base/bs-form.js
```diff
@@ -44,6 +44,7 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
     {{form.element controlType="email" label="Email" placeholder="Email" property="email"}}
     {{form.element controlType="password" label="Password" placeholder="Password" property="password"}}
     {{form.element controlType="checkbox" label="Remember me" property="rememberMe"}}
+    {{bs-button defaultText="Validate" type="info" onClick=(action form.validate)}}
```
simonihmig: I think it would be nice to have this new feature mentioned a bit more explicitly, maybe the yielded `form.submit` action under the `Submitting the form` chapter, and `validate` under `Form validation`?

simonihmig:
@acorncom would you be able to pick up work on this? Would love to land this!
But no worries if you don't have the time ATM, just asking! ;)

svparijs:
Hey guys, this change would help my use case a lot. Any chance this will be merged soonish?

simonihmig:
Closed in favor of #789.

PR #562, opened 2018-03-08
Ember 3.0 Update

simonihmig:


simonihmig:
Failed tests should be fixed by #568

PR #569, opened 2018-03-24
Interactive installer

### LINKED ISSUE #258: Add interactive mode to default blueprint

kolybasov:
Still in progress. Just reminder to finish it and resolve #258 😃


kolybasov:
@simonihmig I am working on it. Need to fix tests and we are ready to go. I will ping you then.

kolybasov:
@simonihmig with the new installer users can specify components to whitelist. Should I add all whitelisted components to whitelist array or bounce between whitelist/blacklist in case enabled/disabled components count?

Example: 
User unchecked only `bs-modal` component. Options:

1. `blacklist: ['bs-modal']`
or
2. `whitelist: ['bs-tooltip', 'bs-alert', …]`

kolybasov:
@simonihmig one more issue I have is with config defaults. I think if user answer is the same as default value then we do not need to write that option to user build file because we do not need it to be there. What you think?

simonihmig:
> Should I add all whitelisted components to whitelist array or bounce between whitelist/blacklist in case enabled/disabled components count?

I think only using `whitelist` is reasonable. This way if the user explicitly selected whitelisted components, and a new component comes up (say `bs-scrollspy`), it will not get included automatically. Which would be the case in the `blacklist` scenario. 

But then I think the user needs an option to basically say "include them all". And that should *not* enumerate all components known at that time in `whitelist`, but rather write `blacklist: []` or better just skip writing `whitelist`, so users who are not concerned about their bundle size will always have all components available, even if new one get added over time...

> I think if user answer is the same as default value then we do not need to write that option to user build file because we do not need it to be there. 

Yes, but I am fine with explicitly writing everything to the config. This way the user knows what options are available just by looking at the config file. And I would guess it makes the implementation simpler, right?

kolybasov:
@simonihmig
> I think only using whitelist is reasonable.

Makes sense. Thanks!

> Yes, but I am fine with explicitly writing everything to the config. This way the user knows what options are available just by looking at the config file. And I would guess it makes the implementation simpler, right?

Definitely simpler 👍

kolybasov:
@simonihmig tests are working now. Remaining:

- [ ] Merge `whitelist` and `blacklist` options when reading user configuration.
- [ ] Add test scenarios for `whitelist` option.

It would be nice if you can review PR and let me know if something need to be changed!

simonihmig:
@kolybasov sorry for the delay here, wasn't able to find the time yet! Will hopefully review this over the weekend!

blueprints/ember-bootstrap/index.js
```diff
@@ -1,241 +1,333 @@
-/* eslint-env node */
-'use strict';
```
simonihmig: 'use strict'  should be preserved for node 4

blueprints/ember-bootstrap/index.js
```diff
@@ -1,241 +1,333 @@
...
+
+    return rsvp.all(promises);
   },
 
-  adjustBootstrapDependencies() {
-    let bootstrapVersion = this.bootstrapVersion;
-    let preprocessor = this.preprocessor;
+  _adjustBootstrapDependencies() {
+    let bs3Version = '^3.3.7';
+    let bs4Version = '^4.0.0';
```
simonihmig: Can you please put them back as `const` at the module root, so it is easier to find them when we need to adjust!

blueprints/ember-bootstrap/wizard.js
```diff
@@ -0,0 +1,162 @@
...
+  let componentsPath = path.resolve(
+    __dirname,
+    '..',
+    '..',
+    'app',
+    'components',
+    '*.js'
+  );
+  let files = fg.sync(componentsPath);
+  return files.map(file => path.basename(file, '.js'));
```
simonihmig: There is just one special case with `bs-contextualhelp` here, which `bs-popover` and `bs-tooltip` extend from, but which itself is not a public component users should see. So it should also not showup in this list, which I guess (haven't tried) is the case now? 

Maybe we have to mark it as private, e.g. by renaming to `-bs-contextulahelp` and filtering out all `-` prefixed components here?

blueprints/ember-bootstrap/wizard.js
```diff
@@ -0,0 +1,162 @@
+const chalk = require('chalk');
```
simonihmig: This should also have `'use strict';`!

Some day I changed the Travis setup to node 6 because of some failing upstream dependencies, but I should probably try again to revert to node 4. Though it's reaching end-of-life, cutting support for it would probably require a major version bump, which I din't really want to do yet...

blueprints/ember-bootstrap/wizard.js
```diff
@@ -0,0 +1,162 @@
...
+}
+
+function questions(options, config) {
+  let mergedConfig = Object.assign({}, config, options);
+  let guards = buildGuards(mergedConfig);
+
+  return [
+    {
+      name: 'customInstall',
+      message: 'Do you want to customize installation of ember-bootstrap?',
```
simonihmig: **your** installation

blueprints/ember-bootstrap/wizard.js
```diff
@@ -0,0 +1,162 @@
...
+      type: 'confirm',
+      default: false
+    },
+    {
+      name: 'bootstrapVersion',
+      message: 'Choose desired Bootstrap version:',
```
simonihmig: **your** desired

blueprints/ember-bootstrap/index.js
```diff
@@ -1,241 +1,333 @@
...
+        // use empty object in case there are no way to read the current config
+        .catch(() => ({}))
+        .then(config => {
+          config.bootstrapVersion = this._detectBootstrapVersion(
+            options,
+            config
+          );
+          config.preprocessor = this._detectPreprocessor(options);
+          this.addonConfig = config;
+          return this.ui.prompt(wizard.questions(options, config));
```
simonihmig: We talked about the issue that FastBoot tests run the default blueprint, though there is currently no way to answer any prompts programmatically. It seems the prompts show up without any escape hatch here (like an env var to suppress them). How is this working now? (i.e. why where the tests even passing :))

kolybasov:
Hey @simonihmig ! Thanks for the review. 
You asked how tests are passing, answer is [stub](https://github.com/kaliber5/ember-bootstrap/pull/569/files#diff-ce3ce94083cff8e3285bc2d7bbc82af9R320) :)
I stole the idea from [ember-cli-eslint](https://github.com/ember-cli/ember-cli-eslint/blob/master/node-tests/blueprints/ember-cli-eslint-test.js#L51). But tests need a little bit more work.

Hope will find any time soon to fix your comments and finish PR.

simonihmig:
> You asked how tests are passing, answer is stub :)

Yeah, I saw that. But that's the `node-tests`, I was wondering about the `fastboot-tests`, as they invoke `ember generate ember-bootstrap` and I was expecting this to fail when no interactive input is made...

blueprints/ember-bootstrap/wizard.js
simonihmig: Never mind, as I was playing with, found that this is not an issue at all, since `bs-contextualhelp` is not epxorted in the `app` tree... 🤦‍♂️

blueprints/ember-bootstrap/wizard.js
```diff
@@ -0,0 +1,162 @@
...
+        "Do you want to automatically insert a div into your index.html, where Bootstrap's modals and tooltips will be rendered to.\n" +
+        chalk.red(
+          'If you set this to false, you will have to add an <div id="ember-bootstrap-wormhole"></div> element by yourself.'
+        ),
+      type: 'confirm',
+      default: getDefault(config, 'insertEmberWormholeElementToDom', true),
+      when: guards(GUARD.customInstall)
+    },
+    {
+      name: 'whitelist',
```
simonihmig: The whitelist question is a bit problematic, when you have all components selected, as this will still write an array with all components to the config. Which is ok for the moment, but when new components get added, those will not be available to the user (as there is a whitelist without that new component), certainly causing a good amount of confusion. I *think* we briefly talked about this, was it on Slack?

I would prefer to have a way to skip writing `whitelist` (or writing `whitelist: null`) to capture the user's intend to not "tree shake" at all. What about a question first asking whether the user wants to tree-shake the available components? And only present the choices if he answered 'yes'?

simonihmig:
@kolybasov don't want to put you under pressure, but any idea when you could address the remarks in the review? Would love to get this merged asap! :)

kolybasov:
@simonihmig sorry, busy right now. I will try to find time at the end of next week. I would like to have it merged too :)

simonihmig:
@kolybasov sorry to bother you again, but any chance of getting back to this? I am thinking of doing a 2.0 release soon (because of some other breaking changes like dropping node 4 support), and also introducing this feature would be a nice fit for the new release!

kolybasov:
@simonihmig really sorry for delaying this but do not have time at all for now. I am keeping this marked as unread in GitHub notifications and will back to this as soon as possible. Thanks for your patience!

Any help with this is much appreciated!

simonihmig:
@kolybasov No problem. I know the difficulties of managing OSS and "the rest of your life" just too well! ;)

Btw, the soon to be released 2.0 version makes BS4 the default, so this introduced some new merge conflicts unfortunately...

simonihmig:
@kolybasov I am closing this as the PR is stale and out-dated. But feel free to pick this up again anytime of course if you want!

PR #570, opened 2018-03-25
Render dropdown menu in a wormhole

### LINKED ISSUE #561: Popper having trouble with parent div overflow
```
ember-bootstrap: 1.2.1 -> 1.2.1
ember-cli: ^2.18.2 -> 2.18.2
bootstrap: 4.0.0 -> 4.0.0
bootstrap-sass: n/a
ember-cli-sass: 7.1.7 -> 7.1.7
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapCSS: false
importBootstrapFont: false
```
**EDIT:** See end of post.

Hello,
Reporting quite a worrying bug here: if dropdown is contained inside of a div which is shorter than the container of the dropdown, the overflow is simply not shown.

calendar
Example attached. The calendar should be fully visible inside of the dropdown (and it is in case the container div is taller), but since the container div is smaller than it should, the dropdown div is cut out.
I suspect it's due to using popper instead of ember-wormhole.
Thank you,

Daniele

**Edit:**
Apparently, this bug is due to liquid-fire. I was using a `liquid-outlet` to wrap the template where I was using the dropdown, and liquid-fire probably makes use of a overflow: hidden for its features, therefore breaking ember-popper. I'll just disable liquid-fire for now.

simonihmig:
By setting `renderInPlace=false` to prevent clipping issues

Closes #561

### DIFF
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -94,6 +94,10 @@ import layout from 'ember-bootstrap/templates/components/bs-dropdown';
  wish to have a dropdown menu item refer to an external link, be sure to apply the `dropdown-item`
  class to the `<a>` tag for Bootstrap 4 compatibility.
 
+ In Bootstrap 4 the dropdown menu will be positioned using the `popper.js` library, just as the original Bootstrap
+ version does. This also allows you to set `renderInPlace=false` on the menu component to render it in a wormhole,
+ which you might want to do if you experience clipping issues by an outer `overflow: hidden` element.
+
  @class Dropdown
  @namespace Components
  @extends Ember.Component
--- a/addon/components/base/bs-dropdown/menu.js
+++ b/addon/components/base/bs-dropdown/menu.js
@@ -57,6 +57,17 @@ export default Component.extend({
    */
   inNav: false,
 
+  /**
+   * Applies only to BS4: by default the menu is rendered in the same place the dropdown. If you experience clipping
+   * issues, you can set this to false to render the menu in a wormhole at the top of the DOM.
+   *
+   * @property renderInPlace
+   * @type boolean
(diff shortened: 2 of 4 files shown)


Techn1x:
@simonihmig Would this be possible to implement for BS3 as well? I have a container with overflow: hidden, and a dropdown inside it that gets clipped by the container. Or has the world moved on from BS3 to BS4 now

simonihmig:
@Techn1x would probably be possible. BS3 does not do this, so the default should still be to match BS3, but an opt-in to use the same ember-popper based positioning and "wormholing" should be possible. I won't have the time for this unfortunately, but if someone wants to tackle this, I would be happy!

bgentry:
I have discovered some menus in my app which appear to be broken by this change. Here is a menu in the top right corner of my page:

| No options specified | with `renderInPlace=false` |
| - | - |
| screen shot 2018-07-23 at 8 48 17 pm |screen shot 2018-07-23 at 8 46 24 pm |

Is that to be expected? Is there a fix necessary to either this library or ember-basic-dropdown to avoid this issue?

simonihmig:
@bgentry are you sure this was caused by this change? The only code change was to enable changing `renderInPlace`, before it was always true (same as when not specifying it now)...

Also, on https://www.ember-bootstrap.com/#/components/dropdown, when using a narrow screen that would cause the menu to be clipped, the menu is correctly moved to the left to fit in the view:

bgentry:
@simonihmig heh, no, clearly it couldn't have been caused by this specific change 🤦‍♂️ Sorry about that. 

Not really sure where else to look for the cause. The dropdowns in question don't have any real customization to them 🤷‍♂️

PR #589, opened 2018-05-03
Fix for issue with destroyed form elements

BartTK:
In some situations I get the following error:
`Assertion Failed: Cannot modify dependent keys for 'validation' on '<???@component:bs-form/element::ember1765>' after it has been destroyed.`

This problem is caused by getting 'hasFeedback' on an object that 'isDestroying'. Changed the order of both checks to fix it.


simonihmig:
@BartTK Curios, what version of Ember were you using to see this? Never happened to me...

BartTK:
I am using Ember 2.18.2. But this issue is rare. The site I am working has dozens of pages with bootstrap forms. Only one of those page is experiencing this issue when it gets refreshed.

Issue #582, opened 2018-04-21
Add crossfade option to Carousel


simonihmig:
Support for `.carousel-fade` added in BS 4.1

PR #591, opened 2018-05-05
Add carousel-fade to BS4 carousels.

nbrookie:
Took a stab at adding the carousel fade functionality to a BS4 only carousel.

When a `transition` param is set to `fade`, it will add the 'carousel-fade' class to the carousel so all slide transitions are fade transitions.

Is there a way to signify that some params are for BS4 only?

Addresses https://github.com/kaliber5/ember-bootstrap/issues/582

### DIFF
--- a/addon/components/bs4/bs-carousel.js
+++ b/addon/components/bs4/bs-carousel.js
@@ -1,8 +1,12 @@
 import Carousel from 'ember-bootstrap/components/base/bs-carousel';
+import { equal } from '@ember/object/computed';
+
 
 export default Carousel.extend({
   nextControlClassName: 'carousel-control-next',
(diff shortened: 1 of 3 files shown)


addon/components/bs4/bs-carousel.js
```diff
@@ -1,8 +1,12 @@
 import Carousel from 'ember-bootstrap/components/base/bs-carousel';
+import { equal, readOnly } from '@ember/object/computed';
```
simonihmig: `readOnly` is not used, which makes ESLint fail.

nbrookie: Ah, whoops. Just pushed a fix.

simonihmig:
Thanks again!

PR #604, opened 2018-06-10
Set all transitions to 0 in testing mode to speed up test runs

### LINKED ISSUE #595: Set transitions to 0 during testing
I love how this addon makes the Bootstrap components available in a native Ember fashion. It's been great for client projects where they are used to having the conveniences of Bootstrap components but are moving to an Ember front-end.

One improvement I'd love to see is to set all of the transition durations to 0 when the consuming app is running in the test environment. I have a number of acceptance tests which are exercising ember-bootstrap components (like the modal), and the test suite runs much slower with the default transition timings than it does when I override the transition durations to 0.

~I've tried extending the `{{bs-modal}}` component to override the transition durations in testing myself, but when I create the following file:~

```js
// app/components/bs-modal.js
import BSModalComponent from 'ember-bootstrap/components/bs-modal';

export default BSModalComponent.extend({
});
```

~Then I get the following error:~

```sh
Uncaught Error: Cannot instantiate a component without a renderer. Please ensure that you are creating <myapp@template:components/bs-modal::ember727> with a proper container/registry.
```

~I haven't had time as of yet to investigate this further, but it seems to indicate that there's something in the way ember-bootstrap's components are being defined that makes extending them difficult.~

---

```
Npm packages:
ember-bootstrap: ^1.2.1 -> 1.2.2
ember-cli: ~3.1.4 -> 3.1.4
bootstrap: ^4.0.0 -> 4.1.1
bootstrap-sass: n/a
ember-cli-sass: ^7.1.7 -> 7.1.7
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapFont: false
importBootstrapCSS: false
```

simonihmig:
Fixes #595

### DIFF
--- a/addon/utils/transition-duration.js
+++ b/addon/utils/transition-duration.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+const { testing: isTesting } = Ember;
+
+export default function transitionDuration(duration) {
+  return isTesting ? duration : duration;
+}
--- a/addon/utils/transition-duration.js
+++ b/addon/utils/transition-duration.js
@@ -3,5 +3,5 @@ import Ember from 'ember';
 const { testing: isTesting } = Ember;
 
 export default function transitionDuration(duration) {
-  return isTesting ? duration : duration;
+  return isTesting ? 0 : duration;
 }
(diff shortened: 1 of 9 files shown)


addon/utils/transition-duration.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
+
+const { testing: isTesting } = Ember;
+
+export default function transitionDuration(duration) {
+  return isTesting ? duration : duration;
```
kpfefferle: Is this intended to return `0` when `isTesting` is true? It looks like this is returning the `duration` argument either way...

simonihmig: Oh no! 😬
I played around with this, and seems I pushed the wrong version 🤦‍♂️
Nice catch, thank you!

simonihmig:
Closing this in light of #608, will come up with a new PR soon...

PR #607, opened 2018-06-12
docs(Button): document pendingText, resolvedText and rejectedText properties

jelhan:


jelhan:
I don't think the failing test is related to this PR. It only fails for the combination of one scenario (`ember-default`) and one browser (`IE11`). It's a unit test for a utility function. Maybe the test is flickering?

simonihmig:
Thank you!

> Maybe the test is flickering?

Yes, indeed. I need to fix this, it's annoying as hell...

PR #619, opened 2018-06-28
Fix `calling set on destroyed element` issue for bs-tooltip

brandynbennett:
It's possible if you very quickly try and open a tooltip while changing 
routes, to get a console error because the tooltip tries to call `set` 
during the transition.

This can happen if the user quickly moves the cursor across a tooltip trigger on their way to click a link. Hard to reproduce reliably, but I can make it happen every so often in my app.


simonihmig:
Thank you! 👍

PR #624, opened 2018-07-09
Render BS3 dropdowns in wormhole

Techn1x:
Mentioned here; https://github.com/kaliber5/ember-bootstrap/pull/570#issuecomment-393400624

Added support for BS3 dropdowns to be rendered in wormhole when `renderInPlace=false` on the menu component.

I mostly just copied the work from BS4, read over it & tested and it seems to work fine.

A few things I'm not sure about:
* I set `tagName` to empty instead of `<ul>`, to match BS4, it seems fine when either a wormhole or not, I think this is correct? But it seems to make some dropdown related tests fail for LTS versions
* The two menu components are identical, should they be combined like I did for the menu component templates?


Techn1x:
Thanks for the feedback. I've put the common dropdown menu code into `base` and put the BS3 menu tagname back to `<ul>`

Techn1x:
Actually, it doesn't quite work as I expected using tagName `<ul>` on the menu, it ends up doing the following;

renderInPlace=true
```
<button /> (dropdown button)
<ul> (the menu)
<div> (ember-popper element)
<li> abc </li> (list items)
</div>
</ul>
```

renderInPlace=false
```
<button /> (dropdown button)
<ul>
</ul>

(ember wormhole)
<div> (ember-popper element)
<li> abc </li> (list items)
</div>
```

So I think I need to set the tagName of the menu back to empty, and somehow set the tagName of the ember-popper to ul?

simonihmig:
> So I think I need to set the tagName of the menu back to empty, and somehow set the tagName of the ember-popper to <ul>?

Yes, makes sense. We already pass the class here for BS4, should add `tagName` (for BS4 it's `div` by default, which is ok there): https://github.com/kaliber5/ember-bootstrap/blob/bbb45998513b84ceab7b9392ad10d56ee38c5080/addon/templates/components/bs4/bs-dropdown/menu.hbs#L3

Techn1x:
I did give that a go, but unfortunately expects its `tagName` to be '' and then it passes that class attribute down to a child `div` - so you can't really set the tagName on ember-popper at the moment

https://github.com/kybishop/ember-popper/blob/master/addon/templates/components/ember-popper.hbs

Techn1x:
Hey @simonihmig just checking in, is a PR to ember-popper the only way to get this resolved?

simonihmig:
Sorry, missed your previous comment while being on vacation...

The only thing that comes to my mind is to wrap the `{{yield ...}}` in a `<ul class="...">`, in the template for the BS3 menu only. Something like...
```hbs
{{#ember-popper
    placement=popperPlacement
    popperTarget=toggleElement
    renderInPlace=renderInPlace
    popperContainer="#ember-bootstrap-wormhole"
    modifiers=popperModifiers
    registerAPI=(action "registerPopperApi")
  }}
  <ul
    class="dropdown-menu {{alignClass}}"
    role={{ariaRole}}
  >
    {{yield ...}}
  </ul>
{{(ember-popper}}
```

*(just written offhand, classes etc. might be wrong)*

This would have a wrapping `<div>` around the `<ul>` menu (from ember-popper), but that should hopefully not be a problem?

Techn1x:
Ah, that makes sense. I gave that a try (see commit https://github.com/kaliber5/ember-bootstrap/pull/624/commits/7728341d5bd93863b245500d4ef2e3f4b2efd49e )

It seems that when the `.dropdown-menu` class (which contains `z-index:1000`) is not on the ember-popper element, the popper element & menu are shown behind other elements:
**renderInPlace=true**
untitled15

**renderInPlace=false**
untitled16

So I also had to put `z-index:1000` on the ember-popper element, and that fixed it! I tested all directions/alignments, works as expected. I think I did this correctly.

Should be good to merge now unless there are more changes needed?

Techn1x:
Hey @simonihmig this is passing the tests now and won't be a breaking change for existing BS3 users. Should be good to be merged in when you've got a chance

If you're not happy with the zindex fix I did let me know

Techn1x:
Woohoo! No worries at all, we're all busy people :)

Issue #605, opened 2018-06-10
Fix flakey carousel tests


simonihmig:
* `bs-carousel: carousel ltr=false does right-to-left automatic sliding`
* `bs-carousel: carousel ltr=true does left-to-right automatic sliding`

PR #625, opened 2018-07-09
feat(Button): bs-button yields state and shorthands

jelhan:
Example usage to show a slider if button triggered by `onClick` event
is pending:

```
{{#bs-button onClick=(action 'download') |button|}}
  Download
  {{#if button.isPending}}
    {{fa-icon "slider"}}
  {{/if}}
{{/bs-button}}
```

### DIFF
--- a/addon/components/base/bs-button.js
+++ b/addon/components/base/bs-button.js
@@ -1,6 +1,7 @@
 import { scheduleOnce } from '@ember/runloop';
 import Component from '@ember/component';
 import { observer, computed } from '@ember/object';
+import { deprecatingAlias, equal, or }  from '@ember/object/computed';
 import layout from 'ember-bootstrap/templates/components/bs-button';
 import TypeClass from 'ember-bootstrap/mixins/type-class';
 import SizeClass from 'ember-bootstrap/mixins/size-class';
@@ -28,33 +29,14 @@ import SizeClass from 'ember-bootstrap/mixins/size-class';
  {{/bs-button}}
  ```
 
- ### States
-
- Use the `textState` property to change the label of the button. You can bind it to a controller property to set a "loading" state for example.
- The label of the button will be taken from the `<state>Text` property.
-
- ```hbs
- {{bs-button type="primary" icon="glyphicon glyphicon-download" textState=buttonState defaultText="Download" loadingText="Loading..." onClick="download"}}
- ```
-
- ```js
- App.ApplicationController = Ember.Controller.extend({
-   buttonState: "default"
-   actions: {
-     download: function() {
-       this.set("buttonState", "loading");
-     }
-   }
- });
- ```
-
  ### Promise support for automatic state change
 
  When returning a Promise for any asynchronous operation from the `onClick` closure action the button will
- manage an internal state ("default" > "pending" > "resolved"/"rejected") automatically, changing its label according to the state of the promise:
+ manage an internal state ("default" > "pending" > "fulfilled"/"rejected") automatically, changing its label
+ according to the state of the promise:
 
  ```hbs
- {{bs-button type="primary" icon="glyphicon glyphicon-download" defaultText="Download" pendingText="Loading..." resolvedText="Completed!" rejectedText="Oups!?" onClick=(action "download")}}
+ {{bs-button type="primary" icon="glyphicon glyphicon-download" defaultText="Download" pendingText="Loading..." fulfilledText="Completed!" rejectedText="Oups!?" onClick=(action "download")}}
  ```
 
  ```js
@@ -67,6 +49,20 @@ import SizeClass from 'ember-bootstrap/mixins/size-class';
    }
  });
  ```
+
+ For further customization `isPending`, `isFulfilled`, `isRejected` and `isSettled` properties are yielded:
+
+ ```hbs
+ {{#bs-button onClick=(action "download") as |button|}}
+  Download
+  {{#if button.isPending}}
+    <span class="loading-spinner"></span>
+  {{/if}}
+ {{/bs-button}}
+ ```
+
+ You can `reset` the state represented by these properties and used for button's text by setting `reset` property to
+ `true`.
 
  @class Button
  @namespace Components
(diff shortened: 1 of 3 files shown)


jelhan:
WIP cause not backward compatible yet. Missing `deprecatingAlias` for `textState`.

jelhan:
Backward compatibility added. Let me know if you don't like the renaming of `textState` to `state`. Maybe `state` is to vague but it's also not only about the state of the text. Actually it represents the state of the promise returned by `onClick` event but I didn't come up with a better naming.

jelhan:
I don't think failing tests are related to this PR. Seems like tests for bs-carousel are flickering since they are failing only for some ember releases.

simonihmig:
> I don't think failing tests are related to this PR. Seems like tests for bs-carousel are flickering since they are failing only for some ember releases.

Yes: #605 🙄

addon/components/base/bs-button.js
```diff
@@ -232,12 +244,59 @@ export default Component.extend(TypeClass, SizeClass, {
...
+
+  /**
+   * @property textState
+   * @alias state
+   * @deprecated
+   * @protected
+   */
+  textState: deprecatingAlias('state', {
+    id: 'ember-bootstrap.bs-button-text-state',
+    until: '3.0.0'
```
simonihmig: Not sure if we really need to do this, as the prop was never declared public. Not sure if the `protected` label makes so much sense, I usually only declare props either `public` or `private`, but not here, probably because the code base is quite old...

But it down not do any harm (the deprecating alias), so let's keep it. But for the future, can you label alls the new props `@private` please?

tests/integration/components/bs-button-test.js
```diff
@@ -105,21 +105,21 @@ module('Integration | Component | bs-button', function(hooks) {
 
   test('button text is changed according to button state', async function(assert) {
 
-    await render(hbs`{{bs-button defaultText="text1" loadingText="text2" textState=textState}}`);
+    await render(hbs`{{bs-button defaultText="text1" loadingText="text2" state=state}}`);
```
simonihmig: The tests are not ideal IMHO, in that they use a non-public property (`state`/`textState`) for testing. Instead they should better test the real public API, which is not about setting `state`, but returning a promise and reacting to its state. 

But don't be offended, I am basically complaining about myself, who has introduced this in the first place! ;)

But this can be refactored in a separate PR as well...

addon/components/base/bs-button.js
jelhan: I wasn't quite sure how we notice in API docs that a property is yielded. This doesn't seem to be covered by JSDoc. But sure will change them to `@private`.

Since passing `textState` in was covered by integration tests, I thought this might be considered public. Therefore I added the deprecation. If it's not, I could just change the until value to a minor release (e.g. `2.2.0`) as ember does it for highly used private APIs?

tests/integration/components/bs-button-test.js
jelhan: Was also confused about that one and therefor thought it's considered public API. Since there isn't any test that it actual reflects the state of returned promise, I will refactor in this PR.

addon/components/base/bs-button.js
simonihmig: > This doesn't seem to be covered by JSDoc.

Yes, indeed. https://github.com/ember-learn/ember-cli-addon-docs has support for this, which I am quite jealous for ;)

> I could just change the until value to a minor release (e.g. 2.2.0) as ember does it for highly used private APIs?

Yes, but honestly I don't expect this to be used at all, so I am fine with just renaming and marking it as private actually!

jelhan:
Addressed all feedback.

addon/components/base/bs-button.js
```diff
@@ -273,22 +336,20 @@ export default Component.extend(TypeClass, SizeClass, {
...
-    this.set('textState', 'default');
+    this.set('state', 'default');
   },
 
   resetObserver: observer('reset', function() {
     if (this.get('reset')) {
-      scheduleOnce('actions', this, function() {
-        this.set('textState', 'default');
-      });
+      scheduleOnce('actions', this, 'resetState');
```
jelhan: Do you think we need test coverage that `state` is not set twice in a single runloop? Since there weren't any issues so far, triggering `reset` twice in a single run loop is not happening in real world applications.

tests/integration/components/bs-button-test.js
```diff
@@ -104,52 +112,163 @@ module('Integration | Component | bs-button', function(hooks) {
...
+        later(() => {
+          resolve();
+        }, 10);
+      });
+    });
+    this.set('reset', false);
 
-    assert.dom('button').hasText('text2');
+    await render(hbs`{{#bs-button onClick=clickAction reset=reset as |button|}}{{button.state}}{{/bs-button}}`);
+    assert.dom('button').hasText('default');
```
jelhan: By the way: I don't like the naming `default` for a `state`  if the button wasn't clicked yet. It was fitting for `defaultText` but as a state it doesn't feel intuitive. Do you have a better naming before we expose that one as public API? As an alternative we may just that `state` to `null` if but wasn't clicked yet? We might or might not also rename `defaultText` to `text` in that case.

jelhan:
Thought about that one another time and pushed some changes:

- Fixed the promise terminology. It's not yielding `isPending`, `isFulfilled`, `isRejected` and `isSettled`. Also renamed `resolvedText` to `fulfilledText`.
- It does not yield `state` anymore. Since ember does not ship a equal comparison helper by default, it's not that helpful. Information is also available and easier to consume through the boolean properties being yielded.
- Removed the chapter about `textState` / `state` since we consider it as private.
- Removed the deprecating alias `textState`. Renaming `resolved` state to `fulfilled` would add other difficulties.

jelhan:
Refactored the tests to use `RSVP.defer()`. They should be more stable now. It's also more readable IMO.

I think the PR is ready to get merged.

jelhan:
I don't think the failing test for *ember-default* scenario is related to this PR. It was canceled cause to output received for 10 minutes. I think it was an issue with Sauce Labs. Seems like tests for Chrome, Firefox and IE11 has passed before.

simonihmig:
Restarted, Soucelabs is a bit flakey unfortunately...

simonihmig:
👍

PR #626, opened 2018-07-09
docs(FormElement): integration plugins for custom control

jelhan:


addon/components/base/bs-form/element.js
```diff
@@ -139,6 +139,11 @@ const nonDefaultLayouts = A([
  {{/bs-form}}
  ```
 
+ If yu are using the custom control quite often, you should consider writing an integration plugin like
```
Techn1x: I hate to be pedantic but second word has a typo :)
*you

jelhan: Thanks for noticing. It's fixed.

simonihmig:
Thanks!

Issue #92, opened 2016-05-26
submit button positioned wrong in horizontal form


jelhan:
Submit button is missing `form-control` class and column definition in horizontal form.

Generated HTML (taken from dummy app):

```
<form id="ember541" role="form" class="ember-view form-horizontal">
    <button id="ember602" type="submit" class="ember-view btn btn-primary"><!---->Submit<!----> </button>
</form>
```

Should be ([bootstrap documentation](http://getbootstrap.com/css/#forms-horizontal)):

```
<form class="form-horizontal">
  <div class="form-group">
    <div class="col-sm-offset-2 col-sm-10">
      <button type="submit" class="btn btn-default">Sign in</button>
    </div>
  </div>
</form>
```

A work-a-round is to wrap `bs-button` in `bs-form-element` if it's a horizontal form:

```
{{#bs-form formLayout="horizontal"}}
  {{#bs-form-element}}
    {{bs-button defaultText="Submit" type="primary" buttonType="submit"}}
  {{/bs-form-element}}
{{/bs-form}}
```

At least this should be documented.

simonihmig:
It is neither the scope of `bs-form` nor `bs-button` to add that markup, so yes, using `bs-form-element` is probably the way to go. So it is indeed more of a documentation thing...

jelhan:
Wrong examples in documentation are at:
* https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form.js#L40-L46
* https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form-element.js#L21-L28
* https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form-element.js#L35-L45
 
But I'm not quite sure if it should be treated as documentation issue. To fix dummy app this code would be needed:

```
{{#bs-form formLayout=formLayout model=this action="submit"}}
  {{#if (eq formLayout 'horizontal')}}
    {{#bs-form-element}}
      {{bs-button defaultText="Submit" type="primary" buttonType="submit"}}
    {{/bs-form-element}}
  {{else}}
    {{bs-button defaultText="Submit" type="primary" buttonType="submit"}}
  {{/if}}
{{/bs-form}}
```

And I guess choosing between horizontal or vertical form layout depending on window width is a common use case. Perhaps a `bs-form-submit` component is needed?

jelhan:
[...]

PR #627, opened 2018-07-10
feat(Form): yields isSubmitting state

jelhan:


addon/components/base/bs-form.js
```diff
@@ -220,26 +246,40 @@ export default Component.extend({
...
 
-    if (!this.get('hasValidator')) {
-      return this.get('onSubmit')(model);
-    } else {
-      let validationPromise = this.validate(this.get('model'));
-      if (validationPromise && validationPromise instanceof RSVP.Promise) {
-        validationPromise
-          .then((r) => this.get('onSubmit')(model, r))
-          .catch((err) => {
-            this.set('showAllValidations', true);
```
jelhan: This also fixes a bug if component is already destroyed when `validationPromise` rejects.

addon/components/base/bs-form.js
```diff
@@ -220,26 +246,40 @@ export default Component.extend({
...
+            }
+          });
+      })
+      .catch((err) => {
+        if (!this.get('isDestroyed')) {
+          this.set('showAllValidations', true);
+          this.set('isSubmitting', false);
+        }
+
+        this.get('onInvalid')(model, err);
```
jelhan: Not quite sure if `onInvalid` should also be called if component is already destroyed. To be honest, I'm not sure about the previous behavior. Is it still executed after `this.set()` has thrown cause of setting on a destroyed object?

tests/integration/components/bs-form-test.js
```diff
@@ -159,6 +161,29 @@ module('Integration | Component | bs-form', function(hooks) {
     assert.dom(`.${formFeedbackClass()}`).hasText('There is an error');
   });
 
+  test('Yields isSubmitting property', async function(assert) {
+    this.set('submit', () => {
+      return new RSVP.Promise((resolve) => {
+        later(() => {
+          resolve();
+        }, 10);
```
jelhan: Could also try awaiting next run loop here, which might be a little bit faster. But it's only 10ms and awaiting next run loop would fail as soon as `submit` function needs more than one runloop to execute.

tests/unit/components/bs-form-test.js
```diff
@@ -0,0 +1,91 @@
...
+import { next, run } from '@ember/runloop';
+import RSVP from 'rsvp';
+
+const nextRunloop = function() {
+  return new RSVP.Promise((resolve) => {
+    next(() => resolve());
+  })
+};
+
+module('Unit | Component | bs-form', function(hooks) {
```
jelhan: Hope it's okay to use a unit test for `isSubmitting` property. Had the feeling that it's a little bit easier to write and read. Could refactor to integration test if you would prefer that one.

addon/components/base/bs-form.js
```diff
@@ -220,26 +246,40 @@ export default Component.extend({
...
-          .catch((err) => {
-            this.set('showAllValidations', true);
-            return this.get('onInvalid')(model, err);
-          });
-      }
+    let validationPromise = this.get('hasValidator') ? this.validate(this.get('model')) : RSVP.resolve();
+    if (!isPromise(validationPromise)) {
+      validationPromise = RSVP.resolve();
     }
+    validationPromise
```
jelhan: Could also return a Promise from `submit()` but not quite sure what that would be usefull for. It also hasn't returned anything before.

addon/components/base/bs-form.js
```diff
@@ -136,6 +137,31 @@ export default Component.extend({
...
+   * ```
+   * {{#bs-form onSubmit=(action 'save') as |form|}}
+   *   {{#bs-button type='submit' disabled=form.isSubmitting}}
+   *     Save {{fa-icon 'spinner'}}
+   *   {{/bs-button}}
+   * {{/bs-from}}
+   * ```
+   *
+   * The value is not accurate if multiple submit events are pending. In that case `isSubmitting` is `false` as soon
+   * as the first submission fulfills.
```
jelhan: If we like to support that use case, this limitation could also be dropped. I think adding a counter of submissions and only setting `isSubmitting` to `false` for the last pending one should be enough.

addon/components/base/bs-form.js
```diff
@@ -220,26 +246,40 @@ export default Component.extend({
...
+      validationPromise = RSVP.resolve();
     }
+    validationPromise
+      .then((r) => {
+        let onSubmitPromise = this.get('onSubmit')(model, r);
+        if (!isPromise(onSubmitPromise)) {
+          onSubmitPromise = RSVP.resolve();
+        }
+        onSubmitPromise
+          .finally(() => {
```
simonihmig: I think we must be careful with `finally()`. It is available when RSVP is used, and has been introduced to the ES-spec lately IIRC, but there still can be native Promise implementation around that don't support it. 

maybe we can solve this by restructuring the code above a bit, to always start with a `RSVP.resolve()` and call `.then(()  => this.get('onSubmit')(model, r))` (doesn't matter if it return a promise now or not). Now it *should* always be a RSVP promise which supports `finally()`, right?

addon/components/base/bs-form.js
simonihmig: Not sure, but I guess a destroyed thing should never call anything!? 😬

addon/utils/is-promise.js
```diff
@@ -0,0 +1,5 @@
+import RSVP from 'rsvp';
+
+export default function(val) {
+  return val instanceof RSVP.Promise || val instanceof Promise;
```
simonihmig: Not sure: wouldn't `val instanceof Promise` throw if `Promise` is undefined (i.e. not supported natively)?

addon/components/base/bs-form.js
```diff
@@ -220,26 +246,40 @@ export default Component.extend({
...
    */
   submit(e) {
+    this.set('isSubmitting', true);
+
     if (e) {
       e.preventDefault();
     }
     let model = this.get('model');
 
     this.get('onBefore')(model);
```
simonihmig: I wonder, should we also introduce promise support for this action?

addon/components/base/bs-form.js
```diff
@@ -136,6 +137,31 @@ export default Component.extend({
...
+   * {{/bs-from}}
+   * ```
+   *
+   * The value is not accurate if multiple submit events are pending. In that case `isSubmitting` is `false` as soon
+   * as the first submission fulfills.
+   *
+   * @property isSubmitting
+   * @type {Boolean}
+   * @readonly
+   * @public
```
simonihmig: I think I would rather declare this private, and move the docs for the yielded `isSubmitting` to the top to the component docs itself. The user is not really using this prop directly, it is just yielded, which is kind of like copying the value to some other context. We could for example rename the (private) prop here and still yield it under the same name...

addon/components/base/bs-form.js
```diff
@@ -136,6 +137,31 @@ export default Component.extend({
...
+   *
+   * `isSubmitting` property is yielded and could be used to customize form submit button. E.g. a loading spinner could
+   * be shown and the submit button been disabled:
+   *
+   * ```
+   * {{#bs-form onSubmit=(action 'save') as |form|}}
+   *   {{#bs-button type='submit' disabled=form.isSubmitting}}
+   *     Save {{fa-icon 'spinner'}}
+   *   {{/bs-button}}
+   * {{/bs-from}}
```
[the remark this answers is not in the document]
simonihmig: Thinking further, does it maybe even make sense to align the APIs here, so we yield something like `submitState` or so, which has all the states introduced in the button component, like `isPending` (equivalent to `isSubmitting`), but also `isResolved` and `isRejected`?

tests/unit/components/bs-form-test.js
simonihmig: Hm, tests look good, but I indeed think an integration test would be better. See my other comment about making `isSubmitting` private, which would require to test for the yielded value, and not the property itself.

addon/components/base/bs-form.js
jelhan: I'm not convinced of adding `ember-concurrency` as dependency to be honest. It's a great addon but is it really needed? In app code it's great to keep things simple but it's a heavy dependency for addons and couples libraries with different scope. Also makes internals harder to understand for users and contributors not used to `ember-concurrency`.

Also I'm not quite sure if we should force the submit to be not fired again before the last submission is fulfilled. It would match my current use cases but there may be others. So it should be at least configurable. I'm willing to add that as a configurable option in another PR. It's basically just checking `isSubmitting === true`.

addon/components/base/bs-form.js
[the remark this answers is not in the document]
jelhan: I'm definitely in favor with a `{{#form.button}}` component but I think it's out of scope for this PR. Should also address #92.

I don't like yielded hash. It's quite complex API to use. You would end up with lot of code, which is hard to read:

```
{{#bs-button disabled=submitState.isPending as |button|}}
  {{#if submitState.isResolved}} Saved {{else}} Save {{/if}}
  {{#if submitState.isPending}} {{fa-icon 'spinner'}} {{/if}}
{{/bs-button}}
```

IMO `isSubmitting` and `isSubmitted` are easier to understand. At least the first one would also be inline with `isSaving` property of `DS.Model`.
```
{{#bs-button disabled=isSubmitting as |button|}}
  {{#if isSubmitted}} Saved {{else}} Save {{/if}}
  {{#if isSubmitting}} {{fa-icon 'spinner'}} {{/if}}
{{/bs-button}}
```

Not quite sure about an `isRejected` equivalent. Not only because of naming but also cause it does not to distinguish between failed validation, server-side validation error and other server-side errors. So might just lead to wrong assumptions.

It's a different story for a `{{bs-button}}` cause we don't know what the attached action is about.

addon/components/base/bs-form.js
jelhan: Yeah, but should be done in a separate PR shouldn't it?

jelhan:
Addressed all feedback except:
- naming (`isSubmitting` vs `submitState.isPending`)
- using `ember-concurrency`

`isSubmitting` is now accurate even if multiple submit actions are fired. It stays `true` until all pending submit has been fulfilled.

Some stuff could be addressed in follow up PRs:
- Add a configuration option to prevent concurrent submissions.
- Provide a `{{form.button}}` component which provides some magic about state handling and layout.
- Expose other submission states (equivalents to `isResolved`, `isRejected`, `isFulfilled`) and allow to `reset` the submission state.

simonihmig:
@jelhan sorry for the delay, I just came back from vacation, and still under load of work... will try to re-review your PRs hopefully soon!

addon/components/base/bs-form.js
simonihmig: sure

tests/integration/components/bs-form-test.js
```diff
@@ -159,6 +161,145 @@ module('Integration | Component | bs-form', function(hooks) {
...
+    await this.render(hbs`{{#bs-form onSubmit=submitAction as |form|}}
+      <div class='state {{if form.isSubmitting 'is-submitting'}}'></div>
+    {{/bs-form}}`);
+
+    assert.dom('form .state').doesNotHaveClass('is-submitting');
+
+    triggerEvent('form', 'submit');
+
+    await waitFor('.is-submitting');
+    assert.dom('form .state').hasClass('is-submitting');
```
simonihmig: There are some test runs where this is failing, not sure why though!?

tests/integration/components/bs-form-test.js
```diff
@@ -159,6 +161,145 @@ module('Integration | Component | bs-form', function(hooks) {
...
+
+    assert.dom('form .state').doesNotHaveClass('is-submitting');
+
+    triggerEvent('form', 'submit');
+    await waitFor('.is-submitting');
+    assert.dom('form .state').hasClass('is-submitting');
+
+    await new Promise((resolve) => {
+      later(() => resolve(), 11)
+    });
```
simonihmig: Can you  please explain this? Shouldn't `submitAction` and `validate` be fulfilled after 11ms?

tests/integration/components/bs-form-test.js
jelhan: I noticed that using timers to ensure not resolving in a too early runloop is not that reliable. I will refactor that tests to use `RSVP.defer()`. Should also improve readability.

jelhan:
Refactored tests to use `RSVP.defer()`. Should be more stable and it's more readable IMO. Still have to await the next runloop at some points to make sure that any UI changes caused by fired action or settled promise has been fulfilled cause `waitUntil` or `waitFor` helpers can't be used for that cases.

Also removed duplicate assertions like `waitFor('form.is-submitting'); assert.dom('form').hasClass('is-submitting')`. If the class is not there `waitFor` would timeout. Therefore it's also an implicit assertion.

I think PR is ready to be merged.

jelhan:
Tests are failing for ember 2.3, 2.8 and 2.12. Could reproduce locally. Will investigate that issue later.

jelhan:
Fixed tests for ember <= 2.12. For unknown reasons `onSubmit` action is not executed on next runloop. Have to await two runloops therefor.

jelhan:
Failing tests for ember-release and ember-beta BS 3 scenarios are due to yarn connection / download issues. Please restart them. Safari on Sauce Labs does not seem to work currently. It's always failing. Same for #625

simonihmig:
Merged! Thank you! 🎉

> Safari on Sauce Labs does not seem to work currently

Disabled them for now... see #636

PR #587, opened 2018-04-26
Yield actions in dropdown & navbar

sdhull:
Navbar already yielded open & close actions but forcing developers to double-bookkeep the open/closed status is unnecessary when we can just yield the toggle action as well.
Dropdown is impossible to control programmatically without yielding the actions.

Fixes #559

PR #639, opened 2018-08-07
bs-dropdown: adding actions to yielded hash

sdhull:
As requested in #587, here is the half of that PR that touches `bs-dropdown`.

As mentioned in that PR, I couldn't rename the actions (eg, from `toggleDropdown` to `toggle`) since there are already items in that hash with those names (eg, `toggle` is the dropdown's toggle component).


sdhull:
Oh yeah forgot about that. Sure I'll add some

sdhull:
@simonihmig check em out! 🎉

PR #664, opened 2018-09-18
Adds Disabled Property to Accordion

### LINKED ISSUE #287: Add disabled property to accordion item
Hi again,
I have another feature request for the accordion: Maybe you could add a disabled property to the item so that, by setting it as true, opening that specific item would be inhibited? It would make sense to make such properties available to all components that trigger an action, actually, to keep the logic consistent.
Thank you.

donaldwasserman:
Fixes #287 

@simonihmig Per our ~slack~ discord conversation. Not sure if this is the approach you'd prefer for dealing with the click handler

### DIFF
--- a/addon/components/base/bs-accordion/item.js
+++ b/addon/components/base/bs-accordion/item.js
@@ -19,10 +19,9 @@ export default Component.extend(TypeClass, {
   layout,
 
   /**
-   * We override the mixin here to include an extra class.
-   * @see TypeClass
+   * Add additional binding to mark the entire component area disabled
    */
-  classNameBindings: ['typeClass', 'disabled'],
+  classNameBindings: ['disabled'],
   /**
    * The title of the accordion item, displayed as a .panel-title element
    *
--- a/addon/components/base/bs-accordion/item.js
+++ b/addon/components/base/bs-accordion/item.js
@@ -19,6 +19,11 @@ export default Component.extend(TypeClass, {
   layout,
 
   /**
+   * We override the mixin here to include an extra class.
+   * @see TypeClass
+   */
+  classNameBindings: ['typeClass', 'disabled'],
+  /**
    * The title of the accordion item, displayed as a .panel-title element
    *
    * @property title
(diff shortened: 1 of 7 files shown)


donaldwasserman:
@simonihmig Sorry, _now_ things should be passing. I forgot to run bootstrap 3 tests locally.

donaldwasserman:
Looks like travis sort of flaked out there for some reason -- looks like the tests just need to get re-run to pass.

simonihmig:
> Looks like travis sort of flaked out there for some reason -- looks like the tests just need to get re-run to pass.

Yeah, I always have at least one test failing for some reason (yarn registry down, browser failure, whatever). I restarted and it's green now! 🎉

Thanks again for your effort, I will review this asap, but probably not before the weekend...

addon/components/base/bs-accordion/item.js
```diff
@@ -18,6 +18,11 @@ import layout from 'ember-bootstrap/templates/components/bs-accordion/item';
 export default Component.extend(TypeClass, {
   layout,
 
+  /**
+   * We override the mixin here to include an extra class.
+   * @see TypeClass
+   */
+  classNameBindings: ['typeClass', 'disabled'],
```
simonihmig: I don't think you need to add `typeClass` here, as `classNameBindings` will merge the items used in parent classes/mixins, IIRC

addon/components/base/bs-accordion/item/title.js
```diff
@@ -23,6 +23,13 @@ export default Component.extend({
    */
   collapsed: null,
 
+  /**
+   * @property disabled
+   * @type boolean
+   * @public
```
simonihmig: Maybe mark this as `@private`, as it is only used internally, at users should set it at the `item`, not the `title`

addon/components/base/bs-accordion/item.js
donaldwasserman: I didn't realize mixins magically combined merged those together...that's some interesting "magic"

simonihmig: Yeah, Ember has some places of extraordinary magic indeed! ;) 

I guess some of these are maybe going to go away some day, when moving to native classes etc. Here for the record: https://emberjs.com/api/ember/3.4/classes/EmberObject/properties/mergedProperties?anchor=mergedProperties

simonihmig:
Thank you @donaldwasserman!

donaldwasserman:
Thank you, @simonihmig!

PR #667, opened 2018-09-25
Autocapitalize and autocorrect attributes

CodingItWrong:
Adds `autocapitalize` and `autocorrect` attributes to text inputs and textareas, to provide more control of typing behavior on iOS devices (and maybe others)


simonihmig:
@CodingItWrong thank you! 👍

Issue #638, opened 2018-08-07
Improve support for handling interim states on buttons and forms


jelhan:
`{{bs-button}}`, `{{bs-form}}` and `{{bs-modal-simple}}` are all dealing with interim states (pending actions). Currently support for them is quite limited. `{{bs-button}}` allows to change the button label depending on promises state. But that's all. The others does not help with showing interim states at all.

This issue should outline the bigger picture and track the work needed. Implementation has already been started. If possible, it should be accomplished without any breaking change to not have to wait for next major release.

# Common use cases

Many application shares the same logic for visual feedback about state of action (*pending*, *fulfilled*, *rejected*). Ember-bootstrap can't support all of them out of the box, but it should be as easy as possible to implement the application specific logic by overriding at little overrides as possible.

## Button

* Button should have another label while action is pending, fulfilled or rejected.
* Button should show a loading spinner while action is pending.
* Button should be disabled while action is pending.
* Button should have another type if action is fulfilled (`btn-success`) or rejected (`btn-danger`).
* Button should not fire another action if one is pending (like using an `ember-concurrency` task with `drop()` modifier).

## Form

* Submit button should have another label if action is pending, fulfilled or rejected.
* Submit button should show a loading spinner while action is pending.
* Submit button should be disabled while action is pending.
* Submit button should have another type if form has been submitted successfully (`btn-success`) or submission has failed (`btn-danger`). It should be reset automatically if form values are changed.
* Rollback button should be disabled while action is pending.
* Form should not submit if another submission is pending (like using an `ember-concurrency` task with `drop()` modifier).

## bs-modal-simple

Modal buttons should reflect submission states as normal form buttons would do if a form is used inside a `{{bs-form-simple}}`.

# Usage example

All mentioned use cases should be supported by overriding `{{bs-button}}` in application *only*. An example might look like:

```
// app/components/bs-button.js

import BsButton from 'ember-bootstrap/components/bs-button';
import { computed } from '@ember/object';
import { reads } from '@ember/object/computed';

export default BsButton.extend({
  disabled: reads('isPending'),
  type: computed('state', function() {
    switch (this.get('state')) {
      case 'fulfilled':
        return 'success';

      case 'rejected':
        return 'danger';

      default:
        return 'default';
    }
  })
});
```

```
// app/templates/components/bs-button.hbs

{{yield}}
{{#if isPending}}
  {{fa-icon "spinner" spin=true}}
{{/if}}
```

# Tasks

[...]

PR #669, opened 2018-09-28
feat(Button): optional prevent concurrent click actions

jelhan:
Adds a `preventConcurrency` option to `{{bs-button}}`. If `true`, clicking a button will not fire `onClick` action if a promise returned by an earlier execution is still pending.

This allows to replace ember-concurrency for simple use cases. It's basically the same as using an ember-concurrency task with `drop()` modifier as `onClick` action.

Part of #638

### DIFF
--- a/addon/components/base/bs-button.js
+++ b/addon/components/base/bs-button.js
@@ -235,6 +235,19 @@ export default Component.extend(TypeClass, SizeClass, {
   value: null,
 
   /**
+   * Controls if `onClick` action is fired concurrently. If `true` clicking button multiple times will not trigger
+   * `onClick` action if a Promise returned by previous click is not settled yet.
+   *
+   * This does not affect event bubbling.
+   *
+   * @property preventConcurrency
+   * @type Boolean
+   * @default false
+   * @public
+   */
+  preventConcurrency: false,
+
+  /**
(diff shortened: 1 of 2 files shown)


simonihmig:
Restarted the build...

addon/components/base/bs-button.js
```diff
@@ -234,6 +234,16 @@ export default Component.extend(TypeClass, SizeClass, {
    */
   value: null,
 
+  /**
+   * Controls if `onClick` action is fired concurrently. If `true` clicking button multiple times will not trigger
+   * `onClick` action if a Promise returned by previous click is not settled yet.
+   *
+   * This does not affect event bubbling.
+   *
+   * @property preventConcurrency
```
simonihmig: Can you please add the `@public`, `@default` and `@type` annotations, so the API docs get generated properly!

simonihmig:
> Added the missing docs.

@jelhan Hm, I am not seeing them!?

jelhan:
> @jelhan Hm, I am not seeing them!?

@simonihmig Missed to push the changes. Sorry for that. Pushed them now.

simonihmig:
Thank you for working on this!

PR #678, opened 2018-10-04
fix deprecated copy function usage

BnitoBzh:


simonihmig:
Thanks @BnitoBzh for your effort here. The changes are fine, compared to the previous state. However I think a better way would be to actually remove the need for the `copy` function, so we don't need to bring in `ember-copy` as an additional dependency.

It seems the copied value is only used at this place: https://github.com/kaliber5/ember-bootstrap/blob/9506f3fb10785307468c19467bb6db177c609a1f/addon/components/base/bs-button-group.js#L175

So first it would be better to only do the copy there, not at the beginning of the function. And instead of `copy()` I think we would be able to just do a simple `.slice()` on the array, to do a shallow copy with vanilla JS.

Would you be able to update your PR?

simonihmig:
Closing this in favor of #714

PR #681, opened 2018-10-17
fix(bs-dropdown/menu): Propagate class to the menu dom element

offirgolan:


tests/integration/components/bs-dropdown/menu-test.js
```diff
@@ -8,34 +8,83 @@ module('Integration | Component | bs-dropdown/menu', function(hooks) {
...
 
-    assert.dom('.dropdown-item').doesNotExist('has item component with no markup');
+    assert
+      .dom('.dropdown-item')
+      .doesNotExist('has item component with no markup');
+  });
+
+  testBS3('dropdown menu propagates class names', async function(assert) {
+    await render(
+      hbs`{{#bs-dropdown/menu align="right" isOpen=true toggleElement=this.element classNames="custom-class-1 custom-class-2"}}Something{{/bs-dropdown/menu}}`
```
simonihmig: I just have this question about this PR: my understanding is that `classNames` is commonly used *inside* the component.js to customize the applied classes, but passing custom class names from the outside is done using the simpler `class` property, as the guides also suggest here: https://guides.emberjs.com/release/components/customizing-a-components-element/#toc_customizing-the-elements-class

At other places I have been passing `class` around like here: https://github.com/kaliber5/ember-bootstrap/blob/263767b46b0f69df6d6c10b764c4ea6148c51dcc/addon/templates/components/common/bs-popover.hbs#L9

What do you think?

bravo-kernel: Good to see this ticket because  need the same functionality.  FWIW I have seen other addons make the distinction between `class` (sets/replaces all) and `extraClass`  (adds to whatever class is already in the component).

Either way, much appreciated, awesome addon 💯

offirgolan: @simonihmig good call. I'll make the change.

offirgolan:
@simonihmig I've updated the PR 😄

tests/integration/components/bs-dropdown/menu-test.js
```diff
@@ -8,34 +8,83 @@ module('Integration | Component | bs-dropdown/menu', function(hooks) {
...
+    assert.dom('.dropdown-menu').exists('menu has dropdown-menu class');
+    assert
+      .dom('.dropdown-menu')
+      .hasClass('custom-class-1', 'menu has custom-class-1 class');
+    assert
+      .dom('.dropdown-menu')
+      .hasClass('custom-class-2', 'menu has custom-class-2 class');
+  });
+
+  testBS4('dropdown menu propagates class names', async function(assert) {
```
simonihmig: Sorry, I missed this one on the first review: it seems both tests are exactly the same (or do I miss something?). If that's the case you can drop one and use the normal `test()` function for the other. `test()` runs always, the `testBSx()` function are just convenience test wrappers that run only for the given Bootstrap version.

offirgolan: Done.

bravo-kernel:
👏

lougreenwood:
Wow, I stumbled across this issue yesterday afternoon and was going to work on fixing it myself - such coincidence! 🐶 

However, it seems that using `class` to propagate classes down to the menu may not be flexible enough...

I see that earlier commits had a `_classNamesString` property, which was removed in favour of `class` - however in my use case this also doesn't work.

My use case involves extending ember-bootstrap components and applying additional styling using css-modules (this is a workaround the whole "BEM and bootstrap don't please nicely because of Sass @extend problems").
So, by using `ember-css-modules` my extending component injects a bunch of additional classes, unfortunately, these additional classes aren't included in the `class` property which the `bs-dropdown/menu` uses - rendering the `class` property useless.

bravo-kernel:
Am I correct that this fix has not been released yet?

@simonihmig used ember-bootstrap as the basis for this bootstrap 4 theme btw and works like a charm. Much much appreciated 💯 . https://github.com/alt3/ember-tabler

simonihmig:
> Am I correct that this fix has not been (npm) released yet?

I released 2.2.0 with this just about an hour ago! 😉 

> @simonihmig used ember-bootstrap as the basis for this bootstrap 4 theme btw and works like a charm. Much much appreciated 💯 . https://github.com/alt3/ember-tabler

Nice! Thank you!

bravo-kernel:
Thank you sir, for real. For both the release and the addon ;)

Issue #675, opened 2018-10-02
Nav: support fill option


simonihmig:
See http://getbootstrap.com/docs/4.1/components/navs/#fill-and-justify

`justified` is already supported. In a similar way `fill=true` should also be supported.

CoryKelly:
I can this a go!

dgavey:
Working on this one, should have a PR soon.

simonihmig:
Closed by #684

PR #684, opened 2018-10-26
Adds fill property to bs-nav

dgavey:
PR for issue #675 

Let me know if I missed anything, as it's the first time I've contributed to this repo.  Tried it in the dummy app and seemed to work as expected.

### DIFF
--- a/addon/components/base/bs-nav.js
+++ b/addon/components/base/bs-nav.js
@@ -29,7 +29,7 @@ import layout from 'ember-bootstrap/templates/components/bs-nav';
  ### Nav styles
 
  The component supports the default bootstrap nav styling options "pills" and "tabs" through the `type`
- property, as well as the `justified` and `stacked` properties.
+ property, as well as the `justified`, `fill` and `stacked` properties.
 
  ### Active items
 
@@ -65,6 +65,8 @@ import layout from 'ember-bootstrap/templates/components/bs-nav';
  You can override `tagName` if you want to use Bootstrap 4's ability to represent more structural
  components with `div` tags.
 
+ The `fill` styling is only available with Bootstrap 4
+
  @class Nav
  @namespace Components
  @extends Ember.Component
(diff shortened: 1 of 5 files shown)


addon/components/base/bs-nav.js
```diff
@@ -65,6 +65,8 @@ import layout from 'ember-bootstrap/templates/components/bs-nav';
  You can override `tagName` if you want to use Bootstrap 4's ability to represent more structural
  components with `div` tags.
 
+ The `fill` styling is only available with Bootstrap 4
```
simonihmig: When this is the case, it would be better to have the changes here be applied rather to the BS4-specific component (https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs4/bs-nav.js).

The setup might be a bit confusing at first, and I definitely should write a proper contributing guide. The idea is to have all common code in the `base` component, and alls version-specific modifications in the respective `bs3` or `bs4` components (that extend from base)

tests/integration/components/bs-nav-test.js
```diff
@@ -23,22 +23,24 @@ module('Integration | Component | bs-nav', function(hooks) {
   testBS3('it supports bootstrap options', async function(assert) {
     // Template block usage:
     await render(hbs`
-      {{bs-nav justified=true stacked=true type="pills"}}
+      {{bs-nav justified=true stacked=true fill=true type="pills"}}
     `);
 
     assert.dom('ul').hasClass('nav-pills', 'has pills class');
     assert.dom('ul').hasClass('nav-justified', 'has justified class');
+    assert.dom('ul').hasClass('nav-fill', 'has fill class');
```
simonihmig: As this property is not supported in BS3, we shouldn't add the tests here, as that would suggest it is supposed to work. Just keep the changes in the `testBS4` test...

addon/components/base/bs-nav.js
dgavey: Great thanks for the tips.  I think it should be set properly now.

PR #685, opened 2018-10-26
feat - allow overriding various yielded components for dropdown & navbar

lougreenwood:
The `bs-nav` component allows overriding yielded components.

This PR adds the same functionality to bs-dropdown & bs-navbar.

(Sidenote - should these be marked as `@private` - these are actually a very useful part of the public API when using `ember-bootstrap` with something like `ember-css-modules`)

### DIFF
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -307,5 +307,26 @@ export default Component.extend({
       && ((menuElement && !menuElement.contains(target)) || this.get('closeOnMenuClick'))) {
       this.send('closeDropdown');
     }
-  }
+  },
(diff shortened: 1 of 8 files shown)


simonihmig:
Thank you for working on this!

> Sidenote - should these be marked as @private - these are actually a very useful part of the public API when using ember-bootstrap with something like ember-css-modules

Valid question. I think it's ok for now to keep them private, as it is the case for the other components. But maybe all of them should eventually be made public. If you have the time, could you open a "RFC" issue suggesting that change, and provide your arguments why this might be needed (e.g. an example of ember-css-modules)? We can then discuss over there...

Issue #673, opened 2018-10-02
Button: support outline style


simonihmig:
See http://getbootstrap.com/docs/4.1/components/buttons/#outline-buttons

An `outline` property as in `{{#bs-button type="primary" outline=true}}` could generate a `btn-outline-primary` class (instead of `btn-primary`).

nickschot:
I’ve already been using this by passing ‘type=“outline-primary”’ and the like. Do you feel the Boolean is a better alternative?

simonihmig:
Hm, right, this way it works already. Or rather *happens* to work, although the docs don't mention this as a supported way of doing it. This would at least require some better documentation.

OTOH the boolean approach would be a bit more decoupled from the actual naming of the classes. Say for example the classes in BS5 would be renamed to `.btn-primary .btn-outline` (instead of `.btn-outline-primary`), it would be possible to adapt without changing the components interface (having some private QPs that generate the appropriate classes).

dgavey:
I would like to create a PR for this one as well.  Will it be ok to create the outline boolean for the button?  Anything else I should be aware of for this one? Sounds like we still need to support ‘type=“outline-primary”’ at a minimum.

simonihmig:
Closed by #687

PR #687, opened 2018-10-27
Adds outline property to bs-button

dgavey:
I think this should do it, and it leaves `outline-primary` as still valid. I was a little unsure on the docs on if I set that up correct for you.  Also I could add a test for each style of outline button, but that didn't seem necessary. As usual, leave me any suggestions and I'll address them.

This is as per Issue #673

### DIFF
--- a/addon/components/bs4/bs-button.js
+++ b/addon/components/bs4/bs-button.js
@@ -2,7 +2,6 @@ import Button from 'ember-bootstrap/components/base/bs-button';
 
 export default Button.extend({
   type: 'secondary',
-  classNameBindings: ['outline:btn-outline'],
 
   /**
    * Property to create outline buttons (BS4+ only)
--- a/addon/components/bs4/bs-button.js
+++ b/addon/components/bs4/bs-button.js
@@ -1,5 +1,16 @@
 import Button from 'ember-bootstrap/components/base/bs-button';
 
 export default Button.extend({
-  type: 'secondary'
+  type: 'secondary',
+  classNameBindings: ['outline:btn-outline'],
+
+  /**
+   * Property to create outline buttons (BS4+ only)
+   *
+   * @property disabled
+   * @type boolean
+   * @default false
+   * @public
+   */
+  outline: false
 });
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -33,6 +33,7 @@ module('Integration | Component | bs-button', function(hooks) {
 
     assert.dom('button').hasClass('btn', 'button has btn class');
     assert.dom('button').hasClass('btn-primary', 'button has type class');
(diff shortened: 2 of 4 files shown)


dgavey:
Ember 2.16 failed to even build for some reason.  All other tests passed.

addon/components/bs4/bs-button.js
```diff
@@ -1,5 +1,16 @@
 import Button from 'ember-bootstrap/components/base/bs-button';
 
 export default Button.extend({
-  type: 'secondary'
-});
\ No newline at end of file
+  type: 'secondary',
+  classNameBindings: ['outline:btn-outline'],
```
simonihmig: This is setting a `btn-outline` class, right? But is this really needed or used somewhere? The examples [here](http://getbootstrap.com/docs/4.1/components/buttons/#outline-buttons) don't mention this class.

tests/integration/components/bs-button-test.js
```diff
@@ -76,6 +77,11 @@ module('Integration | Component | bs-button', function(hooks) {
     assert.dom('button i').hasClass('fa-check');
   });
 
+  testBS4('button with outline property shows as outline', async function(assert) {
+    await render(hbs`{{bs-button type="primary" outline=true}}`);
+    assert.dom('button').hasClass('btn-outline-primary');
```
simonihmig: Can you add another assertion here that it does *not* have the `btn-primary` class? (I know it's working right, but just to make sure it)

addon/components/bs4/bs-button.js
dgavey: Oops accidentally left in some test code.

tests/integration/components/bs-button-test.js
simonihmig: I meant to add this to the existing test:

```suggestion
    assert.dom('button').hasClass('btn-outline-primary');
    assert.dom('button').doesNotHaveClass('btn-primary');
```

tests/integration/components/bs-button-test.js
```diff
@@ -76,6 +77,16 @@ module('Integration | Component | bs-button', function(hooks) {
...
+  });
+
+  testBS4('button with does not have outline class when outline is null', async function(assert) {
+    await render(hbs`{{bs-button type="primary"}}`);
+    assert.dom('button').doesNotHaveClass('btn-outline-primary');
```
simonihmig: I don't think we need a separate test for this. If you want you can add this assertion to the existing `'button has correct type'` test.

simonihmig:
> Sorry about the delay in addressing the feedback. Got really sick last week.

@dgavey Nothing to excuse yourself for, hope you're doing well again! 

Thanks for your recent contributions here, appreciate that! 👍

Issue #674, opened 2018-10-02
Modal: improve accessibility markup


simonihmig:
See http://getbootstrap.com/docs/4.1/components/modal/#accessibility

collinsnji:
Can you please provide details on what you need improved? That'll be much appreciated

simonihmig:
Sure:
* `aria-labelledby` is currently unused, and should be set like specified in the Bootstrap docs
* `role="document"` is missing for the `.modal-dialog`
* add tests for the above points

valentinaperic:
I can work on this! 👍

simonihmig:
Closed by #688

PR #688, opened 2018-10-27
Adds accesibility tags for dialog modals

dgavey:
This was for issue #674

It got a bit more complicated than expected as the aria-labelledby attribute requires the id of the modal-title.  So had to use some js dom calls to get that as we don't know the ember id until it's rendered and there might be a custom modal-title the user creates in their template.

Let me know if it needs any tweaking.

### DIFF
--- a/addon/components/base/bs-modal/dialog.js
+++ b/addon/components/base/bs-modal/dialog.js
@@ -3,7 +3,7 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { htmlSafe } from '@ember/string';
 import { bind } from '@ember/runloop';
-import { alias } from '@ember/object/computed';
+import { readOnly } from '@ember/object/computed';
 import layout from 'ember-bootstrap/templates/components/bs-modal/dialog';
 
 /**
@@ -21,7 +21,7 @@ export default Component.extend({
   attributeBindings: ['tabindex', 'style', 'aria-labelledby'],
   ariaRole: 'dialog',
   tabindex: '-1',
-  "aria-labelledby": alias('titleId'),
+  "aria-labelledby": readOnly('titleId'),
 
   /**
    * Set to false to disable fade animations.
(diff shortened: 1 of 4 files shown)


addon/components/base/bs-modal/dialog.js
```diff
@@ -134,6 +136,44 @@ export default Component.extend({
...
+
+/**
+   * Gets or sets the id of the title element for aria accessibility tags
+   *
+   * @method getSetTitleID
+   * @private
+   */
+  getOrSetTitleId() {
+      //Title element may be set by user so we have to try and find it to set the id
+      const modalNode = document.getElementById(this.get('id'));
```
simonihmig: Isn't this the same as `this.get('element')`? Would be simples than searching the DOM...

addon/components/base/bs-modal/dialog.js
```diff
@@ -17,9 +18,10 @@ export default Component.extend({
   layout,
   classNames: ['modal'],
   classNameBindings: ['fade'],
-  attributeBindings: ['tabindex', 'style'],
+  attributeBindings: ['tabindex', 'style', 'aria-labelledby'],
   ariaRole: 'dialog',
   tabindex: '-1',
+  "aria-labelledby": alias('titleId'),
```
simonihmig: I tend to not use `alias` when possible, as it introduces a two-way binding. Would prefer `readOnly` here.

addon/components/base/bs-modal/dialog.js
```diff
@@ -134,6 +136,44 @@ export default Component.extend({
...
+   * @method getSetTitleID
+   * @private
+   */
+  getOrSetTitleId() {
+      //Title element may be set by user so we have to try and find it to set the id
+      const modalNode = document.getElementById(this.get('id'));
+      let nodeId = null;
+
+      if (modalNode) {
+        const titleNode = modalNode.getElementsByClassName('modal-title');
```
simonihmig: Would prefer to use `modalNode.querySelector('.modal-title')` here. This already returns a single DOM node, so no need for `[0]`.

Issue #693, opened 2018-11-02
Modal header should yield title and close components


lougreenwood:
It looks like the `bs-modal/header` components don't yield out the `bs-modal/header/title` & `...close`  components.

Whilst this seems good enough, it breaks from the usual pattern of child components being yielded out (and AFAIK will also not work with MU because it requires the full path to the nested components instead of the yielded version).

https://github.com/kaliber5/ember-bootstrap/blob/master/addon/templates/components/bs4/bs-modal/header.hbs

simonihmig:
When customizing the header myself, adding a `h4.modal-title` was simple enough, so I didn't feel a need for a yielded component. But I agree, this could be added. 

Up for a PR? 😉

lougreenwood:
I may be able to take a look at a PR :)

simonihmig:
Closed by #701

Issue #648, opened 2018-08-24
Sass import fails in Module Unification app


simonihmig:
Created separate issue for this, as reported by @dguettler [here](https://github.com/kaliber5/ember-bootstrap/issues/381#issuecomment-403035054):

> import statement fails to locate the ember-bootstrap/bootstrap file when included in app.scss

```
Sass Syntax Error (SassCompiler) in src/ui/styles/app.scss:2:

Error: File to import not found or unreadable: ember-bootstrap/bootstrap.
        on line 2 of tmp/sass_compiler-input_base_path-TvLIefue.tmp/src/ui/styles/app.scss
>> @import “ember-bootstrap/bootstrap”;
```

simonihmig:
This PR 👆 fixes the issue in ember-cli!

simonihmig:
This is finally fixed in Ember CLI by https://github.com/ember-cli/ember-cli/pull/8289!

PR #698, opened 2018-11-06
Improve yielded components

lougreenwood:
Follows on from #685

This PR allows overriding the remaining `ember-bootstrap` components.

This allows a user to extend any `ember-bootstrap` component and override the default `bs-*` components yielded out with their own extended implementations.

This PR omits work on `bs-modal-simple`.

A side effect of this PR is that `ember-bootstrap` should now "theoretically" be compatible with the upcoming MU structure where file-system nested components are not allowed :)


simonihmig:
Thank you! I'll merge a bit later, wanted to figure something different out before...

> A side effect of this PR is that ember-bootstrap should now "theoretically" be compatible with the upcoming MU structure where file-system nested components are not allowed :)

What makes you think that? Or better, what makes you think that the current version does not support MU? 

I have been using this addon in a MU app, and works fine so far (except for #648): I think there *could* be an issue with MU if you directly accessed nested components. But here you only directly use root-level components, all the nested ones are yielded (like "private" components, which are supported by MU). Also not sure if that is true when you use the fallback resolver, which is needed AFAIK to support "classic" addons...

lougreenwood:
I suppose you're right, there's actually 2 different PRs to remove any references to nested components (rather than this PR):

* `bs-modal-simple.hbs` - accesses `bs-modal/*` components
* modal header components (#693 - but I have an incoming PR for that anyway)

simonihmig:
Ok, merged!

lougreenwood:
thanks!

PR #701, opened 2018-11-08
Modal header component yields title and close components

lougreenwood:
Addresses #693 

This likely will be a breaking change, for example, when using `bs-modal/header` as a block component, the yielded `close` component must now explicitly be inserted:

Old:
```
{{#modal.header as |header|}}
    {{#header.title}}
        Custom Dialog
        <div class="badge">2</div>
    {{/header.title}}
{{/modal.header}}
```

New:
```
{{#modal.header as |header|}}
    {{#header.title}}
        Custom Dialog
        <div class="badge">2</div>
    {{/header.title}}

    {{header.close}}
{{/modal.header}}
```

This is because `components/base/bs-modal/header.js` has the attribute `closeButton: true` - however, this seems to conflict with the intended composability that yielded components in a block offers - so I removed the following from the block version of the header templates:

```
{{#if closeButton}}
    {{component closeComponent onClick=(action onClose)}}
{{/if}}
```

Because this appears to be a breaking a change, this PR is currently WIP - depending on your comments and how we decide to proceed, the next step will be for me to either refactor or update the docs.


addon/templates/components/bs3/bs-modal/header.hbs
```diff
@@ -1,8 +1,14 @@
-{{#if closeButton}}
-  {{component closeComponent onClick=(action onClose)}}
-{{/if}}
 {{#if hasBlock}}
```
simonihmig: I wouldn't want to make this a breaking change.

What about this: 
* besides the special template keyword `hasBlock` there is also `has BlockParams`
* as previously we didn't yield any components, this usage can be covered with the `else` part of `{{#if hasBlockParams}}`
* the new `{{yield (hash ...)}}` moves into the if block of `{{#if hasBlockParams}}`

If I am not missing something, this way we could support the yielded components (without a close button by default), while still remaining backwards compatible.

What do you think?

tests/integration/components/bs-modal/header-test.js
```diff
@@ -16,7 +16,7 @@ module('Integration | Component | bs-modal/header', function(hooks) {
   });
 
   test('Header can have custom block content', async function(assert) {
-    await render(hbs`{{#bs-modal/header}}<div id="custom">Test</div>{{/bs-modal/header}}`);
+    await render(hbs`{{#bs-modal/header as |header|}}<div id="custom">Test</div>{{header.close}}{{/bs-modal/header}}`);
```
simonihmig: When following the approach I suggested above, the existing test should remain as-is, and a new test for the yielded components case should be added.

lougreenwood:
@simonihmig Sorry this has taken so long, but I've made the requested changes and added some extra tests.

I think I've implemented it as you described and it seems this is no longer a breaking change, but pls carefully check :)

lougreenwood:
@simonihmig could I get a review on this pls? :)

simonihmig:
@lougreenwood Sorry, I somehow missed this! Thanks for keeping at it! There are some merge conflicts, could you please rebase this and ping me afterwards, I will then review quickly!

lougreenwood:
Hey @simonihmig - sorry for the delay.

Merge conflicts fixed 🎉

PR #702, opened 2018-11-08
Don't wrap checkboxes in labels [BS4]

### LINKED ISSUE #699: Difficult to override checkbox styling
The current `bs-form` HTML structure for form label & elements makes it difficult to override checkbox styling.

See: https://medium.com/claritydesignsystem/pure-css-accessible-checkboxes-and-radios-buttons-54063e759bb3

Currently the label element wraps the form element, instead, label & input element should be siblings (making use of the `for` attribute).

I'm happy to attempt a PR for this at some point in the near future.

lougreenwood:
Fixes #699 

Possibly a breaking change since CSS dependent on the previous markup will now be broken.

### DIFF
--- a/addon/templates/components/bs3/bs-form/element/layout/horizontal/checkbox.hbs
+++ b/addon/templates/components/bs3/bs-form/element/layout/horizontal/checkbox.hbs
@@ -1,7 +1,8 @@
 <div class="{{horizontalInputGridClass}} {{horizontalInputOffsetGridClass}}">
   <div class="checkbox">
-    {{yield}}
-    {{component labelComponent}}
+    {{#component labelComponent}}
+      {{yield}}
+    {{/component}}
   </div>
   {{component errorsComponent}}
   {{component helpTextComponent}}


addon/templates/components/bs3/bs-form/element/layout/horizontal/checkbox.hbs
```diff
@@ -1,8 +1,7 @@
 <div class="{{horizontalInputGridClass}} {{horizontalInputOffsetGridClass}}">
   <div class="checkbox">
-    {{#component labelComponent}}
-      {{yield}}
-    {{/component}}
+    {{yield}}
+    {{component labelComponent}}
```
simonihmig: Hm, I am not sure we should do this change also for BS3!? 
The old docs suggest to wrap the checkbox in a label there: https://getbootstrap.com/docs/3.3/css/#checkboxes-and-radios

As the suggested markup is different for BS4 (unwrapped), we could see a BS4-only change as a bug fix, thus (at least technically) not having to declare this a breaking change! 😉

PR #706, opened 2018-11-13
fix(bs-dropdown): Clicking outside should close the menu when closeOnMenuClick is false

offirgolan:
Currently, when `renderInPlace=false` and `closeOnMenuClick=false`, the menu doesnt close when clicking outside of it since the `menuElement` is outside the scope of the bs-dropdown component.

### DIFF
--- a/tests/integration/components/bs-dropdown-test.js
+++ b/tests/integration/components/bs-dropdown-test.js
@@ -149,7 +149,9 @@ module('Integration | Component | bs-dropdown', function(hooks) {
 
   test('clicking outside dropdown menu when closeOnMenuClick is false and renderInPlace is false will close it', async function(assert) {
     await render(
-      hbs`{{#bs-dropdown closeOnMenuClick=false as |dd|}}{{#dd.toggle}}Dropdown <span class="caret"></span>{{/dd.toggle}}{{#dd.menu renderInPlace=false}}<li><a href="#">Something</a></li>{{/dd.menu}}{{/bs-dropdown}}`
+      hbs`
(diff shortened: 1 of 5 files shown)


tests/integration/components/bs-dropdown-test.js
```diff
@@ -135,6 +135,30 @@ module('Integration | Component | bs-dropdown', function(hooks) {
...
+    assert.dom('.dropdown-menu').exists();
+    assert.dom(dropdownVisibilityElementSelector()).hasClass(openClass(), 'Dropdown is open');
+
+    await click(document.body);
+    assert.dom('.dropdown-menu').doesNotExist();
+  });
+
+  test('clicking outside dropdown menu when closeOnMenuClick is false and renderInPlace is false will close it', async function(assert) {
+    await render(
+      hbs`{{#bs-dropdown closeOnMenuClick=false as |dd|}}{{#dd.toggle}}Dropdown <span class="caret"></span>{{/dd.toggle}}{{#dd.menu renderInPlace=false}}<li><a href="#">Something</a></li>{{/dd.menu}}{{/bs-dropdown}}`
```
simonihmig: You will need to add the wormhole container to the test template here, otherwise the `renderInPlace=false` will be ineffective. Like this: https://github.com/kaliber5/ember-bootstrap/blob/c49d1ed6219d267caf29ec1d360759805d14fa33/tests/integration/components/bs-dropdown-test.js#L273

offirgolan: Done 😄

PR #723, opened 2018-12-14
Feat: Enables setting supplied classes on dropdown/ember-popper

lougreenwood:
Allows passing an array of `menuClassNames` to a `bs-dropdown` instance and have these propagate to the `bs-dropdown/menu` component.


lougreenwood:
I'm going to close this, I found a cleaner way to do what I was attempting (set classes on the menu based on some computed prop on the `bs-dropdown` component.

Issue #716, opened 2018-12-06
Disable validations on form submission


Techn1x:
It seems at the moment there isn't an official/public way to disable validations for a form submission.

I have a form which 99% of the time validations should be run, but in one case it should submit no matter what.

I am using ember-bootstrap-cp-validations, and with that I have 2 options;
* Use the global options hash to define a `disabled` property, which is then passed to all validators. This property will be overridden if an individual validator defines its own `disabled` property
* Define a `disabled` property on every validator (tedious for large forms..)

Ideally something like... `{{#bs-form ... disableValidations=model.dontValidate}}` would be great, especially since an approach like that would work no matter which validation library was being used!

Techn1x:
Thanks @jelhan , I was after option 2, I didn't even realise an `onInvalid()` method existed, and that will definitely solve my problem!

jelhan:
@Techn1x Just wondering if we should keep this open as quest issue cause even your use case is not fully supported yet.

jelhan:
@simonihmig Any thoughts on third case?

> 3. Disable the validations at all.
>
> If we support first and second case third one would only be about performance, isn't it?

I'm tending to not add any special support cause it would add a lot of complexity. Disabling validation should be handled by validation library IMO. If you agree this issue could be closed cause last remaining task is tracked at #364.

simonihmig:
If I understand this correctly, the third case was about disabling the validations itself, not just not-displaying them, right? For this I would definitely agree, this seems like a bad thing to do because of the far-reaching consequences (interference with application state). This library should be only concerned with the UI-part.

Closing this for now.

Techn1x:
Thanks for the hard work everyone!

PR #730, opened 2018-12-21
Form submission state reflects Promise returned by onInvalid action

jelhan:
Assuming we have a form that binds same action to onInvalid and onSubmit events:
`{{bs-form onSubmit=(action "save") onInvalid=(action "save")}}`

In that case only difference caused by having validation errors should be that they are shown to user. It should not be any difference in submission state. Especially yielded `isSubmitting` property of `{{bs-form}}` should be `true` as long as promise returned by `save` action is pending regardless of validation.

This problem was raised in #716.

### DIFF
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -300,27 +300,23 @@ export default Component.extend({
     this.get('onBefore')(model);
 
     RSVP.resolve(this.get('hasValidator') ? this.validate(this.get('model')) : null)
-      .then((r) => {
-        RSVP.resolve(this.get('onSubmit')(model, r))
-          .finally(() => {
-            if (!this.get('isDestroyed')) {
-              if (this.get('pendingSubmissions') === 1) {
-                this.set('isSubmitting', false);
-              }
-              this.decrementProperty('pendingSubmissions');
-            }
-          });
-      })
-      .catch((err) => {
-        if (!this.get('isDestroyed')) {
+      .then(
+        (r) => {
+          return this.get('onSubmit')(model, r);
+        },
+        (err) => {
           this.set('showAllValidations', true);
 
+          return this.get('onInvalid')(model, err);
+        }
+      )
+      .finally(() => {
+        if (!this.get('isDestroyed')) {
           if (this.get('pendingSubmissions') === 1) {
             this.set('isSubmitting', false);
           }
-          this.decrementProperty('pendingSubmissions');
 
-          this.get('onInvalid')(model, err);
+          this.decrementProperty('pendingSubmissions');
         }
       });
   },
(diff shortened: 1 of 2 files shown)


addon/components/base/bs-form.js
```diff
@@ -300,27 +300,23 @@ export default Component.extend({
...
-      .catch((err) => {
-        if (!this.get('isDestroyed')) {
+      .then(
+        (r) => {
+          return this.get('onSubmit')(model, r);
+        },
+        (err) => {
           this.set('showAllValidations', true);
 
+          return this.get('onInvalid')(model, err);
```
jelhan: Setting `showAllValidations` to `true` before `onInvalid` fulfills is only for not introducing a breaking change. IMO it would be better to now show validation errors before onInvalid resolves. Maybe value onInvalid resolves with should control if validation errors will be should be shown at all? Let me know if you would except such a breaking change.

simonihmig: > IMO it would be better to now show validation errors before onInvalid resolves. Maybe value onInvalid resolves with should control if validation errors will be should be shown at all?

Makes sense!

> Let me know if you would except such a breaking change.

Yeah, not sure about it. I *guess* it should not make any problems in practice, as currently the result of `onInvalid` is not able to control the validations, so showing the validations after the promise has resolved would only cause a slight timing change I guess, nothing which really breaks things from the user's perspective. It could maybe cause some failing tests though!?

An alternative would be to create a tracking issue for this, and implement it for the next major version bump. Which might be happening rather soon maybe, if e-b drops support for Ember CLI < 2.12 (to be able to upgrade to Babel 7). 

*Tbh, the semver "contract" regarding Ember CLI is quite fuzzy (a problem with all addons), as it nowhere states which versions it supports. But as e-b support Ember 2.3, I would think it would be reasonable to expect the same for CLI!? Hm, getting a bit off-topic...* 😝

Thoughts?

jelhan: > It could maybe cause some failing tests though!?

I think this would be only the case in very few rare cases, cause mostly cases await all promises to be settled after triggering a user interaction (e.g. `await click('button[type="submit"]');`). In that cases `onInvalid` will be resolved anyways before assertion is done. This change might only fail tests if they await an interim state using `waitUntil` or `waitFor`. In my experience that's not used quite often yet.

> An alternative would be to create a tracking issue for this, and implement it for the next major version bump. 
It's your call. Could create another pull request with necessary changes after this one has landed and you are free to merge it at any time.

> Tbh, the semver "contract" regarding Ember CLI is quite fuzzy (a problem with all addons), as it nowhere states which versions it supports. But as e-b support Ember 2.3, I would think it would be reasonable to expect the same for CLI!? Hm, getting a bit off-topic... stuck_out_tongue_closed_eyes

IMO an ember version should be considered as a specific bundle of `ember-source`, `ember-cli` and `ember-data`. Furthermore I would argue that supporting a specific version should mean that it works fine with default blueprint for that ember version. You could always extend that general statement by saying that it also supports older versions of `ember-source`.

simonihmig:
Thanks @jelhan. Will create an issue to track the topic we discussed here...

PR #732, opened 2018-12-28
Form yields isSubmitted and isRejected properties

jelhan:
Belongs to #638

### DIFF
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -84,7 +84,7 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
   {{/bs-form}}
   ```
 
-  Additionaly `isSubmitted` and `isRejected` proprties are yielded. `isSubmitted` is `true` if last submission was successful.
+  Additionaly `isSubmitted` and `isRejected` properties are yielded. `isSubmitted` is `true` if last submission was successful.
   `isRejected` is `true` if last submission was rejected due to validation errors or by action bound to `onSubmit` event.
   Both are reset as soon as any value of a form element changes. It could be used for visual feedback about last submission:
 
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -77,9 +77,21 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
 
   ```
   {{#bs-form onSubmit=(action 'save') as |form|}}
-    {{#bs-button type='submit' disabled=form.isSubmitting}}
+    {{#bs-button buttonType='submit' disabled=form.isSubmitting}}
       Save
       {{#if form.isSubmitting}} {{fa-icon 'spinner'}} {{/if}}
+    {{/bs-button}}
+  {{/bs-form}}
+  ```
+
+  Additionaly `isSubmitted` and `isRejected` proprties are yielded. `isSubmitted` is `true` if last submission was successful.
+  `isRejected` is `true` if last submission was rejected due to validation errors or by action bound to `onSubmit` event.
+  Both are reset as soon as any value of a form element changes. It could be used for visual feedback about last submission:
+
+  ```
+  {{#bs-form onSubmit=(action 'save') as |form|}}
+    {{#bs-button buttonType='submit' type=(if form.isRejected "danger" "primary")}}
+      Save
     {{/bs-button}}
   {{/bs-form}}
   ```
@@ -176,6 +188,33 @@ export default Component.extend({
    * @private
    */
   isSubmitting: false,
+
+  /**
+   * `isSubmitted` is `true` if last submission was successful.
+   * A change to any form element resets it's value to `false`.
+   *
+   * If not using `Components.FormElement`, `resetSubmissionState` action must be triggered on each change to reset
+   * form's submission state.
+   *
+   * @property isSubmitted
+   * @type {Boolean}
+   * @private
+   */
+  isSubmitted: false,
+
+  /**
+   * `isRejected` is `true` if last submission was rejected.
+   * A submission is considered as rejected if form is invalid as well as if `onSubmit` rejects.
+   * A change to any form element resets it's value to `false`.
+   *
+   * If not using `Components.FormElement`, `resetSubmissionState` action must be triggered on each change to reset
+   * form's submission state.
+   *
+   * @property isRejected
+   * @type {Boolean}
+   * @private
+   */
+  isRejected: false,
 
   /**
    * Count of pending submissions.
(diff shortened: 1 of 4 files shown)


addon/components/base/bs-form/element.js
```diff
@@ -928,6 +928,14 @@ export default FormGroup.extend({
    */
   onChange() {},
 
+  /**
+   * Private duplicate of onChange event used for internal state handling between form and it's elements.
+   *
+   * @event _onChange
+   * @private
+   */
+  _onChange() {},
```
jelhan: This feels strange but unless I missed something there isn't any other way than having a duplicated `onChange` event for this internal sate tracking.

tests/integration/components/bs-form-test.js
```diff
@@ -322,6 +322,78 @@ module('Integration | Component | bs-form', function(hooks) {
     assert.dom('form .state').doesNotHaveClass('is-submitting');
   });
 
+  test('Yields #isSubmitted', async function(assert) {
+    this.actions.submit = this.stub().resolves();
```
jelhan: Wasn't sure if you prefer to use sinon to mock actions generally (#686). Could also quickly refactor to use `RSVP.resolve()` since we don't assert that the action was executed at all. IMO relying on another dependency (sinon) is just adding complexity here. But maybe you prefer it to have more similar test setups?

addon/components/base/bs-form.js
```diff
@@ -177,6 +189,33 @@ export default Component.extend({
...
+   *
+   * @property isSubmitted
+   * @type {Boolean}
+   * @private
+   */
+  isSubmitted: false,
+
+  /**
+   * `isRejected` is `true` if last submission was rejected.
+   * A submission is considered as rejected if form is invalid as well as if `onSubmit` rejects.
```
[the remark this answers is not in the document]
jelhan: > Maybe isRejected should not be true if an onInvalid method is passed in and that one resolves?

As more as I think about it I'm tending to prefer that implementation. Assuming a form like `<BsForm @onSubmit={{action "save"}} @onInvalid={{action "save"}}>` it should not make any difference if the model is valid or not.

Will wait for your feedback before updating PR accordingly.

tests/integration/components/bs-form-test.js
simonihmig: That's fine.

addon/components/base/bs-form/element.js
simonihmig: We could call `resetSubmissionState()` in the form's `change` action, and only have a single action called here. But that would break the state handling if the user uses a custom `onChange` handler on the form element. Not sure if we need to support that case, as the user is then handling the form state by her/himself, so she/he could (should?) be responsible for handling the form submission state as well? 

A bot undecided about this one. We could maybe try the latter first, to simplify the implementation, and see if anyone sees this as an issue?

addon/components/base/bs-form.js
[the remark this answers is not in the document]
simonihmig: > Maybe isRejected should not be true if an onInvalid method is passed in and that one resolves?

Agree. 

IMO `isSubmitted` and `isRejected` should only be influenced by the `onSubmit()` action. The result of `onInvalid()` is not used so far (as the submission state is basically something like `isInvalid`, whatever the result of `onInvalid()` is). Are we on the same page?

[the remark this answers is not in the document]
jelhan: If I got you right you are suggesting to add an additional submission state `isInvalid` to handle that case? Lets maybe put all cases together to not get confused.

Table only lists the property which is `true`. All other ones should be `false`.

| case                                                          | version a        | version b        | version c        |
| ------------------------------------------------------ | ------------------ | ------------------ | ------------------ | 
| form is valid, `onSubmit` resolves             | `isSubmitted` | `isSubmitted` | `isSubmitted` |
| form is valid, `onSubmit` rejects                | `isRejected`  | `isRejected`   | `isRejected `   |
| form is invalid, `onInvalid` is `undefined`  | `isRejected`   | `isRejected`   | `isInvalid`       |
| form is invalid, `onInvalid` resolves           | `isRejected`   | `isSubmitted` | `isInvalid`       |
| form is invalid, `onInvalid` rejects              | `isRejected`   | `isRejected`   | `isInvalid`       |

_version a_ is current implementation in this PR. _version b_ is what I had in mind in my first comment. _version c_ is what you are suggesting if I got you right.

A _version d_ could combine _version b_ and _version c_ meaning that `isInvalid` just reflects validation state on submission and `isRejected` / `isSubmitted` indicates if event handler resolves.

A strong argument for _version b_ is in my opinion that it would allow to ignore validation errors on form submission without introducing any edge cases. That was the main reason from #730.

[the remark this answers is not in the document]
simonihmig: > If I got you right you are suggesting to add an additional submission state isInvalid to handle that case?

Possibly, but not really required I guess, as one could as well just use the validation library's appropriate property!?

> Assuming a form like <BsForm @onSubmit={{action "save"}} @onInvalid={{action "save"}}> it should not make any difference if the model is valid or not.

I assume if `onSubmit` does not return a Promise, this will still make `isSubmitted` true, right? In this case there would still be a difference between `onSubmit` and `onInvalid`, even in your preferred version b. Maybe you can add a table row "form is valid, onSubmit is `undefined`"?

I am still a bit hesitant for this `onInvalid` case. The case is #730 is still quite unusual IMHO. And not sure if the support for this will not cause unexpected behavior for other cases, where a user's `onInvalid` action might (accidentally) return a promise (say you do a `(perform task)`, with some async things handled for the invalid case, like logging, flash messages etc.), but the semantic of that promise is not really that this should be considered a successful form submission (which might cause some inappropriate rendering)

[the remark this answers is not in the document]
jelhan: > And not sure if the support for this will not cause unexpected behavior for other cases, where a user's onInvalid action might (accidentally) return a promise [...], but the semantic of that promise is not really that this should be considered a successful form submission

That's a very good argument, which I haven't considered yet. Version b requires `onInvalid` method to be absent or to reject / throw in order to get an `isRejected` state. I think you are totally right that no one would expect such a behavior when implementing an action that handles a submission of invalid form.

jelhan:
Lets come back to this one. If I got our discussion right, we agreed that we define `isSubmitted` and `isRejected` as it was originally implemented:

> `isSubmitted` is `true` if last submission was successful.
> `isRejected` is `true` if last submission was rejected due to validation errors or by action bound to `onSubmit` event.

Tests are now covering these scenarios:
> Yielded #isSubmitted is true if onSubmit resolves
> Yielded #isSubmitted is true if onSubmit is undefined
> Yielded #isSubmitted is true if validation passes
> Yielded #isRejected is true if onSubmit action rejects
> Yielded #isRejected is true if validation fails

Not related to the discussion above but they also cover:
> A change to a form elements resets yielded #isSubmitted
> A change to a form elements resets yielded #isRejected
> Triggering resetSubmissionState resets submission state of form

This is ready to get merged from my side.

addon/components/base/bs-form.js
```diff
@@ -77,13 +77,25 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
...
   {{#bs-form onSubmit=(action 'save') as |form|}}
-    {{#bs-button type='submit' disabled=form.isSubmitting}}
+    {{#bs-button buttonType='submit' disabled=form.isSubmitting}}
       Save
       {{#if form.isSubmitting}} {{fa-icon 'spinner'}} {{/if}}
     {{/bs-button}}
   {{/bs-form}}
   ```
 
+  Additionaly `isSubmitted` and `isRejected` proprties are yielded. `isSubmitted` is `true` if last submission was successful.
```
simonihmig:
```suggestion
  Additionaly `isSubmitted` and `isRejected` properties are yielded. `isSubmitted` is `true` if last submission was successful.
```

addon/components/base/bs-form.js
```diff
@@ -77,13 +77,25 @@ import layout from 'ember-bootstrap/templates/components/bs-form';
...
+  `isRejected` is `true` if last submission was rejected due to validation errors or by action bound to `onSubmit` event.
```
simonihmig:
```suggestion
  `isRejected` is `true` if last submission was rejected due to validation errors or by an action bound to `onSubmit` event, returning a rejected promise.
```

jelhan:
Thanks for fixing that spelling errors. Accepted both changes. Do you like me to squash?

PR #738, opened 2019-01-06
Modal should not violate CSP style-src 'none'

jelhan:
Binding style attribute causes a Content-Security-Policy violation unless `style-src` allows `'unsafe-inline'`. To avoid this CSS Object Model (CSSOM) must be used.

CSP violation is triggered by glimmer executing `element.setAttribute('style', value)`.

This has been discussed here without coming up with a better solution: https://discuss.emberjs.com/t/binding-style-attribute-and-content-security-policy-best-practice/10921

### DIFF
--- a/addon/components/base/bs-modal/dialog.js
+++ b/addon/components/base/bs-modal/dialog.js
@@ -1,7 +1,6 @@
 import { isBlank } from '@ember/utils';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import { htmlSafe } from '@ember/string';
 import { bind } from '@ember/runloop';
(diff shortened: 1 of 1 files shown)


addon/components/base/bs-modal/dialog.js
```diff
@@ -201,6 +194,11 @@ export default Component.extend({
     // directly to the element
     this.element.onclick = bind(this, this._click);
     this.getOrSetTitleId();
+    this.updateStyles();
```
jelhan: Could not use `didReceiveAttrs` cause that one is executed before element is insert into DOM.

PR #741, opened 2019-01-10
Add option to hide all validations after form submission

simonihmig:
After a successful form submission (i.e. no validation errors) which changes some state (e.g. creates a record), this could make the model's validation invalid (think of an `exclude` validation that a property is unique when creating an model, which becomes invalid after the form has been submitted and the new name has been added to a live updating collection). Previously if the form remains visible this would show a new validation error, after a successful form submission, because the form elements `showOwnValidation` has been activated before!

With this change, the display of validations is resetted after a successful submission, as if a fresh form component was rendered, so the user would have to resubmit or trigger the element's validation display (focusout) again to show the validation error that has been added meanwhile.

### DIFF
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -238,10 +238,10 @@ export default Component.extend({
   /**
    * @property showAllValidations
    * @type boolean
-   * @default false
+   * @default undefined
    * @private
    */
-  showAllValidations: false,
+  showAllValidations: undefined,
 
   /**
    * Action is called before the form is validated (if possible) and submitted.
@@ -302,6 +302,7 @@ export default Component.extend({
     RSVP.resolve(this.get('hasValidator') ? this.validate(this.get('model')) : null)
       .then(
         (r) => {
+          this.set('showAllValidations', false);
           return this.get('onSubmit')(model, r);
         },
         (err) => {
@@ -317,6 +318,11 @@ export default Component.extend({
           }
 
           this.decrementProperty('pendingSubmissions');
+
+          // reset forced hiding of validations
+          if (this.get('showAllValidations') === false) {
+            this.set('showAllValidations', undefined);
+          }
         }
       });
   },
--- a/addon/components/base/bs-form.js
+++ b/addon/components/base/bs-form.js
@@ -211,6 +211,18 @@ export default Component.extend({
   preventConcurrency: false,
 
   /**
+   * If true, after successful validation and upon submitting the form, all current element validations will be hidden.
+   * If the form remains visible, the user would have to focus out of elements of submit the form again for the
+   * validations to show up again, as if a fresh new form component had been rendered.
+   *
+   * @property hideValidationsOnSubmit
+   * @type {Boolean}
+   * @default false
+   * @public
+   */
+  hideValidationsOnSubmit: false,
+
+  /**
    * If set to true novalidate attribute is present on form element
    *
    * @property novalidate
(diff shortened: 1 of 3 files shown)


simonihmig:
@jelhan as you have been working on e-b forms a lot as well, I would love your eyes here!

This changes the behavior a bit, and as such *could* be regarded a breaking change. Also I believe this could rather be seen as a bug fix, as the previous behavior did not make too much sense. And as this is rather edge-case-y it will hopefully not affect anyone/many.

simonihmig:
Seems CI is failing for unrelated issues... 🙄

jelhan:
I'm not 100%ly sure if I understand your use case. A unique validator should only consider an input as invalid if another record has the same value isn't it? So this shouldn't be an issue for an edit form. Are we dealing with an input form that doesn't trigger a redirect neither resets it's values after successful submission?

Maybe I missed something but that sounds like a very special edge case. On the other hand I see a way more common case for which this might not be the expected behavior: imagine an edit form that saves the data on submission but doesn't trigger a transition. Wouldn't a reset of validation state be confusing to the user in that case? IMO this is even more important cause there is currently no officially supported way to always show the validation errors (#364).

TL;TR: In my opinion this should be configurable and shouldn't be the default behavior.

jelhan:
> A field should be unique, using `ember-cp-validation`'s exclusion validator, which in turn uses a CP to collect the unique fields of existing models. [...] It does redirect, but shortly before that happens the internal store has already been updated with the newly created record, which updates the exclusion validator and thus makes the (still visible) create(!) form show a validation error for a fraction of a second.

Wouldn't it be an option to ignore new / in-flight records for validation?

> It has a slight side visual side effect though, in that success validations (i.e. inputs with that green border) are removed (normal border) after submission.

This is the case I'm worried about.

> But that seems actually reasonable to me, as the green validation borders are to show the user which fields are filled in correctly before submitting the form.

This seems to be a complex UX decision. I've a use case to show validation status always to user for edit forms. Only create forms should not show validation errors before user interaction to not start with validation errors. This is the motivation for #364. I've setup an Ember Twiddle to demonstrate how this looks in practice: https://ember-twiddle.com/#/dc989bc6bf5dc2360e04b700a7cf3986?openFiles=controllers.add.js%2C&route=%2Fadd This use case would be broken by this change, wouldn't it?

Also I'm worried that hiding validation success while form is submitting might also cause confusion - especially if server response quickly this might be seen as flickering.

I'm currently in a hurry. Will try to put more detailed thoughts together later.

simonihmig:
> In my opinion this should be configurable and shouldn't be the default behavior.

I would be ok with that, if it mitigates your concerns!?

simonihmig:
@jelhan just added a `hideValidationsOnSubmit` property, to make this opt-in. Happy with that?

simonihmig:
>  If I didn't missed something this server-side validation errors will only be shown if user has interacted with that form element before. 

Yes, seems true. 🤔

Wouldn't the solution be to have https://github.com/kaliber5/ember-bootstrap/issues/364 implemented? (the guy who suggested this seemed to think well ahead! 😉)

In the `null` case (see https://github.com/kaliber5/ember-bootstrap/issues/364#issuecomment-309417556) this would work like now, including `hideValidationsOnSubmit=true`, but with the public property being set to `true`, this would take over and always show the validations.

jelhan:
Actually I'm not quite sure how #364 and this one would play together. Also I'm thinking about a more declarative API like `<BsForm @showValidations="never">` with the values `never`, `always`, `after-interaction` (current default), `on-change`. For `after-interaction` and `on-change` controlling the validation state after form-submission would be great. But it would also be possible in user land by changing `showValidations` in `onSubmit` handler.

How urgent is this PR for you? IMO it would be better to start with #364 and then have a look what's missing. I may find some time this weekend to work on it.

simonihmig:
> Actually I'm not quite sure how #364 and this one would play together.

I thought that you could force the display of validations, after assigning the server-side errors (see your changeset example).

> How urgent is this PR for you?

Well, not super urgent, but would like to have a solution in at least 2 weeks or so...

PR #746, opened 2019-01-16
CSP: refactor to not require `style-src 'unsafe-inline` and add test coverage

jelhan:
This PR fixes all violations of a strict Content-Security-Policy and add test coverage to prevent regression. Tests will fail if they trigger a CSP violation.

Fixes for CSP are binding of style attribute which requires `unsafe-inline` cause Glimmer will do a `element.setAttribute('style', value)`. This is true for `Component.extend({ attributeBindings: ['style'], style: '...' })` as well as `<div style="...">` in templates. The last case was often present in tests. For that one a test helper that inserts CSS rules needed for testing is added. Attribute binding was mostly refactored to methods that changes the style using CSSOM and triggered in component lifecycle hooks. There might still be room for optimization.

I'm not convinced by the code needed to dynamically bind style. That doesn't feel correct in an ember environment. But as far as I know there isn't any better way currently. Hopefully there will be as soon as Element Modifiers ([RFC 353](https://github.com/emberjs/rfcs/pull/353)) land and are supported by all targeted version. But that may need a while and would also require a refactor for outer HTML.

### DIFF
--- a/addon/components/base/bs-collapse.js
+++ b/addon/components/base/bs-collapse.js
@@ -1,10 +1,11 @@
 import { not, alias, and } from '@ember/object/computed';
 import Component from '@ember/component';
-import { isPresent, isEmpty } from '@ember/utils';
-import { observer, computed } from '@ember/object';
+import { isPresent } from '@ember/utils';
+import { observer } from '@ember/object';
 import { next } from '@ember/runloop';
-import { htmlSafe, camelize } from '@ember/string';
+import { camelize } from '@ember/string';
 import transitionEnd from 'ember-bootstrap/utils/transition-end';
+import { assert } from '@ember/debug';
 
 /**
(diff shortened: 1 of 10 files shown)


simonihmig:
> would also require a refactor for outer HTML.

You mean they are only useable with Glimmer component? You sure? I didn't read that out of the RFC, but might have missed something. I believe they should work with any component, as well as any template in general.

jelhan:
> @jelhan nothing short of awesome! clap

Thanks.

> You mean they are only useable with Glimmer component?

They should work for all type of components for elements defined in template. But they won't work for the element defined by component. The three cases of dynamic styles addressed by this pull request and #738 are only dealing with elements defined by component. Therefore a refactor for outer HTML would be needed.

simonihmig:
Ah, ok, got it! yes, indeed...

PR #747, opened 2019-01-21
Expand scenarios where BsButton click event will bubble

lougreenwood:
When extending a `BsButton` by composition into some new Button component, it is necessary to pass through the `onClick` arg, eg:

my-composed-button.hbs
```
<BsButton @onClick={{@onClick}}>
  {{yield}}
</BsButton>
```

However, if the passed `{{@onClick}}` is not set when the composed button is initialised, it will not trigger bubbling on the base `BsButton` (because null !== undefined):

```
<MyComposedButton>
  Hello
</MyComposedButton>
```

This PR allows either null or undefined to trigger bubbling by exiting the `click` handler early


simonihmig:
Thanks!

PR #753, opened 2019-02-06
Forms use the set function of the model if it is present

nickschot:
This fixes things like `ember-changeset` which intercepts the `.set()` call to properly propagate updates.


simonihmig:
Thanks @nickschot!

PR #754, opened 2019-02-16
Fix transition for bs-collapse

### LINKED ISSUE #629: No transition for bs-collapse
```
Npm packages:
ember-bootstrap: ^1.2.0 -> 1.2.0
ember-cli: ~2.14.2 -> 2.14.2
bootstrap: ^4.0.0 -> 4.0.0
bootstrap-sass: n/a
ember-cli-sass: ^6.2.0 -> 6.2.0
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapFont: false
importBootstrapCSS: false
```

https://www.ember-bootstrap.com/#/components/collapse

^ On this demo if you "Hide" the collapsed content it has no transition of 350 ms. It just abruptly closes.

Contrast that with Bootstrap 4's demo:

https://getbootstrap.com/docs/4.1/components/collapse/#example

Also I don't understand why there's a configurable called transitionDuration when that doesn't exist in bootstrap itself. It seems like the entire solution is a rewrite in terms of the javascript behavior and the only bootstrap aspect is the CSS itself.

simonihmig:
Fixes #629, introduced by #746.

@jelhan fyi. I tried to add some tests, asserting that the height increases/decreases during the transition. Which kinda worked, but were quite flaky, so skipped them for now.

### DIFF
--- a/addon/components/base/bs-collapse.js
+++ b/addon/components/base/bs-collapse.js
@@ -113,8 +113,8 @@ export default Component.extend({
    */
   transitionDuration: 350,
 
-  setCollapseSize() {
-    let { collapseSize: size, collapseDimension: dimension } = this.getProperties('collapseSize', 'collapseDimension');
+  setCollapseSize(size) {
+    let dimension = this.get('collapseDimension');
 
     assert(`collapseDimension must be either "width" or "height". ${dimension} given.`, ["width", "height"].includes(dimension));
 
@@ -165,9 +165,9 @@ export default Component.extend({
 
     this.setProperties({
       transitioning: true,
-      collapseSize: this.get('collapsedSize'),
       active: true
     });
+    this.setCollapseSize(this.get('collapsedSize'));
 
     transitionEnd(this.get('element'), this.get('transitionDuration')).then(() => {
       if (this.get('isDestroyed')) {


jelhan:
Oh missed that one. Sorry.

> I tried to add some tests, asserting that the height increases/decreases during the transition. Which kinda worked, but were quite flaky, so skipped them for now.

I guess related to hard-coded duration of 0ms in testing? https://github.com/kaliber5/ember-bootstrap/blob/v2.5.0/addon/utils/transition-end.js#L11-L13 That makes asserting interim states difficult, isn't it?

simonihmig:
> I guess related to hard-coded duration of 0ms in testing? https://github.com/kaliber5/ember-bootstrap/blob/v2.5.0/addon/utils/transition-end.js#L11-L13 That makes asserting interim states difficult, isn't it?

Yeah, with that 0ms it obviously didn't work at all. But I played with a function to override this duration in tests, but even when setting this to the original 350ms, it wasn't working reliably. And it wasn't debuggable, as setting any breakpoint during the transition didn't allow to inspect what was going on, as the CSS transition wasn't stopped obviously. It was working like in 80% of the test runs, but failing without an obvious reason in the other ones... 🙄

simonihmig:
> Oh missed that one. Sorry.

No problem at all! 🙂

PR #764, opened 2019-03-03
Drop Saucelabs in favour of Browserstack

### LINKED ISSUE #728: Move to Browserstack for cross browser tests
As Saucelaubs always has issues (timeouts etc.)

simonihmig:
Closes #728


.travis.yml
```diff
@@ -19,8 +19,8 @@ env:
     - AFTER_SCRIPT="true"
     - BEFORE_SCRIPT="true"
     - TEST_COMMAND=""
-    - SAUCE_USERNAME="ember-bootstrap"
-    - SAUCE_ACCESS_KEY="16ef42b7-b940-4a05-bb1e-9e3fef3bf634"
+    - BROWSERSTACK_USERNAME="simonihmig1"
+    - BROWSERSTACK_ACCESS_KEY="9RfKspHGCip5XjBDHWGW"
```
lolmaus: Is it reasonable?

I'd add passwords via CI configuration UI rather than open-source code.

simonihmig: In principle you're right. But I don't know of any better way with Travis. IIRC one issue with Saucelabs, which I used before and works the same in that regard, was that when adding "private" env vars, they are not used for PRs from other users, so tests won't run there.

The only "danger" here is that someone could use the account to run his own tests.

FWIW, ember does the same: https://github.com/emberjs/ember.js/blob/master/.travis.yml#L20-L21

PR #766, opened 2019-03-04
Fix inline radio element used with optionLabelPath (BS3)

kumy:
I've not used this part, but found something that looks like an issue while exploring the code.

### DIFF
--- a/addon/templates/components/bs3/bs-form/element/control/radio.hbs
+++ b/addon/templates/components/bs3/bs-form/element/control/radio.hbs
@@ -22,7 +22,7 @@
           {{yield}}
         {{else}}
           {{#if optionLabelPath}}
-            {{get option op tionLabelPath}}
+            {{get option optionLabelPath}}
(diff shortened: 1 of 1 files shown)


jelhan:
Good catch! Was introduced by https://github.com/kaliber5/ember-bootstrap/commit/508ae1f3db5d044fadbf5f63dca5195541716ae2. Seems like we are missing test coverage for the combination of having an `optionLabelPath` and `inline=true`.

jelhan:
@kumy Opened a pull request against your branch to add missing tests for this case: https://github.com/kumy/ember-bootstrap/pull/1

simonihmig:
@jelhan thanks for putting the test together. I think I will merge this quickly, as the fix itself is fine and I wanted to cut a release today. Can you target your PR to the main repo's master instead?

jelhan:
@simonihmig Sure. Please find it as #768.

PR #767, opened 2019-03-05
Fix IE11 errors in Collapse and Progress components

simonihmig:


simonihmig:
@jelhan fyi. It seems some of the broken things relate to your CSP/CSSOM changes. But don't worry about that, this is super easy to miss, and I did so as well in my review. I had the Saucelabs-based tests disabled as they proved to be very flaky, so CI didn't catch these. Now we have Browserstack integrated, which seems to work better so far, so we don't have to worry anymore! :wink:

jelhan:
@simonihmig I guess I'm used to much to `@babel/polyfill`. Had a quick look but didn't found any ESLint rules helping with that ones. Great that we have test coverage for IE11 again.

PR #768, opened 2019-03-05
test(BsForm): adds failing test for pull request #766

jelhan:
A test for BS4 is not needed cause `inline=true` only affects rendering an additional class for that one.


simonihmig:
Thank you!

Btw, please don't use this conventional changelog style of PR titles (`test(x): foo`, `feat(x): bar`), but just plain human readable descriptions of the changes. The PR titles are used for automatic changelog generation (via `lerna-changelog`), when one of the supported labels has been applied (see https://github.com/kaliber5/ember-bootstrap/blob/master/package.json#L129-L133). Should probably add this to the Contributing.md! Fyi ;)

Issue #769, opened 2019-03-05
`input radio` saves the whole dict in model


kumy:
I'm trying to use an `input radio` in a form, having label explicitely defined, and expecting a numeric id to be stored in `model`.

I've declared so far:
```hbs
    {{form.element
      id="moveType"
      model=move
      name="moveType"
      controlType="radio"
      label="Move type"
      options=moveTypeOptions
      optionLabelPath="label"
      property="type"
    }}
```
```js
export default Controller.extend({
  moveTypeOptions: [{
      value: 0,
      label: "I've dropped a GeoKret"
    }, {
      value: 1,
      label: "I've grabbed a GeoKret"
    }, {
      value: 3,
      label: "I've met GeoKret"
    }, {
      value: 5,
      label: "I've dipped a GeoKret"
    }, {
      value: 2,
      label: "Comment"
    }]
});
```
Now when an option is selected, the `model.type` is filled with the whole dict: `{value: 2,label: "Comment"}`. It would be better to have only value `2` in model. I would have expected to find a specific option in order to give the path of the value to be stored such as it's done for label (`optionLabelPath`), but didin't found something like that when exploring the code and templates. What about calling it `optionValuePath`? So we could write:
```hbs
    {{form.element
      id="moveType"
      model=move
      name="moveType"
      controlType="radio"
      label="Move type"
      options=moveTypeOptions
      optionLabelPath="label"
      optionValuePath="value"
      property="type"
    }}
```

```
$ ember bootstrap:info
DEPRECATION: ember-cli-babel 5.x has been deprecated. Please upgrade to at least ember-cli-babel 6.6. Version 5.2.8 located: geokrety-frontend -> ember-flag-icon-css -> ember-cli-babel
Npm packages:
ember-bootstrap: ^2.6.0 -> 2.6.0
ember-cli: ^3.8.1 -> 3.8.1
bootstrap: n/a
bootstrap-sass: ^3.4.1 -> 3.4.1
ember-cli-sass: ^10.0.0 -> 10.0.0
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 3
importBootstrapFont: true
importBootstrapCSS: false
```

jelhan:
Did you considered using a computed property? It's not only about setting `type` to the `id` of your object but also to map `type` to an object of your array that is selected.

jelhan:
[...]

PR #770, opened 2019-03-06
radio controls should be usable in block mode

jelhan:
Related to #769


jelhan:
I don't think the failing tests are related. Some tests of suite `Integration | Component | bs-carousel` are failing on `ember-canary` scenario. Also some node tests are failing.

simonihmig:
> Some tests of suite Integration | Component | bs-carousel are failing on ember-canary scenario.

Yes, related to ember-concurrency and canary. Nothing we can do about.

> Also some node tests are failing.

Strange. Will look into it.

Issue #762, opened 2019-02-28
Refactor Nav/Item active class support to use Router service


simonihmig:
The current hack is ugly, but was necessary when the router service wasn't a thing yet. We can query `isActive()` now, removing the need to observe any child `link-to` components. Also removes the need to have a dedicated yielded `nav/link-to` component extending from `link-to`, so that could be deprecated.

We can do this either now using https://github.com/rwjblue/ember-router-service-polyfill, or wait for the next major bump that drops support for older 2.x versions (preferable).

Issue #743, opened 2019-01-10
Update to ember-cli-babel 7+


simonihmig:
This will require at least Ember CLI 2.12 IIRC, so would be a breaking change! (assuming that our support for Ember 2.3 means Ember CLI 2.3 is also supported, which we have never articulated explicitly, but seems reasonable to assume)

PR #778, opened 2019-03-18
Updated demo examples to angle brackets syntax

### LINKED ISSUE #776: [Quest] Update documentation for angle bracket invocation
As the [Ember Guides](https://guides.emberjs.com/release/components/defining-a-component/) already have transitioned to angle bracket invocation instead of classic curlies, we should do the same here:

- [x] Update API docs (Yuidoc annotations in source files)
- [x] Update demo examples (in `dummy` app)

Help from the community would be very much appreciated here! 🙏

Comprehensive details of angle bracket invocation can be found on the [RFC](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md)

Just announce here if you want to work on something. And feel free to ask for help if things are unclear!

rajasegar:
Fixes #776


rajasegar:
Thanks @simonihmig  for the review comments, i will fix all of them and make sure all the tests are passing. Actually i have transformed them using this codemod https://github.com/rajasegar/ember-angle-brackets-codemod, looks like need to fix some items there.

rajasegar-c:
@simonihmig I have fixed all the lint errors and other bugs related to templates while converting to angle brackets syntax. A couple of things i wanted to highlight:

1. What are we going to do about the {{#nav.link-to }} components having route names as positional params like if i `<nav.linkTo "index">` it throws an error 
`"Assertion Failed: You must provide one or more parameters to the link-to component."`

2. There are still some lint errors about the usage of observers in component js

jelhan:
@simonihmig We have a lot of properties that have a name which could also refer to a HTML attribute. Most of them are converted to HTML attributes in this PR. I would tend to treat all documented properties as properties (`@` prefix) and consider passing them as HTML attributes as not supported for now - even if it might work.

In many cases both should work, especially if the property is bind to an attribute of the same name and doesn't use a computed property with a setter. We have a lot of these scenarios. Some other cases would break. One example is `type` property of `<BsButton>`. This does not refer to `type` attribute, which is controlled by `buttonType` but to Bootstrap layout type (e.g. `warning`).

This is also an important decision long term cause if we support both `<BsButton title="foo" />` and `<BsButton @title="foo" />` we limit our future ability to control `type` with a computed property setter. This might not be needed in most cases but was needed for #775 as an example. Not an easy decision IMO.

simonihmig:
> What are we going to do about the {{#nav.link-to }} components

I think for now we have no other choice than to continue using curlies here, anything else would be out of scope. But this needs to be addressed somehow, *probably* by deprecating/dropping `nav.linkTo` (see #762). 

> There are still some lint errors about the usage of observers in component js

They are configured as warnings, and getting rid of those is again out of scope here. So don't worry!

> We have a lot of properties that have a name which could also refer to a HTML attribute. Most of them are converted to HTML attributes in this PR. I would tend to treat all documented properties as properties (@ prefix) and consider passing them as HTML attributes as not supported for now - even if it might work.

I agree. As you said in many cases it probably would work, but pretty there are some where we would have to actually pass the attributes explicitly to the correct element (using `...attributes`), so this would become feature work, not documentation changes. We need to tackle this, but in separate effort.

@rajasegar can you please update that, so always pass args with `@` for now? I hope this should be easy by tweaking your codemod transform!?

> One example is type property of <BsButton>. This does not refer to type attribute, which is controlled by buttonType but to Bootstrap layout type (e.g. warning).

Good catch. Maybe we might have to introduce a breaking change here for ember-bootstrap 3.0, to remove that ambiguity?

---

Apart from all of that, we have a different problem here, which I didn't really think about: although the dummy app primarily serves as the demo app for ember-bootstrap.com, it nevertheless is part of the addon's build, which now breaks for all older Ember versions lacking support for angle bracket invocation: https://travis-ci.org/kaliber5/ember-bootstrap/builds/508229650?utm_source=github_status&utm_medium=notification

We *could*  add https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill as a `devDependency` here, but that would bring us only down to Ember 2.12, while this addon still supports Ember 2.3. 

Other than that:
* we could postpone merging this until we drop support for old Ember versions as part of a future ember-bootstrap 3.0 (need to write a quest issue for that!)
* we could do some broccoli-magic to remove the dummy app files in CI tests
* anything else?

jelhan:
> @rajasegar can you please update that, so always pass args with `@` for now? I hope this should be easy by tweaking your codemod transform!?

I'm not sure if I would go with passing everything as a property. I would only pass it as a property if it's documented as a property. So e.g. I wouldn't pass a `class` as property to `<BsButton>` but as an HTML attribute. So `{{bs-button class="foo"}}` shouldn't be changed to `<BsButton @class="foo">` but to `<BsButton class="foo">`. I think this might be automateable by passing a variable as property if it's consumed in JavaScript or template of the target component. I'm not sure if that information is available inside codemod.

> > One example is type property of . This does not refer to type attribute, which is controlled by buttonType but to Bootstrap layout type (e.g. warning).
> 
> Good catch. Maybe we might have to introduce a breaking change here for ember-bootstrap 3.0, to remove that ambiguity?

I think we should discuss this in a separate issue. It might affect even more components. Also I'm not sure if this is really an issue. Maybe `<BsButton type="submit" @type="success" />` doesn't feel strange anymore after we got used to angle bracket invocation syntax?

---

> Apart from all of that, we have a different problem here, which I didn't really think about: although the dummy app primarily serves as the demo app for ember-bootstrap.com, it nevertheless is part of the addon's build, which now breaks for all older Ember versions lacking support for angle bracket invocation

Oh that's a tough one. It definitely blocks the *Eventually update tests (less of a priority though)* task from #776.

I would tend to adding the polyfill for ember 2.12 as a long term solution. As far as I got #743 only dropping support for Ember CLI < 2.12 was considered. But dropping support for Ember Source < 2.12 is fine for me. This also gives us some more flexibility for long term. It only feels strange that we discuss this because of the dummy app build in tests.

Having that said I would like to land this one as soon as possible as I see a high risk of merge conflicts due to the amount of files touched. Especially if any manual work is invested and resolving conflicts would require more work than running the codemod again against latest master.

Therefor I tend to the option to remove dummy app from build in tests. But we have two acceptance tests that rely on dummy app. So we can't remove full dummy app but only most of it's content. Shouldn't be a big difference so, as the parts used in acceptance tests are namespaced under `/acceptance` route.

simonihmig:
> Therefor I tend to the option to remove dummy app from build in tests.

I tried this, spent quite some time on it, but didn't succeed. This should have been quite easy with `ember-cli-funnel`, but it seems to not work anymore: https://github.com/kellyselden/ember-cli-funnel/issues/77 😢

simonihmig:
> I would tend to adding the polyfill for ember 2.12 as a long term solution. As far as I got #743 only dropping support for Ember CLI < 2.12 was considered. But dropping support for Ember Source < 2.12 is fine for me. This also gives us some more flexibility for long term. It only feels strange that we discuss this because of the dummy app build in tests.

@jelhan I had thoughts for what needs to go into ember-bootstrap 3.0 for a while, including changing its support matrix, so no, this did not come up because of the issue here! :wink: Finally put this quest issue together: #780

rajasegar:
As of now, i am passing all args with @ except class , is this okay

tests/dummy/app/templates/demo/accordion.hbs
```diff
@@ -1,61 +1,58 @@
...
-      <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur ea eos hic neque quia sequi sunt. Amet at atque corporis earum et fugiat illum magnam nisi sapiente voluptatem. Laudantium, quam?</p>
-    {{/acc.item}}
-  {{/bs-accordion}}
+  <BsAccordion as |acc|>
+    <acc.item value="1" @title="First item">
+      <p> Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur ea eos hic neque quia sequi sunt. Amet at atque corporis earum et fugiat illum magnam nisi sapiente voluptatem. Laudantium, quam?  </p>
+    </acc.item>
+    <acc.item value="2" as |aitem|>
+      <aitem.title>
+        <strong> Second </strong>
```
simonihmig: These additional white spaces look weird, I guess they were introduced by the AST transform?

Haven't (yet) looked deeply into your codemod, are you using https://github.com/ember-template-lint/ember-template-recast for it?

.gitignore
```diff
@@ -29,3 +29,5 @@
 /.node_modules.ember-try/
 /bower.json.ember-try
 /package.json.ember-try
+
+.DS_Store
```
simonihmig: Can you remove that? This should rather live in your global git-ignore config IMHO.

simonihmig:
> As of now, i am passing all args with @ except class , is this okay

I think so. Just add a few minor additional remarks. But this is still blocked on the build issue discussed above, so probably not going to be merged as quick as we all want it. Sorry about that! And thanks for your effort again!

.gitignore
rajasegar: Done

tests/dummy/app/templates/demo/accordion.hbs
rajasegar: Yes, it was due to the codemod, actually i am using prettier at the final stage to pretty print the templates using the glimmer parser. My original idea was to use the ember-template-recast but it has got some issues with the current version
https://github.com/ember-template-lint/ember-template-recast/issues/16

simonihmig:
@rajasegar we now have a branch `dev-v3` to work on the upcoming 3.0 release, which will drop support for Ember<2.18. This should allow us to make this one pass the tests. 

Can you please:
* rebase this branch to `dev-v3`
* `yarn add ember-angle-bracket-invocation-polyfill`
* change the target of this PR to `dev-v3` instead of master

I hope this should fix most if not all issues here!

rajasegar:
Sure @simonihmig will let you know once done

rajasegar:
@simonihmig The rebase and invocation polyfill changes are done

tests/dummy/app/templates/demo/accordion.hbs
```diff
@@ -1,61 +1,58 @@
...
-      {{#aitem.body}}
-        <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur ea eos hic neque quia sequi sunt. Amet at atque corporis earum et fugiat illum magnam nisi sapiente voluptatem. Laudantium, quam?</p>
-      {{/aitem.body}}
-    {{/acc.item}}
-    {{#acc.item value="3" title="Third item"}}
-      <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Consectetur ea eos hic neque quia sequi sunt. Amet at atque corporis earum et fugiat illum magnam nisi sapiente voluptatem. Laudantium, quam?</p>
-    {{/acc.item}}
-  {{/bs-accordion}}
+  <BsAccordion as |acc|>
+    <acc.item value="1" @title="First item">
```
jelhan: `value` must be passed as property. It's not a HTML attribute in this case. There are several similar issues in this PR. It's related to this open issue in codemod: https://github.com/rajasegar/ember-angle-brackets-codemod/issues/3

I'm not sure how to address this issue in codemod as I'm not that used to ast. A general rule would be something like: If the property passed in curly invocation syntax is used in template or is used in JavaScript file of the component but not only in `attributeBindings` it must be passed as a property. But not sure if required information is available.

tests/dummy/app/templates/demo/progress.hbs
```diff
@@ -1,52 +1,105 @@
 <div class="bs-example">
   <!-- BEGIN-SNIPPET progress-demo -->
-  {{#bs-progress as |p|}}
-    {{p.bar value=50}}
-  {{/bs-progress}}
+  <BsProgress as |p|>
+    <p.bar @value="50" />
```
simonihmig: This seems to be failing in Ember 2.18, see the Travis log! Maybe it's related to `<p>` looking like a regular tag? Can you try changing this to something else, like `PG` to see if it fixes the problem?

tests/dummy/app/templates/demo/accordion.hbs
simonihmig: Can you remove those manually?

tests/dummy/app/templates/demo/button-group.hbs
```diff
@@ -1,42 +1,43 @@
...
-    {{#bg.button value=3}}3{{/bg.button}}
-  {{/bs-button-group}}
-  <p>You selected:</p>
+  <BsButtonGroup
+    value={{buttonGroupValue}}
+    @type="checkbox"
+    @onChange={{action (mut buttonGroupValue)}} as |bg|>
+    <bg.button value="1"> 1 </bg.button>
+    <bg.button value="2"> 2 </bg.button>
+    <bg.button value="3"> 3 </bg.button>
```
simonihmig: Same here, regarding `@value`

tests/dummy/app/templates/demo/dropdown.hbs
```diff
@@ -1,70 +1,79 @@
...
-        {{#menu.link-to "index"}}Home{{/menu.link-to}}
-      {{/menu.item}}
+  <BsDropdown @tagName="span" as |dd|>
+    <dd.toggle>
+      Dropdown
+      <span class="caret"></span>
+    </dd.toggle>
+    <dd.menu as |menu|>
+      <menu.item>
+        <menu.linkTo>Home</menu.linkTo>
```
simonihmig: This is wrong. The positional param has been swallowed. This is basically an extended `{{#link-to}}`, and with its required positional params it cannot be refactored currently into angle bracket syntax. Same for all other instances of `menu.link-to` or `nav.link-to`

tests/dummy/app/templates/demo/navbars.hbs
```diff
@@ -1,36 +1,68 @@
...
-            {{#ddm.item}}{{#ddm.link-to "demo.navbars"}}Navbars{{/ddm.link-to}}{{/ddm.item}}
-          {{/dd.menu}}
-        {{/nav.dropdown}}
-      {{/navbar.nav}}
-    {{/navbar.content}}
-  {{/bs-navbar}}
+    <navbar.content>
+      <navbar.nav as |nav|>
+        <nav.item>
+          <nav.linkTo >Home</nav.linkTo>
```
simonihmig: Same here

rajasegar:
@simonihmig No problem, will take a look at these issues

rajasegar-c:
@jelhan @simonihmig Addressed the review comments, still not able to get the custom tabs component to work, can you please check it and let me know what could be the issue?

tests/dummy/app/templates/demo/tabs.hbs
```diff
@@ -1,98 +1,103 @@
...
 <div class="bs-example">
   <!-- BEGIN-SNIPPET tab-route-->
   <div>
-    {{#bs-nav type="tabs" as |nav|}}
-      {{#nav.item}}{{#nav.link-to "demo.tabs.index"}}Tab 1{{/nav.link-to}}{{/nav.item}}
-      {{#nav.item}}{{#nav.link-to "demo.tabs.other"}}Tab 2{{/nav.link-to}}{{/nav.item}}
-    {{/bs-nav}}
+    <BsNav @type="tabs" as |nav|>
+      <nav.item>
+        {{#nav.link-to "demo.tabs.index"}}tab 1{{/nav.link-to}}
```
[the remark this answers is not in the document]
rajasegar: You mean to change back to angle brackets syntax, but it will error out without the positional params of the route

rajasegar:
@simonihmig Thanks for the review comments, now the Custom tabs component is fixed after addressing some of the review comments

tests/dummy/app/templates/demo/tabs.hbs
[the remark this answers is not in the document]
simonihmig: No, I was referring to the content block. I don't know why, but it was changed here from "Tab 1" to "tab 1". Did the codemod do that? Same in a few other places.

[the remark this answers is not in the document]
rajasegar: I am not sure about how it got changed, but i understood the issue, and fixed it in the latest commit

rajasegar:
@simonihmig Tabs names issue fixed

tests/dummy/app/templates/demo/dropdown.hbs
```diff
@@ -1,70 +1,79 @@
...
-  {{#bs-dropdown tagName="span" as |dd|}}
-    {{#dd.toggle}}Dropdown <span class="caret"></span>{{/dd.toggle}}
-    {{#dd.menu as |menu|}}
+  <BsDropdown @tagName="span" as |dd|>
+    <dd.toggle>
+      Dropdown
+      <span class="caret"></span>
+    </dd.toggle>
+    <dd.menu as |menu|>
       {{#menu.item}}
```
simonihmig: Should be angle brackets

tests/dummy/app/templates/demo/tabs.hbs
```diff
@@ -1,98 +1,103 @@
 <h2>Simple example</h2>
-
 <div class="bs-example">
   <!-- BEGIN-SNIPPET tab-demo -->
-  {{#bs-tab as |tab|}}
+  <BsTab as |tab|>
     {{#tab.pane title="Tab 1"}}
```
simonihmig: Should be angle brackets

tests/dummy/app/templates/demo/tabs.hbs
```diff
@@ -1,98 +1,103 @@
...
       <p>Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?</p>
     {{/tab.pane}}
     {{#tab.pane title="Tab 2"}}
```
simonihmig: same

tests/dummy/app/templates/demo/progress.hbs
```diff
@@ -1,52 +1,105 @@
...
+    <pg.bar
+      @value="5"
+      @minValue={{minValue}}
+      @maxValue={{maxValue}}
+      @showLabel={{showLabel}}
+      @roundDigits={{roundDigits}}
+      @type="danger"
+      @striped={{striped}}
+      @animate={{animate}}/>
+  </BsProgress>
```
simonihmig: This is still failing with a template compiler error for Ember 2.18, no idea really why?
https://travis-ci.org/kaliber5/ember-bootstrap/jobs/528738875#L662

simonihmig: Ah, I don't know why, but I was able to fix this by putting the closing `/>` on the following line. Same for the other two `pg.bar`s.

Pushed to your branch, should hopefully get green now...

rajasegar-c:
@simonihmig Fixed the dropdown and tabs demo, the build is passing now i guess

PR #782, opened 2019-03-21
nav items should not have role="presentation"

### LINKED ISSUE #781: Tabs role="presentation" isn't a11y compliant
When running a11yAudit(); I am seeing the following issue:

```
Error: Assertion Failed: The page should have no accessibility violations. Violations:
[serious]: <ul> and <ol> must only directly contain <li>, <script> or <template> elements 
Violated 1 time. Offending nodes are: 
<ul id="ember553" class="nav nav-tabs nav-justified ember-view">
```

Here is the "offending" markup.

Captura de pantalla 2019-03-21 a la(s) 3 40 54 p  m

I asked Mel about this on Discord and she pointed out that the `role="presentation"` changes the `li` so that it's not `li` anymore.

I looked to see if this role was added by bootstrap itself but it doesn't appear to be:

https://getbootstrap.com/docs/4.3/components/navs/#tabs

Then I looked that the ember-bootstrap tabs demo page:

https://www.ember-bootstrap.com/#/components/tabs

And inspected the tab examples which showed:

Captura de pantalla 2019-03-21 a la(s) 4 08 07 p  m

So it appears that ember-boostrap adds the `role="presentation"`.

jelhan:
Nav items component is also used for tabs.

Fixes #781

### DIFF
--- a/addon/components/base/bs-nav/item.js
+++ b/addon/components/base/bs-nav/item.js
@@ -21,7 +21,6 @@ export default Component.extend(ComponentParent, {
   layout,
   classNameBindings: ['disabled', 'active'],
   tagName: 'li',
-  ariaRole: 'presentation',
 
   /**
    * Render the nav item as disabled (see [Bootstrap docs](http://getbootstrap.com/components/#nav-disabled-links)).
--- a/addon/components/bs3/bs-nav/item.js
+++ b/addon/components/bs3/bs-nav/item.js
@@ -1,1 +1,5 @@
-export { default } from 'ember-bootstrap/components/base/bs-nav/item';
+import NavItem from 'ember-bootstrap/components/base/bs-nav/item';
+
+export default NavItem.extend({
+  ariaRole: 'presentation',
+});
--- a/addon/components/bs3/bs-nav/item.js
+++ b/addon/components/bs3/bs-nav/item.js
@@ -1,5 +1,1 @@
-import NavItem from 'ember-bootstrap/components/base/bs-nav/item';
-
-export default NavItem.extend({
-  ariaRole: 'presentation',
-});
+export { default } from 'ember-bootstrap/components/base/bs-nav/item';
--- a/tests/integration/components/bs-nav/item-test.js
+++ b/tests/integration/components/bs-nav/item-test.js
@@ -4,7 +4,6 @@ import { render, click } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import test from 'ember-sinon-qunit/test-support/test';
 import setupNoDeprecations from '../../../helpers/setup-no-deprecations';
-import { testBS3, testBS4 } from '../../../helpers/bootstrap-test';
 
 module('Integration | Component | bs-nav/item', function(hooks) {
   setupRenderingTest(hooks);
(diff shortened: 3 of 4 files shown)


simonihmig:
I just wonder why we should have that role for BS3, if it makes them indeed inaccessible, or at least violate the a11y linter. Maybe just omit it?

Related: https://stackoverflow.com/questions/31028008/why-do-bootstrap-tabs-have-role-presentation

jelhan:
To be honest I don't know much about a11y and I got the feeling this is a complex question after reading that StackOverflow post. So I don't feel like the correct one to make that decision.

BitBrit:
This is an interesting subject.   I don't quite understand why role="presentation" was added to the bootstrap tabs and I don't really know how screen readers read these roles.  According to the stackoverflow link above the role="presentation" is so that the screen reader doesn't interpret the list as a list (described here: https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role_effects) but as set of tabs. However, this seems to be superseded by: https://www.w3.org/TR/WCAG20-TECHS/H48.html 

As an example: this is how list elements with role="presentation" are viewed: https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role_examples

There doesn't seem to be many reliable examples for how it should be used. This page basically says that you can't rely on it:
https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/presentation/PresentationRoleExamples.html

This page has no examples:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role

My questions are: should the ariaAudit() test be flagging the use of lists as tabs links as an error? Or should this be accepted? Should a different role be used instead of presentation? Like `tablist` and `tab`?

addon/components/base/bs-tab/pane.js
```diff
@@ -20,7 +20,6 @@ export default Component.extend(ComponentChild, TransitionSupport, {
   layout,
   classNameBindings: ['active', 'usesTransition:fade'],
   classNames: ['tab-pane'],
-  ariaRole: 'tabpanel',
```
simonihmig: I am just a bit puzzled: didn't you want to remove the `role="presentation"` from a `nav.item`? This is something different here!?

jelhan: :100: Seems like I was a little bit tired when I prepared this PR. :confused:

jelhan:
Fixed implementation and rebased to fix merge conflicts. Also added some tests.

Implementation currently follows markup in Bootstrap docs. That means that nav items have an `role="presentation"`. I have the feeling that this is a complex topic and maybe application should decided to not follow Bootstrap defaults? Also I'm not sure if removing `role="presentation"` for BS3 should be considered a bug fix or a breaking change.

jelhan:
Tests for IE11 seems to be still flickering even after migration to BrowserStack. Restarted the BrowserStack tests and another test was failing in IE11. All in all not related to this PR.

simonihmig:
> Implementation currently follows markup in Bootstrap docs. 

I still find it inconsistent and illogical to have different roles for basically the same markup, just because of the used Bootstrap version (which is more about different implementation of CSS classes and so on, but a11y-wise should behave the same). Either that `role` does more than good or not. So we should remove it or add it to all versions IMO. 

Given that...
* the old BS3 docs are probably not maintained anymore, at least such a11y fixes won't find their way to the old docs
* the latest (BS4) markup removed the `role`
* and the a11y audit mentioned in #781 sees them as errors

...I would tend to remove those for both versions!

> Also I'm not sure if removing role="presentation" for BS3 should be considered a bug fix or a breaking change.

Let's declare it a bug fix!? :wink:

Honestly, considering every change of markup a breaking change leaves us little ground to iterate. Even for things that are more like adding an optimization or feature, they *could*  break things, e.g. in case of some weird CSS rules defined by the host app. But this would leave us with no ground to change anything, right?

jelhan:
> ...I would tend to remove those for both versions!

Very valid arguments. :+1: I'm convinced. :smiley: Will refactor accordingly.

> > Also I'm not sure if removing role="presentation" for BS3 should be considered a bug fix or a breaking change.
> 
> Let's declare it a bug fix!? :wink:
>
> Honestly, considering every change of markup a breaking change leaves us little ground to iterate. Even for things that are more like adding an optimization or feature, they could break things, e.g. in case of some weird CSS rules defined by the host app. But this would leave us with no ground to change anything, right?

I wasn't worried about the markup change but about not following markup in Bootstrap docs anymore. But I'm fine with considering it a bug fix.

jelhan:
Refactored accordingly. Left tests as I think it's easy to regress for BS3 as it isn't following markup in docs anymore.

simonihmig:
Thanks @jelhan!

PR #777, opened 2019-03-10
Update API docs for Angle Bracket Invocation

betocantu93:
Yuidocs for #776

Issue #780, opened 2019-03-20
[QUEST] Road to 3.0


simonihmig:
This aims at collecting all the issues/Todos that would result in breaking changes, so we can aggregate all those in a single major version bump.

- [x] several smaller [breaking change issues](https://github.com/kaliber5/ember-bootstrap/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking)

~- [ ] Migrate to the Octane programming model~
 ~- [ ] Refactor from Ember object model to native class + decorators. This *might*  break users `.extend`ing from e-b components for older Ember versions I believe. (codemod: https://github.com/ember-codemods/ember-es6-class-codemod)~
  ~- [ ] Use angle bracket invocation internally~
  ~- [ ] Use angle bracket invocation in tests (see #776)~

- [x] Related to angle bracket invocation, make sure that setting HTML attributes works consistently
  - ~[ ] deprecate props including documentation (form element!) whose only purpose is supporting HTML attributes~ deprecations will follow, see #826 
  - [x] make sure HTML attributes get propagated to the appropriate HTML element using `...attributes`
  - ~[ ] figure out if we should remove name clashes (attributes vs. props), e.g `type` for Buttons (see https://github.com/kaliber5/ember-bootstrap/pull/778#issuecomment-474323269)~ *no last-minute breaking changes should be introduced. Deprecations for 4.0 could be added though during 3.x*

- [x] drop node 6 support (end of life at end of April)

- [x] depending on the further outcome of #778, we might be forced to delay introducing angle bracket invocation in docs to 3.0  

### Ember support

~To be decided is the minimum Ember version to be supported starting with 3.0:~
*We will go forward support only Ember 2.18+*

* **2.12** seems to be the most conservative choice, which still enables us to move forward (as it includes Glimmer2, and is supported by `ember-angle-bracket-invocation-polyfill`
* **2.18** is the middle ground, as some users might still be stuck at that version, not able yet to upgrade to Ember 3.0 due to its breaking changes
* **3.4** is the only LTS that is still supported by Ember itself. So when introducing a breaking change anyway, do we have to support Ember versions that aren't supported by Ember itself? Or can we be that bold to require 3.4? It would allow us to not have to use `ember-angle-bracket-invocation-polyfill`, and have more predictable native class semantics (see https://github.com/pzuraq/ember-native-class-polyfill)

jelhan:
[...]

PR #783, opened 2019-03-24
Add linkTo support for nav.item, deprecate nav.link-to

### LINKED ISSUE #762: Refactor Nav/Item active class support to use Router service

simonihmig:
Closes #762. 

As Bootstrap expects to have the `active` class on a Nav's `<li>` rather than a child `<a>` (which `{{link-to}}` supports out of the box), the `nav.item` component had to be aware of the active state of its child link. We had to do quite some trickery before: 
* required the use of `nav.link-to` instead of a plain `{{link-to}}`
* this extended `Ember.LinkComponent` and registered itself to the parent `nav.item`
* the `nav.item` then observed the `active` property of its child `nav.link-to` component

Before having Ember's (not so new anymore) router service, I found no better way to do this (with an ergonomic DX), but the implementation is quite awkward (and was error prone in the past, though stable now), in that it queries internal properties of a component (link-to) for some state (is linked route active), that actually lives somewhere else (the router). 

The new proposed API here expects you to set the link params to the `nav.item`, so the component can be aware of its link, and if present will implicitly generate the link by itself, not requiring the use of `nav.link-to` anymore. Uses the Router service (with a polyfill for older versions). The `link-to` component is still around for compatibility reasons, but is being deprecated here and would be removed for 3.0!

Before:

```hbs
{{#nav.item}}{{#nav.link-to "index"}}Home{{/nav.link-to}}{{/nav.item}}
{{#nav.item}}{{#nav.link-to "foo" this.model}}Foo{{/nav.link-to}}{{/nav.item}}
```

After: 

```hbs
{{#nav.item linkTo="index"}}Home{{/nav.item}}
{{#nav.item linkTo=(array "foo" this.model)}}Foo{{/nav.item}}
```

If that's to inflexible for some use case, you can still setup things manually, e.g. by using https://github.com/rwjblue/ember-router-helpers:

```hbs
{{#nav.item active=(is-active "index")}}{{#link-to "index"}}Home{{/link-to}}{{/nav.item}}
```

This was the best solution I could think of, which makes the most common case easy to use without a ton of verbosity. Which does not mean there isn't any possibility for a better one. 🙂 So keeping this open for a while, as this will eventually introduce a breaking change, so folks can voice their thoughts! /cc @jelhan @lolmaus 

PS: this will cause a few merge conflicts with #777 and #778, so will try to get at least #777 merged first, and cleanup the conflicts here...


addon/components/base/bs-nav.js
```diff
@@ -17,11 +17,11 @@ import layout from 'ember-bootstrap/templates/components/bs-nav';
 
  ```hbs
  {{#bs-nav type="pills" as |nav|}}
-   {{#nav.item}}
-      {{#nav.link-to "foo"}}Foo{{/nav.link-to}}
+   {{#nav.item linkTo="foo"}}Foo
+     Foo
```
jelhan: Having `Foo` two times. Should be removed on first line.

addon/components/base/bs-nav/link-to.js
```diff
@@ -12,4 +13,15 @@ import ComponentChild from 'ember-bootstrap/mixins/component-child';
...
 export default LinkComponent.extend(ComponentChild, {
+  init() {
+    this._super(...arguments);
+    deprecate('The link-to component yielded by bs-nav is deprecated. Use the linkTo property of the yielded item component instead!\n' +
+      'Before: {{#nav.item}}{{#nav.link-to "index"}}foo{{/nav.link-to}}{{/nav.item}}\n' +
+      'After: {{#nav.item linkTo="index"}}foo{{/nav.item}}',
+      false,
+      {
+        id: 'ember-bootstrap.nav.link-to',
+        until: '3.0'
```
jelhan: This will be a very short deprecation period isn't it?

simonihmig: Well, who knows? :wink: Don't how quickly all of #780 can be finished...
But 4.0 would be very long I assume, and as this is public API, we can only remove with a major version, right?

simonihmig:
Failing on older versions of Ember, need to investigate this... 🤔

addon/components/base/bs-nav/item.js
```diff
@@ -22,6 +24,22 @@ export default Component.extend(ComponentParent, {
   classNameBindings: ['disabled', 'active'],
   tagName: 'li',
   ariaRole: 'presentation',
+  router: service(),
+
+  /**
+   * If set will wrap the item's content in a link. Accepts either a string with the route name, or an array
+   * with the route name and one or multiple models / query params, similar to the positional params of `{{link-to ...}}`
```
jelhan: Maybe docs should also mention query params?

simonihmig: It (briefly) does: "with the route name and one or multiple models / query params". Would you expect/suggest something more verbose?

jelhan: Oh, of course. 😧 Sorry for the noice.

PR #786, opened 2019-03-26
Dropdown respects closeOnMenuClick=false even after re-opening the menu

marceloandrader:
If we don't add the volatile in the menuElement, it will cache the first dom element generated for the menu, but because is being lazily rendered the menuElement will not be the same dom element next time is opened.

### DIFF
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -214,7 +214,7 @@ export default Component.extend({
    */
   menuElement: computed(function() {
     return document.getElementById(`${this.get('elementId')}__menu`);
-  }),
+  }).volatile(),
 
   /**
    * @property toggleElement
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -147,7 +147,7 @@ import layout from 'ember-bootstrap/templates/components/bs-dropdown';
  @extends Ember.Component
  @public
  */
-export default Component.extend({
+let component = Component.extend({
   layout,
   classNameBindings: ['containerClass'],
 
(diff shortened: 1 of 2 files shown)


addon/components/base/bs-dropdown.js
```diff
@@ -214,7 +214,7 @@ export default Component.extend({
    */
   menuElement: computed(function() {
     return document.getElementById(`${this.get('elementId')}__menu`);
-  }),
+  }).volatile(),
```
simonihmig: `.volatile()` has been deprecated, so we shouldn't use it. (in fact I removed a few of those instance recently because of this). As the tests are set up to fail for thrown deprecations, Travis is failing here.

Instead of this we should be able to use a standard ES5 getter, though it requires some awkward setup (until we move to native classes), see e.g. https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-modal.js#L696-L725

tests/integration/components/bs-dropdown-test.js
```diff
@@ -137,6 +137,32 @@ module('Integration | Component | bs-dropdown', function(hooks) {
     assert.dom(dropdownVisibilityElementSelector()).hasClass(openClass(), 'Dropdown is open');
   });
 
+  test('clicking dropdown menu when closeOnMenuClick is false will not close it, even after opening the menu again', async function (assert) {
```
simonihmig: Looks good. Did you check that this was failing without the change in the component?

marceloandrader:
Yes it was failing without the volatile. And in my app.

addon/components/base/bs-dropdown.js
marceloandrader: I've updated the code to use that syntax, thanks for the quick answer.

simonihmig:
Thank you @marceloandrader! 🙏

Issue #731, opened 2018-12-21
Improve the semantics of form's onInvalid action


simonihmig:
Picking up what was discussed [here](https://github.com/kaliber5/ember-bootstrap/pull/730#discussion_r243525543):
* validations should be shown after an eventually returned Promise from `onInvalid` has settled
* provide some means to skip showing the validations based on the action's return value

*Potentially a breaking change!*

simonihmig:
@jelhan would you be interested in working on this, as this was based on your suggestion?

jelhan:
@simonihmig, thanks a lot for picking this one up. I'm sorry for not replying earlier. Haven't had much time the last weeks.

> provide some means to skip showing the validations based on the action's return value

I think this needs some more thoughts. E.g. it's not clear yet, when the validations should be shown again. Also we haven't seen an actual use case yet. If an application is able to resolve the validation errors automatically, it should also update the validation state. In that case there isn't any need to hide the validations. If the validation shouldn't be shown before a transaction caused by an invalid form submission, a long-pending promise could be returned. The same applies if some user interaction should be done on an invalid submission (e.g. showing a modal). Last but not least applications should be able to implement this on their own without overriding internal APIs as soon as we provide a way to control the visibility of validations in general (#364). So we should focus on that one first in my opinion.

simonihmig:
I agree, see also my comment here https://github.com/kaliber5/ember-bootstrap/pull/820#issue-283854425.

I will add a note to #364 and keep that issue to track controlling the visibility of validations.

simonihmig:
> validations should be shown after an eventually returned Promise from onInvalid has settled

Implemented by #820, closing this.

PR #789, opened 2019-03-27
Yield submit action from BsForm

jelhan:
Partly implements #476


jelhan:
@svparijs Does this suit your needs?

addon/components/base/bs-form.js
```diff
@@ -431,6 +431,10 @@ export default Component.extend({
     resetSubmissionState() {
       this.set('isSubmitted', false);
       this.set('isRejected', false);
-    }
+    },
+
+    submit() {
+      this.submit();
```
simonihmig: Shouldn't we do return a promise here? 

It would require some refactoring of `this.submit()` to also correctly return the promise. Including a rejected one, which is currently catched, so would need to throw again. Probably a good opportunity to refactor the .then-hell to async/await!? ;)

svparijs: I agree, for this:
"{{bs-button onClick=(action "authenticate" form) buttonType="submit" preventConcurrency=true type="primary" defaultText=(t 'button.login') pendingText=(t 'button.pending.login') fulfilledText=(t 'button.resolved.login') rejectedText=(t 'button.rejected.login') class="btn-lg btn-block mb-3"}}"

We need to get the Promise to the button.

jelhan: > Shouldn't we do return a promise here?

Of course. Implementation was more like a PoC. Should have added PoC to title.

> I agree, for this:
> "{{bs-button onClick=(action "authenticate" form) buttonType="submit" preventConcurrency=true type="primary" defaultText=(t 'button.login') pendingText=(t 'button.pending.login') fulfilledText=(t 'button.resolved.login') rejectedText=(t 'button.rejected.login') class="btn-lg btn-block mb-3"}}"

What you are trying to achieve is the `<form.button>` or maybe better `<from.submitButton>` component missing of #638. I would recommend to use form submission state directly as long as this one is not available. Otherwise the form and button state may fail out of sync if form is submitted in another way. I have something like this in mind:

```hbs
<BsForm @preventConcurrency={{true}} as |form|>
  <BsButton type="submit" @type="primary">
     {{#if form.isSubmitting}}
       {{t "button.pending.login"}}
     {{else if form.isSubmitted}}
       {{t "button.resolved.login"}}
     {{else if form.isRejected}}
       {{t "button.rejected.login"}}
     {{else}}
       {{t "button.login"}}
     {{/if}}   
  </BsButton>
</BsForm>

Comment (svparijs) [reply addon/components/base/bs-form.js]: Well the initial isSubmitting works but then when the submit isRejected its not working properly, I'm not sure if this is a package issue or a programming issue, see the action below.
The request done returns a 401 authentication failed.
```
authenticate() {
  const credentials = this.getProperties('username', 'password'),
        authenticator = 'authenticator:jwt';

      return this.get('session').authenticate(authenticator, credentials).then(() => {
        this.set('showError', false);
      }).catch(() => {
        this.set('showError', true);
      });
}
```

Comment (jelhan) [reply addon/components/base/bs-form.js]: I assume that you bind `authenticate` action to `onSubmit` event of `<BsForm>`. Since you are catching rejects thrown in authentication, the Promise returned by `authenticate` will always resolve. Therefor you won't see an `isRejected` state but should see a `isSubmitted` one. Please open a separate ticket if you can reproduce any issue with `form.isRejected` / `form.isSubmitted` logic.

**Comment by jelhan** [comment]
@simonihmig Ready to review from my side.

I'm not sure about the testing of value returned by yielded action. I guess one could write an integration test using submission state handling of `<BsButton>` but that would also add a lot of complexity to the test. Therefor I've choosen unit testing for that case but if you know any better way, I'm open to any suggestions.

**Comment by simonihmig** [comment]
> I'm not sure about the testing of value returned by yielded action. I guess one could write an integration test using submission state handling of <BsButton> but that would also add a lot of complexity to the test. Therefor I've choosen unit testing for that case but if you know any better way, I'm open to any suggestions.

I remembered some PRs to Ember, e.g. this one: https://github.com/emberjs/ember.js/pull/17227/files#diff-e23a01fa7d1b74bfda7985619d28ec69
It would need to look a bit different, as the test harness is different, but the idea would be the same: 
* you create a dummy component in the test, and `this.owner.register()` it
* the rendered template passes the yielded action to this component
* the component sets the passed action to some variable of the test itself, e.g. in `init`, as it has access to its scope
* this way we get a reference to the action into our test, at which point it is just a function, so we can call it and assert for its return value/promise state

Looks a bit weird, but has the advantage that it resembles how the component actually works. In your unit test, you access `component.actions.submit`, which has the drawback that
* having the submit function on the `actions` hash is an implementation detail (and not required for Glimmer component for example)
* it is not necessarily the same thing as the yielded action

File: addon/components/base/bs-form.js [hunk 2]
```diff
@@ -343,64 +343,67 @@ export default Component.extend({
...
+      await this.get('onInvalid')(model, error);
+
+      if (this.get('isDestroyed')) {
+        return;
+      }
+
+      this.set('isRejected', true);
+      this.decrementProperty('pendingSubmissions');
+
+      throw error;
```
jelhan: Should we reject the validation error or a generic error? Do we consider the value the Promise resolves with as part of public API or just the promise state (rejected / fulfilled)?

addon/components/base/bs-form.js
```diff
@@ -343,64 +343,67 @@ export default Component.extend({
...
+      }
+    }
+  },
+
+  async submit() {
+    try {
+      await this.submitHandler(...arguments);
+    } catch(err) {
+      // catch error throws by form's validation
+      // TODO: do not catch other errors
```
jelhan: Catching errors that are not validation errors might make debugging much more difficult. But I'm not sure how we should handle this one.

jelhan:
@simonihmig If you have time, I would appreciate your input in the two questions raised a few days ago. Beside that ones the PR is finished from my side. Tests are refactored to integration tests using the test component approach mentioned in your comment.

addon/components/base/bs-form.js
simonihmig: > Should we reject the validation error or a generic error

The former I would say. So keep it like it is.

> Do we consider the value the Promise resolves with as part of public API or just the promise state

Hm, good question. I would say also the former!?

addon/components/base/bs-form.js
simonihmig: What was the reason to catch here in the first place?

simonihmig:
@jelhan sorry for the delay on my part here!

addon/components/base/bs-form.js
jelhan: We must not end up with uncaught errors in promises for submit event due to throwing validation errors. There are to options to achieve that one (if I didn't missed something):

- Not rethrowing validation errors in `submitHandler()` if it was called by submit event (`submit()` hook).
- Catching validation errors rethrown in `submitHandler()` in `submit()` event hook.

The second one is currently implemented but has the issue of catch all errors thrown by sumit handler. That makes debugging way more difficult as the errors aren't reported to developer console. So maybe the first option would be better? I twould look like:

```js
Component.extend({
  submitHandler(e, rethrowValidationErrors = true) {
    // ...

    try {
      record = await (this.get('hasValidator') ? this.validate(model) : null);
    } catch (error) {
       // ...

       if (!rethrowValidationErrors) {
         return;
       }

       throw error;
    }

    // ...
  },

  async submit(event) {
    // should not reject on validation error as this would end up
    // as an uncaught rejection
    await this.submitHandler(event, false);
  },

  actions {
    submit() {
      // should reject on validation errors
      return this.submitHandler();
    }
  }
});

File: addon/components/base/bs-form.js [hunk 4]
```diff
@@ -431,6 +434,10 @@ export default Component.extend({
     resetSubmissionState() {
       this.set('isSubmitted', false);
       this.set('isRejected', false);
-    }
+    },
+
+    async submit() {
```
simonihmig: this is duplicated

addon/components/base/bs-form.js
simonihmig: > So maybe the first option would be better?

Agree, we shouldn't catch all errors!

addon/components/base/bs-form.js
jelhan: I'm not sure what you mean. This is the `submit` action that's yielded by the component. We also implement the `submit` hook executed on submit events. Both are using the `submitHandler()`.

simonihmig: Nevermind! In the diff view this looked as if was overriding the other `submit`. The `actions: {` part was not shown, so I got confused 🤦‍♂️😂

simonihmig:
This is failing in Ember <= 2.18. @jelhan any idea why? 🤔

jelhan:
> This is failing in Ember <= 2.18. @jelhan any idea why?

@simonihmig Sorry, didn't had much time the last weeks. I investigated it a little bit and noticed that it was caused by `async`/`await`. I guess some weird issue with runloop and native Promises? Could fixed most of them for Ember 2.18 by refactoring tests but one was left and < 2.18 was still failing. After dropping `async`/`await` and reverting to an ugly `then`/`catch` hell, tests are passing (except for IE 11).

simonihmig:
> tests are passing (except for IE 11).

@jelhan added a commit to fix that IE11 issue: refactored the `for ... of` loop (was only in tests!), as it requires `Symbol` support. Should hopefully turn green now...

Other than that, this should be ready to merge, right?

jelhan:
> @jelhan added a commit to fix that IE11 issue: refactored the for ... of loop (was only in tests!), as it requires Symbol support. Should hopefully turn green now...

Thanks!

> Other than that, this should be ready to merge, right?

I didn't verified yet but I think this changes behavior of `onInvalid` action and showing of validation errors (#731). This may be considered a breaking change and may require an additional test. Could have a look into it next week. Sorry for the delay.

simonihmig:
> I think this changes behavior of onInvalid action and showing of validation errors (#731) 

I don't see this happening: `onInvalid` is still called *after* setting `showAllValidations`, so not affecting if/when validations are shown (which #731 suggests): https://github.com/kaliber5/ember-bootstrap/pull/789/files#diff-3692032a42f2bb2d542b42a5eb27b6c3R401

simonihmig:
@jelhan thanks for working on this! Merging this as it LGTM. I am not pushing for an immediate release, so if you come to the conclusion that there is an (breaking) issue here, please tell me! :)

PR #791, opened 2019-03-28
Add ember-bootstrap-model-select to addon integrations

nickschot:
I'm about to release the first stable release of ember-bootstrap-model-select, so seems like a good time to add it here.


simonihmig:
Thanks Nick! :smile:

PR #795, opened 2019-03-30
Refactored templates to Octane style (angle bracket invocation, named args, this.args), allows setting HTML attributes using angle bracket invocation to the proper element

simonihmig:
- [x] enable skipped splattributes tests when https://github.com/kybishop/ember-popper/issues/98 is resolved


addon/templates/components/common/bs-modal.hbs
```diff
@@ -1,33 +1,35 @@
...
+        @fade={{this.fade}}
+        @showModal={{this.showModal}}
+        @keyboard={{this.keyboard}}
+        @size={{@size}}
+        @backdropClose={{this.backdropClose}}
+        @inDom={{this.inDom}}
+        @paddingLeft={{this.paddingLeft}}
+        @paddingRight={{this.paddingRight}}
+        @centered={{bs-eq this.position "center"}}
+        ...attributes
```
jelhan: Not sure if you are aware of but using `...attributes` and innerHTML components together results in attributes being applied twice: https://ember-twiddle.com/74a339b46d1a87421927992e5a073d1f?openFiles=templates.application.hbs%2Ctemplates.components.my-component.hbs

simonihmig: Hm, yes, good catch! I think we could remove the default wrapper element here actually...

simonihmig:
Failing tests due to https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/issues/67

simonihmig:
This should be ready to merge from my PoV. @jelhan wanna have a 2nd look?

addon/templates/components/bs3/bs-accordion/body.hbs
```diff
@@ -1,5 +1,5 @@
-{{#component collapseComponent ariaRole="tabpanel" collapsed=collapsed class="panel-collapse"}}
-  <div class="panel-body {{class}}">
+{{#component this.collapseComponent ariaRole="tabpanel" collapsed=@collapsed class="panel-collapse"}}
```
jelhan: Shouldn't this be `aria-role="tabpanel"` as the HTML attribute is name `aria-role` and not `ariaRole`?

addon/templates/components/common/bs-modal-simple.hbs
```diff
@@ -1,40 +1,41 @@
...
+      @showModal={{this.showModal}}
+      @keyboard={{this.keyboard}}
+      @size={{@size}}
+      @backdropClose={{this.backdropClose}}
+      @inDom={{this.inDom}}
+      @paddingLeft={{this.paddingLeft}}
+      @paddingRight={{this.paddingRight}}
+      @centered={{bs-eq this.position "center"}}
+      ...attributes
+      @class={{@class}}
```
jelhan: Any reason for the order of `...attributes` and `@class` here? I thought properties won't be affected anyways by `...attributes`. If I would prefer an order of having all properties before attributes for readability.

addon/templates/components/bs3/bs-accordion/body.hbs
simonihmig: As the component helper does not allow to be used in a angle brackets like fashion (it's not a component), we cannot set HTML attributes in the way we could w/ angle bracket components. So this sets the `ariaRole` *property* of `Ember.Component` (see https://api.emberjs.com/ember/3.10/classes/Component/properties/ariaRole?anchor=ariaRole), that is mapped to the `role` attribute (through our beloved `attributeBindings`) internally.

addon/templates/components/common/bs-modal-simple.hbs
simonihmig: Hm, no. I probably have used `class={{@class}}` before, and wanted to make sure that when setting `@class` this takes precedence of the splattributes. But as this is a prop again, this does not make really sense, will fix it!

addon/templates/components/bs3/bs-accordion/body.hbs
jelhan: Thanks a lot for this detailed explanation. :+1:

PR #804, opened 2019-04-16
Assertion to prevent setting both property and value on a form element

rrglomsk:
The documentation does mention that the element's value will be bound if you use the `property` property, but we wound up with a few stray `value` properties set along with `property` on form elements in our application which lead to some unexpected bugs cropping up in production. It seems like a mistake that's not too difficult to make, so just an extra safety measure :)!

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -983,7 +983,7 @@ export default FormGroup.extend({
       this.set('showValidationOn', ['focusOut']);
     }
     if (!isBlank(this.get('property'))) {
-      assert('You cannot set both property and value on a form element', isBlank(this.get('value')));
+      assert('You cannot set both property and value on a form element', this.get('value') === null);
       defineProperty(this, 'value', alias(`model.${this.get('property')}`));
       this.setupValidations();
     }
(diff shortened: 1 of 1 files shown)


jelhan:
Tests fails cause of a deprecation triggered in ember beta by carousel:

> DEPRECATION: Please migrate from Ember._setComputedDecorator to Ember._setClassicDecorator [deprecation id: ember._setComputedDecorator]

Not related to this PR at all.

addon/components/base/bs-form/element.js
```diff
@@ -983,6 +983,7 @@ export default FormGroup.extend({
       this.set('showValidationOn', ['focusOut']);
     }
     if (!isBlank(this.get('property'))) {
+      assert('You cannot set both property and value on a form element', isBlank(this.get('value')));
```
simonihmig: This would allow having `property` but `value` being set to an empty string, which is probably a very common initial value. So the assertion wouldn't trigger in that case. Checking for a `null` value (which is its default value) might be better?

```suggestion
      assert('You cannot set both property and value on a form element', this.get('value') === null);
```

simonihmig:
Thanks @rrglomsk, that's a good improvement. Just a minor suggestion on my part...

> Tests fails cause of a deprecation triggered in ember beta by carousel:

Yes, fixed by #802.

addon/components/base/bs-form/element.js
rrglomsk: Oh that's a good point about initial values- a null check sounds better to me too, thanks!! 😄

simonihmig:
Thanks again!

rrglomsk:
No problem 😄!!

PR #825, opened 2019-06-13
Add support for ember-autoresize

### LINKED ISSUE #110: Text area to auto resize as user types
we have a need for a feature that expands the textarea as the user types. For that right now we don't use ember bootstrap textarea components and just use the normal textarea helper with this addon as well https://github.com/tim-evans/ember-autoresize. For consistency sake, we'd love to continue using components created by your addon, would there be any way to have this functionality built in?

Redsandro:
Add support for [`ember-autoresize`](https://github.com/adopted-ember-addons/ember-autoresize). Adding these attribute bindings removes a whole lot of fork magic to get this done.

Fix #110


simonihmig:
First of all thank you for taking the time to contribute!

Unfortunately I see some problems here:
* I had a brief look into `ember-autoresize`, and neither in the Readme nor in the actual code could I find a reference that adding *HTML attributes* is expected to make something work. The examples in the Readme do *not* affect HTML attributes, they are regular component properties that are used by the addon provided mixin. So I do not see how adding support for these attributes will make this work, without actually applying the mixin. Or do I miss something?
* Assuming adding the attributes would work, I still have a problem with that, as these are not attributes defined by the HTML spec (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#Attributes). And I would be hesitant to add specific code just for a single addon. That look like opening a can of worms...
* As mentioned on Discord already - regarding adding HTML attributes - the way forward with the upcoming v3 release would be to use angle bracket invocation (ABI), which would allow to apply any attributes without having to explicitly support them using the old-fashioned `attributeBindings`. In this specific case, that should even work now with v2.x of the addon. More than that, I expect we will deprecate all of these properties, that are used to actually set *attributes* only sooner than later in favor of ABI.

For these reasons I will have to close this. Feel free to speak out if you disagree!

Redsandro:
Thank you for the in-depth response as to why this was rejected. I hope no one notices I mistook html attributes for component attributes. Don't tell anyone.

Issue #822, opened 2019-06-04
Automatically disable button while in pending state


simonihmig:
* add a property `disableWhilePending`, for compatibility being `false` by default
* When `true`, will disable the button while being in pending state, i.e. promise returned from action is not settled
* In v3.0.0 set `disableWhilePending` to `true` by default?

Related to #638. With a button yielded by a form, that would auto-disable the button while the form is submitting.

@jelhan thoughts?

jelhan:
Would be a helpful addition IMO.

> In v3.0.0 set `disableWhilePending` to `true` by default?

I think an alias to `preventConcurrency` would be more helpful as for most use cases the button should be disabled if `preventConcurrency` is `true`. Maybe we want to consider toggling that default in v3.0.0?

simonihmig:
> I think an alias to preventConcurrency would be more helpful

Yes, but this is describing the case where the button is yielded by the form, so when yielding we can override `disableWhilePending` with whatever makes sense in the context of a `<BsForm>`. For a plain `<BsButton>` the default could still be set to `true` I think? (again, in v3)

jelhan:
> > I think an alias to preventConcurrency would be more helpful
> 
> Yes, but this is describing the case where the button is yielded by the form, so when yielding we can override `disableWhilePending` with whatever makes sense in the context of a `<BsForm>`. For a plain `<BsButton>` the default could still be set to `true` I think? (again, in v3)

Maybe I'm missing something but why shouldn't this be different for a plain `<BsButton>`? `<BsButton>` also has a [`preventConcurrency` option](https://www.ember-bootstrap.com/api/classes/Components.Button.html#property_preventConcurrency). I don't see use cases for `<BsButton @preventConcurrency={{true}} @disableWhilePending={{false}}>` or `<BsButton @preventConcurrency={{false}} @disableWhilePending={{true}}>`.

I'm especially worried about edge cases if `preventConcurrency` defaults to `false` and `disableWhilePending` defaults to `true` and the defaults are not overriden. There might be timing issues in that case which still allows `onClick` action to be fired concurrently. At least I'm worried about the `disabled` attribute of the button not being updated yet by Glimmer VM but a second click event being fire. Not sure if this is prevented by some browser internals but we wouldn't have to worry about it, if both of them are in sync as long as consumer doesn't pass in both.

[...]

PR #834, opened 2019-06-22
Form's and Button's preventConcurrency defaults to true

jelhan:
As discussed in #822. Does not touch `disabled` property of `<BsButton>` yet.

This is a breaking change.

### DIFF
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -326,9 +326,6 @@ module('Integration | Component | bs-button', function(hooks) {
   });
 
   test('preventConcurrency=false allows onClick action to be fired concurrently', async function(assert) {
-    // assertions are done implicitly by awaiting the expected result
-    assert.expect(0);
-
     let deferredClickAction = defer();
     let clickActionExecutionCount = 0;
 
@@ -338,10 +335,11 @@ module('Integration | Component | bs-button', function(hooks) {
     });
     await render(hbs`{{bs-button onClick=clickAction preventConcurrency=false}}`);
 
-    click('button');
-    await waitUntil(() => clickActionExecutionCount === 1);
-    click('button');
-    await waitUntil(() => clickActionExecutionCount === 2);
+    await click('button');
+    assert.equal(clickActionExecutionCount, 1);
+
+    await click('button');
+    assert.equal(clickActionExecutionCount, 2);
 
     deferredClickAction.resolve();
   });
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -298,7 +298,7 @@ module('Integration | Component | bs-button', function(hooks) {
     assert.ok(parentClick.called);
   });
 
-  test('preventConcurrency=true prevents onClick action to be fired concurrently', async function(assert) {
+  test('prevents onClick action to be fired concurrently', async function(assert) {
     let deferredClickAction = defer();
     let clickActionHasBeenExecuted = false;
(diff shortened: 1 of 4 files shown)


tests/integration/components/bs-button-test.js
```diff
@@ -324,4 +324,25 @@ module('Integration | Component | bs-button', function(hooks) {
...
+    let clickActionExecutionCount = 0;
+
+    this.set('clickAction', () => {
+      clickActionExecutionCount++;
+      return deferredClickAction.promise;
+    });
+    await render(hbs`{{bs-button onClick=clickAction preventConcurrency=false}}`);
+
+    click('button');
+    await waitUntil(() => clickActionExecutionCount === 1);
```
simonihmig: Do we really have to `waitUntil` here? Isn't the action handled synchronously, so we can just `assert.equal(clickActionExecutionCount, 1)`?

tests/integration/components/bs-form-test.js
```diff
@@ -794,6 +791,35 @@ module('Integration | Component | bs-form', function(hooks) {
...
+    });
+    this.set('beforeAction', beforeActionFake);
+    this.set('validate', validateFake);
+    await render(hbs`
+      {{#bs-form preventConcurrency=false onSubmit=submitAction onBefore=beforeAction validate=validate hasValidator=true}}
+      {{/bs-form}}
+    `);
+
+    triggerEvent('form', 'submit');
+    await waitUntil(() => submitActionExecutionCounter === 1);
```
simonihmig: same here

tests/integration/components/bs-button-test.js
jelhan: As far as I've learned from my tests the action is not executed in sync. I guess this is related to runloop. Awaiting next runloop has worked for most scenarios but wasn't a stable pattern. I've set up a CodeSandbox to play around with that behavior: https://codesandbox.io/s/my-app-c276f?fontsize=14&initialpath=%2Ftests&module=%2Ftests%2Fintegration%2Fcomponents%2Fmy-component-test.js

Hope the sandbox is working fine. I'm not that used to CodeSandbox but Ember Twiddle feels so outdated.

simonihmig: I think the problem is that `await` is missing for the `click()` call!

Forked your sandbox, with that additional `await` the sync test is also working: https://codesandbox.io/s/my-app-3ptr6

jelhan: `await click` is working for the very basic example in CodeSandox but isn't working for testing pending promises. `await click` won't resolve before the application is in settled state:

> Settled generally means that there are no pending timers, no pending waiters, no pending AJAX requests, and no current run loop. However, new settledness metrics may be added and used as they become available.

Or to put it the other way around: `await click()` would await for the promise returned by `onClick` action to resolve. But the test is about testing how the component behaves while this promise is pending. So this won't work out but result in a timeout.

simonihmig: > Or to put it the other way around: await click() would await for the promise returned by onClick action to resolve

Are you sure? AFAIK `settled()` does not include arbitrary promises. 

> Settled generally means that there are no pending timers, no pending waiters, no pending AJAX requests, and no current run loop. However, new settledness metrics may be added and used as they become available.

The pending promise is not included in that enumeration, is it?

jelhan: :man_facepalming: Mixed up testing "real world" interim state, which includes network requests and these mocked ones. Replaced `waitUntil` with `await click()`.

PR #842, opened 2019-06-25
Button is disabled by default if in pending state

### LINKED ISSUE #822: Automatically disable button while in pending state

jelhan:
Closes #822

### DIFF
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -1,9 +1,10 @@
 import { run } from '@ember/runloop';
 import { defer, reject, resolve } from 'rsvp';
-import { module } from 'qunit';
+import { module, skip } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { find, render, click, settled, waitUntil } from '@ember/test-helpers';
 import { test, testBS4, defaultButtonClass } from '../../helpers/bootstrap-test';
+import { gte } from 'ember-compatibility-helpers';
 
 import hbs from 'htmlbars-inline-precompile';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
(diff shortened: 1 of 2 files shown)


tests/integration/components/bs-button-test.js
```diff
@@ -199,6 +199,65 @@ module('Integration | Component | bs-button', function(hooks) {
...
+      return deferredClickAction.promise;
+    });
+
+    await render(hbs`<BsButton @onClick={{clickAction}} disabled={{false}} />`);
+    await click('button');
+    assert.dom('button').isNotDisabled();
+
+    deferredClickAction.resolve();
+    await settled();
+  });
```
jelhan: This test case is mainly about preventing a regression when we refactor to outer HTML syntax and spread attributes. Basically these test should ensure that we are applying default value for `disabled` before `...attributes` and not after.

simonihmig: Nice!

But this test is actually failing for Ember 2.18, pretty sure due to a bug in the polyfill: https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/issues/69

Because of this I skipped some tests for Ember <3.4 (i.e. versions w/o native angle bracket support) here: https://github.com/kaliber5/ember-bootstrap/commit/a46ed2e1ea62fd4c451655c14aea6ec0a8520956#diff-fd815e23c0edac5d5ea76999058689dd. We should do the same here.

simonihmig:
@jelhan I pushed that suggested test skipping change to your branch, fyi! 😏 Let's see...

simonihmig:
Finally green - after restarting the BrowserStack for ~5 times 🤨

Thanks @jelhan!

jelhan:
> @jelhan I pushed that suggested test skipping change to your branch, fyi!  Let's see...

Thanks. Actually I had done it locally but forgot to push. :man_facepalming: I guess doing to much things on the side isn't a good pattern...

PR #845, opened 2019-06-29
Make FormElement a tagless component, not extending FormGroup anymore

### LINKED ISSUE #844: Form element w/ Angle Bracket Invocation has HTML attrs applied twice
Applying HTML attributes to a form element V3.0.0-rc.0 applies HTML attributes to the form control (good) as well as the `.form-group` div (bad).

simonihmig:
Extending from FormGroup looks like a bad decision anyway, *using* it seems more appropriate. This also makes FormElement a tagless component, allowing to use `...attributes` without applying HTML attributes twice. Fixes #844

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -13,7 +13,8 @@ import { typeOf, isBlank } from '@ember/utils';
 import { A, isArray } from '@ember/array';
 import { getOwner } from '@ember/application';
 import layout from 'ember-bootstrap/templates/components/bs-form/element';
-import FormGroup from 'ember-bootstrap/components/bs-form/group';
+import Component from '@ember/component';
+import { guidFor } from '@ember/object/internals';
 
 const nonDefaultLayouts = A([
   'checkbox'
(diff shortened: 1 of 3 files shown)


jelhan:
This will break using test selectors on a form element with classic component invocation syntax, won't it?
Consider this template:

```js
{{#bs-form as |form|}}
  {{form.element property="name" label="Name" data-test-form-element-for="name"}}
{{/bs-form}}
```

I expect ember-test-selectors to throw:

> ember-test-selectors could not bind data-test-* properties on ${component} automatically because 
tagName is empty. If you did this intentionally, see https://github.com/simplabs/ember-test-selectors#usage-in-computed-properties for instructions on how to disable this assertion.

I'm not sure if there is a pattern that allows to refactor to outer HTML components without breaking test selectors in classic invocation syntax. Or am I wrong and test selectors supports that case somehow magically?

simonihmig:
> This will break using test selectors on a form element with classic component invocation syntax, won't it?

Yes I think so. 

> I'm not sure if there is a pattern that allows to refactor to outer HTML components without breaking test selectors in classic invocation syntax.

Not aware of any. But I think it nevertheless makes sense to migrate to tagless/outerhtml components, to prevent such unexpected problems with HTML attrs and splattributes, as well as pave the was to eventually migrate to Glimmer components.

I will close this however, due to the problems mentioned in #844, which make it apparent that most probably this is not the solution we want!

Issue #826, opened 2019-06-13
Deprecate properties whose only purpose is supporting HTML attributes


simonihmig:
After v3 with 100% angle bracket invocation support has landed, we should eventually show a deprecation warning when using any of those props that are just used for setting *attributes* (via `attributeBindings`), and eliminate all those `attributeBindings` w/ v4.

jelhan:
Started to investigate how we could address this one as we have a lot of such attribute bindings. Seems like we can create a `deprecatedAttributeBindings` decorator to ease the process for many of them. Hope I can get an initial PoC ready by the end of the week.

While working on it, I noticed that we can't deprecate attribute bindings yet if we need to yield the component with initial HTML attributes. Setting initial HTML attributes is simply not supported yet by `{{component}}` helper: https://github.com/emberjs/rfcs/issues/497 :disappointed: Not sure how many attribute bindings are affected by that one.

simonihmig:
Closed by #933

PR #847, opened 2019-07-03
FormElement: revert passing HTML attributes to its control component w/ angle brackets

### LINKED ISSUE #844: Form element w/ Angle Bracket Invocation has HTML attrs applied twice

simonihmig:
As we need to also apply HTML attributes to the form elements `.form.group`, e.g. for test selectors, while not having those attributes applied twice, we don't pass them to the control component anymore. To apply HTML attributes to the control, they should be passed to the yielded component:

```hbs
  <el.control placeholder="Enter your name" />
</form.element>
```

Fixes #844

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -179,7 +179,24 @@ const nonDefaultLayouts = A([
 
   ### HTML attributes
 
-  To set HTML attributes on the control element provided by this component, set them as properties of this component:
+  To set HTML attributes on the control element provided by this component when using the modern angle bracket invocation,
+  you can pass them to the yielded `control` component:
+
+  ```hbs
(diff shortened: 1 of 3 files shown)


simonihmig:
This basically implements the suggestion: https://github.com/kaliber5/ember-bootstrap/issues/844#issuecomment-506981624.

Setting control attributes as properties (on the form element) is still supported, although it remains to be seen if we should deprecate this eventually as part of #826, as passing attributes as properties (using that weird `attributeBindings`) is not very Octane-ish.

simonihmig:
> Restarted Fastboot test. Not sure why ember-canary fails. Doesn't seem to be related to this PR.

Yeah, bot scenarios fail also on master. Will look into that separately. The canary errors also are visible in other repos, so probably something in canary itself!?

PR #851, opened 2019-07-06
Render Modals, Tooltips and Popovers in a wormhole even in tests

simonihmig:
Previously those components would fallback to `renderInPlace=true` if the wormhole destination is not found, which was the case in tests. Now they will render wormholed to the test container element, so they work just as they would in the real app. This might break some tests, if they make assumptions about the place in DOM of those components in tests.


addon/components/base/bs-contextual-help.js
```diff
@@ -265,7 +278,7 @@ let component = Component.extend(TransitionSupport, {
    * @private
    */
   _renderInPlace: computed('renderInPlace', function() {
-    return this.get('renderInPlace') || typeof document === 'undefined' || !document.getElementById('ember-bootstrap-wormhole');
+    return this.get('renderInPlace') || !this.destinationElement;
```
jelhan: Should we omit a warning if destination element is missing?

simonihmig: Makes sense!

Issue #858, opened 2019-07-17
Bootstrap 4 - Glyphicons migration?


absemetov:
'importBootstrapFont': true, dont help

jelhan:
Ember Bootstrap does not provide a way to import Glyphicons with Bootstrap 4. There are different migration strategies. You could find an overview in [this question on StackOverflow](https://stackoverflow.com/questions/32612690/bootstrap-4-glyphicons-migration). Personally I would recommend refactoring to Fontawesome. They provide an officially support [ember addon](https://github.com/FortAwesome/ember-fontawesome). Importing Glyphicons manually as discussed in that StackOverflow question might not be to complex but you may face license issues.

absemetov:
@jelhan Thanks!

PR #859, opened 2019-07-17
Warn if configured to import Glyphicon but Boostrap 4 is used

jelhan:
Adds a warning if Bootstrap 4 is used and `importBootstrapFont` is `true`. Also removes `importBootstrapFont` from configuration injected by default blueprint if Bootstrap 4 is used. This should help to reduce confusion about that config option. It's related to #858.


simonihmig:
Thanks!

PR #861, opened 2019-07-19
Change nav.item link support to match Ember's `<LinkTo>`

### LINKED ISSUE #852: nav.item doesn't passthrough query params v3.0.0-rc.0
With this template: 
```
<nav.item @linkTo={{array "lessons" (query-params filter="upcoming")}}>
  Upcoming Lessons
</nav.item>
```
I receive an error:
`More context objects were passed than there are dynamic segments`
It's probably a good idea to not use positional params in this case since API for angel bracket LinkTo component has changed so each paras is passed as individual arguments.

simonihmig:
This removes the `@linkTo` property that was introduced in 3.0.0-rc.0 in favor of `@route`, `@model`/`@models` and `@query` to match the API of Ember's angle bracket `<LinkTo>` component (RFC459).

Fixes #852

*Waiting for a new release of `ember-angle-bracket-invocation-polyfill` (needs https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/pull/72) before this can be merged,*

### DIFF
--- a/addon/components/base/bs-nav/item.js
+++ b/addon/components/base/bs-nav/item.js
@@ -1,5 +1,5 @@
 import Component from '@ember/component';
-import { observer, computed } from '@ember/object';
+import { observer } from '@ember/object';
 import { filter, filterBy, gt } from '@ember/object/computed';
 import { scheduleOnce } from '@ember/runloop';
 import LinkComponent from '@ember/routing/link-component';
@@ -7,7 +7,7 @@ import layout from 'ember-bootstrap/templates/components/bs-nav/item';
 import ComponentParent from 'ember-bootstrap/mixins/component-parent';
 import overrideableCP from '../../../utils/overrideable-cp';
 import { inject as service } from '@ember/service';
-import { isArray } from '@ember/array';
+import { assert } from '@ember/debug';
 
 /**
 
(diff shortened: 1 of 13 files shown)


addon/components/base/bs-nav/item.js
```diff
@@ -64,10 +85,29 @@ export default Component.extend(ComponentParent, {
    * @type boolean
    * @public
    */
-  active: overrideableCP('_active', 'linkToParams.[]', 'router.currentURL', function() {
-    let params = this.get('linkToParams');
+  active: overrideableCP('_active', 'route', 'model', 'models', 'query', 'router.currentURL', function() {
+    let { route, model, models, query } = this.getProperties('route', 'model', 'models', 'query');
+    let params = [];
+
+    assert('You cannot pass both `@model` and `@models` to a nav item component!', !model || !models);
```
jelhan: I would not have expected this assertion to be part of the `active` computed property. It's a general requirement not only related to active state of a link. Did you considered moving it to `init`?

simonihmig: Makes sense!

simonihmig:
> I would only add an additional assertion that @model, @models and @query arguments are only passed @route argument is present.

AFAIK you can have an empty `@route` in `<LinkTo>`, which defaults to the current route.

jelhan:
> > I would only add an additional assertion that @model, @models and @query arguments are only passed @route argument is present.
> 
> AFAIK you can have an empty `@route` in `<LinkTo>`, which defaults to the current route.

That maybe supported by `<LinkTo>` but is not by current implementation of `<nav.item>`. `{{#if @route}}` is `false` for an empty string so `<nav.item @route="" />` won't render a `<LinkTo>`. I thought that was by intent.

simonihmig:
Ah, right, good catch! 
I should probably add a CP like `hasLink: or('route', 'model', 'models', 'query')` and use that as `{{#if this.hasLink}}`

jelhan:
> I should probably add a CP like `hasLink: or('route', 'model', 'models', 'query')` and use that as `{{#if this.hasLink}}`

Are you sure that passing `@model` or `@models` but no `@route` is supported? I wasn't able to find any integration test for that one. So maybe it should be like:

- Render a `<LinkTo>` if `@route` or `@query` is passed.
- Throw an assertion if `@model` or `@models` is passed but `@route` is not.

Some other arguments for `<LinkTo>` are also not suppoted, e.g. `current-when`, `activeClass`, `disabledWhen`. I haven't faced the need to use yet so that's fine for me. But maybe it should be documented?

simonihmig:
> Are you sure that passing @model or @models but no @route is supported? I wasn't able to find any integration test for that one. 

What are you referring to? Supported in e-b, or in Ember itself? Ember's `<LinkTo>` seems to support an empty `@route`:
* https://github.com/emberjs/ember.js/blob/91656e1154afe39791514cbe937ada954eab8d11/packages/%40ember/-internals/glimmer/tests/integration/components/link-to/rendering-angle-test.js#L23
* https://github.com/rwjblue/ember-angle-bracket-invocation-polyfill/pull/72#discussion_r303253925

jelhan:
> Ember's `<LinkTo>` seems to support an empty `@route`:

Didn't saw that assertion and that comment. Just had a look for an integration tests for something like `<LinkTo @model={{hash}} />` or `<LinkTo @models={{array}} />` but wasn't able to find one. Maybe I just missed it or the test coverage isn't that good. But nevertheless your suggestion of `hasLink: or('route', 'model', 'models', 'query')` seems to be correct.

PR #862, opened 2019-07-26
BsForm should not catch errors thrown by onSubmit action

jelhan:
This may be considered a breaking change. So would be great to include it in 3.0 release. We had discussed that one before in #789.

### DIFF
--- a/tests/integration/components/bs-form-test.js
+++ b/tests/integration/components/bs-form-test.js
@@ -19,6 +19,7 @@ import { defer } from 'rsvp';
 import { next, run } from '@ember/runloop';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
 import RSVP from 'rsvp';
+/* global Ember */
 
 const nextRunloop = function() {
   return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


tests/integration/components/bs-form-test.js
```diff
@@ -682,7 +699,15 @@ module('Integration | Component | bs-form', function(hooks) {
...
 
   test('Yielded #isRejected is true if onSubmit action rejects', async function(assert) {
-    this.actions.submit = this.fake.rejects();
+    // tests fail by default on unhandled errors
+    let expectedError = new Error();
+    Ember.onerror = (error) => {
+      if (error !== expectedError) {
+        throw error;
+      }
+    };
```
simonihmig: Shoudn't we restore the original `Ember.onerror` handler after the test? Or is it undefined by default?

jelhan: If I didn't missed something `Ember.onerror` is initialized with `undefined` for each test. I'm not sure if this is documented behavior and should be treated as public API. Could reset it for clarity if you prefer that.

PR #864, opened 2019-08-05
Refactored mixins

simonihmig:
Removed almost all mixins (except for parent/child mixins which contain more logic) in favor of CP macros:
* most mixins were basically providing a simple CP anyway
* classes w/ mixins are less readable and more "magic", as the source of props is obfuscated
* an upcoming refactor to native classes will benefit from less mixins


addon/utils/cp/uses-transition.js
```diff
@@ -0,0 +1,7 @@
+import { computed } from '@ember/object';
+
+export default function usesTransition() {
+  return computed('fade', 'fastboot.isFastBoot', function() {
+    return !this.get('fastboot.isFastBoot') && this.get('fade');
+  })
```
jelhan: This is the only place where `fastboot` CP is consumed isn't it? Do we need to observe the `isFastBoot` property of `fastboot` service? I would expect that it can't change. Either the application runs in Fastboot or not. If that's the case we may could drop the `fastboot` cp and inline it's logic here. Would remove the dependencies between the CPs which could easily cause bugs.

addon/utils/cp/size-class.js
```diff
@@ -0,0 +1,12 @@
+import { computed } from '@ember/object';
+import { isBlank } from '@ember/utils';
+import { assert } from '@ember/debug';
+
+export default function sizeClass(prefix) {
+  assert('You have to provide a prefix for sizeClass', prefix);
+
+  return computed('size', function() {
+    let size = this.get('size');
+    return isBlank(size) ? null : `${prefix}-${size}`;
```
jelhan: We may want to add an assertion that the value of size is valid. At least that it's a string.

addon/utils/cp/form-validation-class.js
```diff
@@ -0,0 +1,15 @@
...
+  return computed('validationType', function() {
+    let validationType = this.get('validationType');
+    switch (validationType) {
+      case 'error':
+        return 'is-invalid';
+      case 'success':
+        return 'is-valid';
+      case 'warning':
+        return 'is-warning'; // not officially supported in BS4 :(
+    }
```
jelhan: We may want to add an assertion as default case to ensure that `validationType` is supported.

addon/utils/cp/type-class.js
```diff
@@ -0,0 +1,15 @@
+import { computed } from '@ember/object';
+import { assert } from '@ember/debug';
+
+export default function typeClass(prefix) {
+  assert('You have to provide a prefix for typeClass', prefix);
+
+  return computed('type', function() {
+    let type = this.get('type') || 'default';
```
jelhan: We may want to add an assertion that `type` is a non-empty string.

addon/utils/cp/uses-transition.js
simonihmig: > This is the only place where fastboot CP is consumed isn't it? 

No, it's also used in `bs-modal`.

addon/utils/cp/type-class.js
simonihmig: Why, if we have a default value here? (literally "default" 😉)

addon/utils/cp/form-validation-class.js
simonihmig: But an empty value (`null`) is valid, in which case no class should be applied.

addon/utils/cp/size-class.js
simonihmig: ok.

simonihmig:
> IMO it would be better if CPs do only depend on properties which are passed in as arguments.

I agree. I can make all dependent keys explicit in  the CP macro call...

addon/utils/cp/type-class.js
jelhan: I was talking about `type` and not `this.get('type')`. Since `type` is such a general term and current implementation hides the dependency, this would at least catch some issues (e.g. `this.get('type')` being a number, array or object). We could even assert that it's a supported string but I'm not sure if that wouldn't be too much.

addon/utils/cp/form-validation-class.js
jelhan: Missed that one.

addon/utils/cp/uses-transition.js
jelhan: I think an `isFastboot` utility function would serve both cases and reduce the magic:

```js
export default function(context) {
  let owner = getOwner(context);
  let fastbootService = owner.lookup('service:fastboot');
  return fastbootService ? fastbootService.get('isFastboot') : false;
}
```

PR #871, opened 2019-08-14
Workaround nav.item not applying `active` class correctly due to Ember 3.12-beta bug

simonihmig:
Was caused by the CP not recomputing when `router.currentURL` changes.

### DIFF
--- a/addon/components/base/bs-nav/item.js
+++ b/addon/components/base/bs-nav/item.js
@@ -105,7 +105,11 @@ export default Component.extend(ComponentParent, {
       params.push({ queryParams: true, ...query});
     }
 
-    return params.length ? this.get('router').isActive(...params) : this.get('_active');
+    // workaround for https://github.com/emberjs/ember.js/issues/18264
+    // @todo remove once this is resolved
+    this.get('router.currentURL');
+
+    return params.length > 0 ? this.get('router').isActive(...params) : this.get('_active');
   }),
   _active: false,
 


addon/components/base/bs-nav/item.js
```diff
@@ -105,7 +105,7 @@ export default Component.extend(ComponentParent, {
       params.push({ queryParams: true, ...query});
     }
 
-    return params.length ? this.get('router').isActive(...params) : this.get('_active');
+    return params.length > 0 ? this.get('router.currentURL') && this.get('router').isActive(...params) : this.get('_active');
```
jelhan: If I got it the bug report at emberjs repo right, we are only consuming `router.currentURL` to ensure that CP is recomputed. This should not change any of the logic here, isn't it? If so, I would move that one to a separate line and add a inline comment. Otherwise it's easily removed by next refactor.

simonihmig:
> Why not wait for a bugfix upstream?

Yes, I think we should do that, i.e. not merge this. I created the PR to show the workaround, and for the case that this new "behavior" is somehow not considered a bug. But after the bug has been (kind of) confirmed, we should probably only merge this as a workaround in case the upstream bug fix is not ready for the next stable Ember release.

addon/components/base/bs-nav/item.js
simonihmig: done

simonihmig:
As the upstream bug is not fixed yet, I am going to merge this to get our builds green again. As commented in the code, we should remove this again when fixed in Ember.

PR #872, opened 2019-08-14
Fix form validation for Ember 3.12-beta

simonihmig:
The current form of resetting `showOwnValidation` in form.element by passing `showAllValidations=false` is quite brittle, as it relies
on `showAllValidations=false` being passed to form.element before it is set to `undefined` again. It seems in Ember 3.12-beta the timing is different, i.e. form.element only receives `undefined`.

This is a workaround to fix the problem short-term, but probably the architecture has to be rethought.


jelhan:
> This is a workaround to fix the problem short-term, but probably the architecture has to be rethought.

:100: I think current architecture will also not support https://github.com/kaliber5/ember-bootstrap/issues/364.

PR #875, opened 2019-08-15
Refactor FastBoot tests to use ember-cli-fastboot-testing

simonihmig:
Previous tests based on `ember-fastboot-addon-tests` were very slow and flaky at times, due to the extensive app setup and dependency installations done by `ember-cli-addon-tests`.

These (smoke) tests now reuse the existing component pages of the dummy app, rather than separate templates.


simonihmig:
Some tests are failing due to CSP violations caused by ember-power-select dropdowns used in the dummy app (`style` attribute). 

@jelhan any idea how to ignore these specific violations (which we are not responsible for), without compromising the test strictness for our own component integration tests? (i.e. not allowing `unsafe-inline`)

Btw, by coincidence when looking for a solution, I stumbled upon your PR https://github.com/rwjblue/ember-cli-content-security-policy/pull/91 that has just been merged it seems. Great work! Looks like it was inspired by your previous work here!? 😄

simonihmig:
> Some tests are failing due to CSP violations caused by ember-power-select dropdowns used in the dummy app (style attribute).

Hm, that is weird. So they did fail for me running locally, but they don't do so in CI!? 🤔

Also, need to adapt the tests when running for Bootstrap 3 (as they run now with every ember-try scenario as part of the regular test suite)...

jelhan:
> > Some tests are failing due to CSP violations caused by ember-power-select dropdowns used in the dummy app (style attribute).
> 
> Hm, that is weird. So they did fail for me running locally, but they don't do so in CI!?

This is caused by a configuration error. Should be fixed by #876.

simonihmig:
> any idea how to ignore these specific violations (which we are not responsible for)

I am now specifically ignoring violation events coming from power-select/basic-dropdown here: 
https://github.com/kaliber5/ember-bootstrap/pull/875/files#diff-8f60a6be7648f7eda8ce189a29741c5eR12

Fixed/updated a few more things, this should be ready now! and green! 🎉

jelhan:
> I am now specifically ignoring violation events coming from power-select/basic-dropdown here:
> https://github.com/kaliber5/ember-bootstrap/pull/875/files#diff-8f60a6be7648f7eda8ce189a29741c5eR12

Oh sorry, totally forgot about that one. I was planing to fix it upstream but didn't head the time yet. Great that you found a solution.

PR #877, opened 2019-08-16
support hash helper used as form model

jelhan:
I'm not sure why, but if yielding `{{@model}}` the model objects used in `<BsForm>` and `<FormElement>` are not the same. I guess `{{@model}}` yields the hash helper but not it's result.


jelhan:
> That's weird indeed. Looks more like a bug in Ember, doesn't it? If so let's create an issue there, shouldn't we?

Just opened https://github.com/emberjs/ember.js/issues/18279 after I tried to clarify in discord without success.

PR #882, opened 2019-08-20
yield model from <BsForm>

jelhan:
Example use case:

```hbs
<BsForm @model={{hash}} as |form|>
  <form.element @controlType="date" @property="start" />
  <form.element @controlType="date" @property="end" as |el|>
   <el.control min={{form.model.start}} />
  </form.element>
</BsForm>
```

### DIFF
--- a/addon/templates/components/common/bs-form.hbs
+++ b/addon/templates/components/common/bs-form.hbs
@@ -14,6 +14,7 @@
     isSubmitting=this.isSubmitting
     isSubmitted=this.isSubmitted
     isRejected=this.isRejected
+    model=this.model
     resetSubmissionState=(action "resetSubmissionState")
     submit=(action "submit")
   )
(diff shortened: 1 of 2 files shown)


jelhan:
I'm not sure how to document this best. :confused: @simonihmig, do you have any suggestions?

simonihmig:
I am not against adding this, but FWIW I think you could easily workaround this using the `let` helper:

```hbs
{{#let (hash) as |model|}}
<BsForm @model={{model}} as |form|>
  <form.element @controlType="date" @property="start" />
  <form.element @controlType="date" @property="end" as |el|>
   <el.control min={{model.start}} />
  </form.element>
</BsForm>
{{/let}}
```

> I'm not sure how to document this best. 😕 @simonihmig, do you have any suggestions?

Nothing better than mentioning it in the API docs somewhere here: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-form.js

We could explicitly list all the yielded values, including the yielded components (`group` and `element`). Eventually we will start to yield more, like `button`...

jelhan:
> I am not against adding this, but FWIW I think you could easily workaround this using the `let` helper:

You are right. I totally missed that obvious solution. I think we should not unnecessary bloat the API surface.

Issue #881, opened 2019-08-20
Component module not found when using with ember-cli v3.13.0 with ember-bootstrap


veebs:
---

Output from `ember bootstrap:info`:

```
PS C:\dev\svu-ui-poc\ember-app> ember bootstrap:info
Npm packages:
ember-bootstrap: ^3.0.0-rc.2 -> 3.0.0-rc.2
ember-cli: github:ember-cli/ember-cli#a07650e8465fc8abe35a4a6f269b0f88d047ec58 -> 3.13.0-beta.1
bootstrap: ^4.3.1 -> 4.3.1
bootstrap-sass: n/a
ember-cli-sass: n/a
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 4
importBootstrapCSS: true
```

Components are being registered with ember as `ember-bootstrap/templates/common/components/bs-alert` rather than  `ember-bootstrap/templates/components/bs-alert`.

This is because the `treeForAddonTemplates` hook which is no longer invoked as of [#8633.](https://github.com/ember-cli/ember-cli/pull/8633)

See discussions [here](https://github.com/ember-cli/ember-cli/issues/8693)

The end result is the template folder does not get a clean up and contains the bs3, bs4 and common folders.

![image](https://user-images.githubusercontent.com/367073/63337365-37c0ea80-c384-11e9-9a6c-2210d27dd20c.png)

A quick fix is to move the code form `treeForAddonTemplates()`  to `treeForAddonTemplates()` in index.js.

```javascript
  treeForAddon(tree) {
    let bsVersion = this.getBootstrapVersion();
    let otherBsVersion = this.getOtherBootstrapVersion();
    let componentsPath = 'components/';

    tree = this.debugTree(tree, 'addon-tree:input');
    tree = mv(tree, `${componentsPath}bs${bsVersion}/`, componentsPath);
    tree = rm(tree, `${componentsPath}bs${otherBsVersion}/**/*`);

    tree = this.debugTree(tree, 'addon-tree:bootstrap-version');
    tree = this.filterComponents(tree);
    tree = this.debugTree(tree, 'addon-tree:tree-shaken');

    // START - Moved from treeForAddonTemplates()
    let templatePath = 'templates/components/';

    tree = this.debugTree(tree, 'addon-templates-tree:input');
    tree = mv(tree, `${templatePath}common/`, templatePath);
[...]

PR #883, opened 2019-08-22
Workaround for Ember CLI 3.13 bug (Depreciated/Removed treeForAddonTemplates)

veebs:
See https://github.com/kaliber5/ember-bootstrap/issues/881

### DIFF
--- a/index.js
+++ b/index.js
@@ -234,16 +234,31 @@ module.exports = {
     let componentsPath = 'components/';
     let templatePath = 'templates/components/';
 
-    // Merge bs3 or bs4 component files
     tree = this.debugTree(tree, 'addon-tree:input');
     tree = mv(tree, `${componentsPath}bs${bsVersion}/`, componentsPath);
     tree = rm(tree, `${componentsPath}bs${otherBsVersion}/**/*`);
 
+    tree = mv(tree, `${templatePath}common/`, templatePath);
+    tree = mv(tree, `${templatePath}bs${bsVersion}/`, templatePath);
+    tree = rm(tree, `${templatePath}bs${otherBsVersion}/**/*`);
+    
     tree = this.debugTree(tree, 'addon-tree:bootstrap-version');
     tree = this.filterComponents(tree);
     tree = this.debugTree(tree, 'addon-tree:tree-shaken');
+  
+    return this._super.treeForAddon.call(this, tree);
+  },
 
-    // Merge bs3 or bs4 template files
+  /*
+    This method is depreciated by ember-cli when building other apps
+    but it is still called when building ember-bootstrap
+    https://github.com/kaliber5/ember-bootstrap/pull/883
+  */
+  treeForAddonTemplates(tree) {
+    let bsVersion = this.getBootstrapVersion();
+    let otherBsVersion = this.getOtherBootstrapVersion();
+    let templatePath = 'components/';
+
     tree = this.debugTree(tree, 'addon-templates-tree:input');
     tree = mv(tree, `${templatePath}common/`, templatePath);
     tree = mv(tree, `${templatePath}bs${bsVersion}/`, templatePath);
@@ -253,7 +268,7 @@ module.exports = {
     tree = this.filterComponents(tree);
     tree = this.debugTree(tree, 'addon-templates-tree:tree-shaken');
 
-    return this._super.treeForAddon.call(this, tree);
+    return this._super.treeForAddonTemplates.call(this, tree);
   },
 
   contentFor(type, config) {
--- a/index.js
+++ b/index.js
@@ -232,7 +232,9 @@ module.exports = {
     let bsVersion = this.getBootstrapVersion();
     let otherBsVersion = this.getOtherBootstrapVersion();
     let componentsPath = 'components/';
+    let templatePath = 'templates/components/';
 
+    // Merge bs3 or bs4 component files
     tree = this.debugTree(tree, 'addon-tree:input');
     tree = mv(tree, `${componentsPath}bs${bsVersion}/`, componentsPath);
     tree = rm(tree, `${componentsPath}bs${otherBsVersion}/**/*`);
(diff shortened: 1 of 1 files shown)


jelhan:
This seems straight forward but doesn't work. :cry:

> Error: Could not find module `ember-bootstrap/templates/components/bs-navbar` imported from `ember-bootstrap/components/base/bs-navbar`

@veebs Do you have time to investigate what's happening here?

veebs:
I've updated my branch.  Not sure why GitHub not showing it.

veebs:
Finally github updated the PR.

OK found that `treeForAddonTemplates()` is still called when building ember-bootstrap.  However when building your own app that uses ember-bootstrap, it does not call it.  

Not sure why but the work around is to reinstate the `treeForAddonTemplates()` function.

jelhan:
> OK found that `treeForAddonTemplates()` is still called when building ember-bootstrap.

By "building ember-bootstrap" you are referring to dummy app, isn't it?

veebs:
With my PR, I've been able to `ember test` on my fork of ember-bootstrap.

image

I have also replaced `C:\dev\my-ember-app\node_modules\ember-bootstrap` in my ember project that uses ember-bootstrap and can confirm that ember-bootstrap components are registered with the right module paths within `C:\dev\my-ember-app\dist\assets\vendor.js`.

I'm cannot recreate the failure on the travis-ci build.

Is it because I'm using ember-cli 3.13.0-beta.1 ?

simonihmig:
> I'm cannot recreate the failure on the travis-ci build.

After your second commit tests were actually passing! Only the Browserstack scenario was failing, but that is flaky at times, and seems unrelated to your changes.

> Is it because I'm using ember-cli 3.13.0-beta.1 for my app?

Yes, I think so. Travis runs with the Ember CLI version specified in the addon's `package.json`. 

I would still like to confirm if we really have to do the template transformations twice as we do now, once in `treeForAddon` (new) and once in `treeForAddonTemplates` (as before)...

veebs:
I just got back to having a quick look at this issue.

```
PS C:\Windows\system32> cd C:\dev\svu-ui-poc\ember-app
PS C:\dev\my-ember-app\ember-app> ember --version
ember-cli: 3.13.0-beta.1
node: 10.16.3
os: win32 x64

PS C:\dev\my-ember-app\ember-app> cd C:\dev\ember-bootstrap
PS C:\dev\ember-bootstrap> ember --version
ember-cli: 3.11.0
node: 10.16.3
os: win32 x64
PS C:\dev\ember-bootstrap>
```

Maybe the issue is that the ember-cli version for ember-bootstrap is version `3.11.0`.  

Should we change that to `3.13.0-beta.1`?

simonihmig:
> Should we change that to 3.13.0-beta.1?

No! The addon should work with *any* Ember CLI version (starting at 2.18 for the 3.0.0(-rc.x) version). So we should keep testing with the existing version.

> After your second commit tests were actually passing!

But as said before (👆) your changes do work now with that version. See the green PR. The only remaining "issue" from my PoV is this one:

> I would still like to confirm if we really have to do the template transformations twice as we do now, once in treeForAddon (new) and once in treeForAddonTemplates (as before)...

I see [here](https://github.com/ember-cli/ember-cli/issues/8693#issuecomment-525064919) you tried that out already, so idk if there is anything left we could try out. If not, we should still be able to merge as is I guess, even with that bit of code duplication...

Just to make sure (as CI runs with the older Ember CLI version): the current state of this PR work with the latest Ember CLI beta locally, right?

jelhan:
> If not, we should still be able to merge as is I guess, even with that bit of code duplication...

I'm not that sure that merging in this stage would be best option.

1. This is only a work-a-round for an upstream bug. We have a confirmation from Ember CLI team member (rwjblue) that it will be fixed upstream.
2. This is not affecting any stable releases yet.
3. There is the plan to deprecate `treeForAddonTemplates` but it's not yet deprecate. There isn't even a RFC for deprecation.
4. It's likely that there will be a better migration path when `treeForAddonTemplates` is actually deprecated.
5. We don't have any test coverage for any other Ember CLI version than 3.11 and the current implementation seems to rely highly on some internals of Ember CLI.

It's a different story in my opinion if a stable release of Ember CLI will include that bug. But that's not the case yet.

Ember Leaflet also hasn't implemented a work-a-round yet: https://github.com/miguelcobain/ember-leaflet/pull/342

veebs:
Yes - the current PR works locally for me on `3.13.0-beta.1`

Maybe it is best to wait for a full release of ember-cli (rather than the beta).  The issue and my work around is documented for other users.

I'm only using the latest cause I wanted to try out Ember Octane.

Let's review again once Ember 3.13 is out in about 3-4 weeks?

jelhan:
I've manually confirmed that this is working for

- Ember CLI ~3.13.0
- Ember CLI ~3.12.0
- Ember CLI ~3.8.0
- Ember CLI ~3.4.0
- Ember CLI ~2.18.0

with using `<BsButton>` in application template of a new project.

I've also manually compared asset size for Ember CLI ~3.12.0. If I didn't made any mistake this is *decreasing* the asset size from 1.09 MB (233.41 KB gzipped) to 899.67 KB (215.41 KB gzipped). I'm very confused by that one. I would have expected that we may include code twice due to having duplicated logic but I don't have any idea why we might ship less code...

jelhan:
I've tested with a small real-world application. Working fine for that one also with Ember CLI ~3.12.0 and 3.13.0.

I'm seeing the same reduction of asset size for production builds (>10 KB after gzip). While I'm still confused about what is going on there, it's actually amazing. But we should dig deeper into that one _after_ releasing a bug fix.

simonihmig:
Thanks @jelhan for testing, and @veebs for your work. Will merge then and publish another RC!

> I'm seeing the same reduction of asset size for production builds (>10 KB after gzip). While I'm still confused about what is going on there, it's actually amazing. But we should dig deeper into that one after releasing a bug fix.

Agree. https://github.com/kaliber5/ember-cli-bundle-analyzer could come to a rescue! :wink:

PR #890, opened 2019-09-05
Remove module-unification blueprints

dcyriller:
MU support has been disabled in Ember.

Should this be considered as a breaking change?


simonihmig:
> Any objections against including this one in v3?

No! Also I don't really see this as a breaking change, as it will only affect users installing the addon for the first time in an app, not for anyone updating.

@dcyriller thank you!

PR #896, opened 2019-09-16
deps: upgrade ember-maybe-in-element to v0.4.x

jelhan:


jelhan:
All about dependency linting. Ember Power Select v3 requires `ember-maybe-in-element@^0.4.0` which conflicts with the outdated version used here.

mariowilde:
"ember-maybe-in-element": "^0.2.0"  means, that ember-bootstrap should work happily with "0.4.0".
If you install "0.4.0" there should be no conflict.

jelhan:
> "ember-maybe-in-element": "^0.2.0" means, that ember-bootstrap should work happily with "0.4.0".

That's not true. Behavior of caret ranges depends on leading zeros. Let me cite the docs:

> **Caret Ranges ^1.2.3 ^0.2.5 ^0.0.4**
>
> Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.
> 
> Many authors treat a 0.x version as if the x were the major “breaking-change” indicator.
>
> Caret ranges are ideal when an author may make breaking changes between 0.2.4 and 0.3.0 releases, which is a common practice. However, it presumes that there will not be breaking changes between 0.2.4 and 0.2.5. It allows for changes that are presumed to be additive (but non-breaking), according to commonly observed practices.
>
> - ^1.2.3 := >=1.2.3 <2.0.0
> - ^0.2.3 := >=0.2.3 <0.3.0
> - ^0.0.3 := >=0.0.3 <0.0.4
>
> https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

mariowilde:
You are right. My bad. :facepalm:

I'm all for the upgrade then.

simonihmig:
The thing is that `ember-popper` also has a dependency on 0.2.0, so we would still run into dependency linter issues.

I am currently thinking about dropping maybe-in-element (both in e-b and in e-popper), as it does not work with the RFC-proposed public API of in-element and my polyfill, and making it work proved to be unexpectedly hard somehow (see https://github.com/DockYard/ember-maybe-in-element/pull/16)

jelhan:
> The thing is that `ember-popper` also has a dependency on 0.2.0, so we would still run into dependency linter issues.

I've 

> I am currently thinking about dropping maybe-in-element (both in e-b and in e-popper), as it does not work with the RFC-proposed public API of in-element and my polyfill, and making it work proved to be unexpectedly hard somehow (see [DockYard/ember-maybe-in-element#16](https://github.com/DockYard/ember-maybe-in-element/pull/16))

jelhan:
Sorry, pressed the wrong button by accident.

> The thing is that `ember-popper` also has a dependency on 0.2.0, so we would still run into dependency linter issues.

I've also opened a pull request for ember-popper: https://github.com/kybishop/ember-popper/pull/109 It should be safe to do that upgrade cause `0.4.0` should be a re-publish of `0.2.0` accordingly to https://github.com/DockYard/ember-maybe-in-element/pull/16#issuecomment-509412852.

simonihmig:
Thanks @jelhan. Merged and published ember-popper. Can you update our dependency of ember-popper to 0.10.3 here, so we make sure the dependency requirements align?

PR #898, opened 2019-09-26
upgrade ember-power-select and remove resolutions for ember-concurrency

jelhan:
Ember Power Select is only consumed through ember-bootstrap-power-select. This one is updated also. It does not require ember-power-select-blockless anymore.

The resolution for ember-concurrency was added caused by outdated versions used by Ember Power Select long ago in 771d6f3ad9f080dc2858e092524d7667d869ab2b.

This also prevents using an outdated version of Ember Concurrency in tests (and therefore getting wrong test results).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "chai": "^4.0.0",
     "chai-things": "^0.2.0",
-    "ember-bootstrap-power-select": "^1.1.0",
+    "ember-bootstrap-power-select": "^2.0.0",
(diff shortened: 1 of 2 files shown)


jelhan:
Didn't had in mind that Ember Power Select does not support Ember <= 3.11. This breaks tests for Ember 2.18 and 3.4. Will close as we can't do that upgrade right now. We might still want to consider increasing the resolution for `ember-concurrency` to `^1.0.0`.

Noticed that we are missing a test for Ember 3.8, which should also be covered as it's a LTS release.

PR #900, opened 2019-09-26
use ember-concurrency@^1.0.0 in tests

jelhan:
Still need to use yarn's resolutions features as we can not upgrade Ember Power Select (see #898 for details). But we can configure it to use an up-to-date version of ember-concurrency.


simonihmig:
Thanks!

PR #901, opened 2019-09-26
run tests against Firefox in CI

jelhan:
If I didn't missed something the test suite is not run against Firefox at all in CI. It was running tests against Chrome on Travis and IE 11, Safari and Edge on BrowserStack. Seems like Firefox was missed. This PR configures Travis + Testem to run the test suite against Firefox also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,6 @@ jobs:
   - env: EMBER_TRY_SCENARIO=fastboot-tests
   - env: EMBER_TRY_SCENARIO=node-tests
   - env: EMBER_TRY_SCENARIO=ember-default TEST_COMMAND="ember test --launch Firefox"
-  - env: EMBER_TRY_SCENARIO=ember-default BOOTSTRAPVERSION=3 TEST_COMMAND="ember test --launch Firefox"
 
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,8 @@ jobs:
   - env: EMBER_TRY_SCENARIO=ember-beta BOOTSTRAPVERSION=3
   - env: EMBER_TRY_SCENARIO=fastboot-tests
   - env: EMBER_TRY_SCENARIO=node-tests
+  - env: EMBER_TRY_SCENARIO=ember-default TEST_COMMAND="ember test --launch Firefox"
+  - env: EMBER_TRY_SCENARIO=ember-default BOOTSTRAPVERSION=3 TEST_COMMAND="ember test --launch Firefox"
 
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
--- a/tests/integration/components/bs-popover-test.js
+++ b/tests/integration/components/bs-popover-test.js
@@ -10,6 +10,7 @@ import {
 import setupStylesheetSupport from '../../helpers/setup-stylesheet-support';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
 import { gte } from 'ember-compatibility-helpers';
+import { isFirefox } from '../../helpers/user-agent';
 
 module('Integration | Component | bs-popover', function(hooks) {
   setupRenderingTest(hooks);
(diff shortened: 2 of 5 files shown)


simonihmig:
2 tests are failing for popovers, due to some rather minor pixel offsets 🤔

jelhan:
> 2 tests are failing for popovers, due to some rather minor pixel offsets

Yeah. Haven't had any time to dig into that ones. Will try to do it over the weekend.

jelhan:
The tests are reporting an actual styling bug in Firefox. The tooltip is 4px to far right and therefore not fully visible.

This is a screenshot taken in Firefox from the tooltip rendered by test `Integration | Component | bs-popover: should adjust popover arrow`:

firefox

As a reference a screenshot taken from the same test on Chrome:

chrome

The grey overlay is caused by breakpoint.

jelhan:
I traced the 4px difference down to the `data.offsets.popper.left` passed to `offset` function returned by `popperModifiers` here:

https://github.com/kaliber5/ember-bootstrap/blob/12acad1988c17cd9a3aaca74d1835a4fbac5add4/addon/components/base/bs-contextual-help/element.js#L164

If I didn't missed something `Popper` is initialized with the same arguments in Chrome as in Firefox here: https://github.com/kybishop/ember-popper/blob/6710421197e164420a2b0e312106b51ba75d5831/addon/components/ember-popper-base.js#L271

Seems like this is caused by popper itself. Or did I missed something?

simonihmig:
> Seems like this is caused by popper itself. 

Maybe. I am pretty sure we have nothing on our side specifically treats FF differently. Can you maybe compare FF to Chrome with the original Bootstrap, e.g. here: https://getbootstrap.com/docs/4.3/components/popovers/. Anything similar happening there?

It seems this happens when the popover has not enough space to fit into the viewport when centered, so popper moves that around to make it fit, right?

jelhan:
> Maybe. I am pretty sure we have nothing on our side that specifically treats FF differently. Can you maybe compare FF to Chrome with the original Bootstrap, e.g. here: https://getbootstrap.com/docs/4.3/components/popovers/. Anything similar happening there?

Seeing the exact same 4px difference on getbootstrap.com. Will report upstream.

popover-firefox-4px-missing

> It seems this happens when the popover has not enough space to fit into the viewport when centered, so popper moves that around to make it fit, right?

Yes. If I got it right, the test is especially about that case.

jelhan:
Reported the issue upstream and changed the expected values for Firefox in test. Should be ready to merge if test are passing.

tests/integration/components/bs-popover-test.js
```diff
@@ -78,6 +79,13 @@ module('Integration | Component | bs-popover', function(hooks) {
     this.insertCSSRule('#wrapper p { margin-top: 200px }');
 
     let expectedArrowPosition = versionDependent(225, 219);
+
+    if (isFirefox()) {
+      // Popover arrow has a different positioning by 4px on Firefox
+      // https://github.com/twbs/bootstrap/issues/29485
+      expectedArrowPosition = versionDependent(225 - 4, 219 - 4);
```
simonihmig:
```suggestion
      expectedArrowPosition = expectedArrowPosition - 4;
```

This would make it a bit more clear that FF is always 4px off, and makes it less error prone in case the base values (225/219) need to change, e.g. in case of changing the test. Just a very minor suggestion though

tests/integration/components/bs-tooltip-test.js
```diff
@@ -390,6 +391,13 @@ module('Integration | Component | bs-tooltip', function(hooks) {
     this.insertCSSRule('.margin-top { margin-top: 200px; }');
 
     let expectedArrowPosition = versionDependent(155, 150);
+
+    if (isFirefox()) {
+      // Popover arrow has a different positioning by 4px on Firefox
+      // https://github.com/twbs/bootstrap/issues/29485
+      expectedArrowPosition = versionDependent(155 - 4, 150 - 4);
```
simonihmig:
```suggestion
      expectedArrowPosition = expectedArrowPosition - 4;
```

Same here.

.travis.yml
```diff
@@ -70,6 +70,8 @@ jobs:
   - env: EMBER_TRY_SCENARIO=ember-beta BOOTSTRAPVERSION=3
   - env: EMBER_TRY_SCENARIO=fastboot-tests
   - env: EMBER_TRY_SCENARIO=node-tests
+  - env: EMBER_TRY_SCENARIO=ember-default TEST_COMMAND="ember test --launch Firefox"
+  - env: EMBER_TRY_SCENARIO=ember-default BOOTSTRAPVERSION=3 TEST_COMMAND="ember test --launch Firefox"
```
simonihmig: Do you think it's worth it to specifically test BS3 in FF? Sure, it *could* capture a bug, but the chances that it happens only for BS3 and FF are rather low I guess. Boils down to a tradeoff between test coverage and CI running time. 

Just thinking aloud here, have no strong opinions...

jelhan: I guess you are right. Tests for IE11 are also run against BS4 only. I guess that's much more likely to catch a bug.

PR #902, opened 2019-09-27
Workaround for Ember CLI 3.13 bug (Depreciated/Removed treeForAddonTemplates)

simonihmig:
Backport of #883 to `v2` branch


jelhan:
`ember-code-snippet` dropped support for node 6 in a minor release. :disappointed: Already reported here: https://github.com/ef4/ember-code-snippet/pull/60#issuecomment-535496518 I think we need to pin to `2.4.0`.

simonihmig:
Right. I switched Travis to run w/ node 8, but pinning is actually the better solution. Will do that...

Issue #839, opened 2019-06-24
Bootstrap supports dynamic transition durations controlled by CSS now


jelhan:
Since https://github.com/twbs/bootstrap/pull/25662 Bootstrap supports variable transition durations that are controlled by CSS. It looks up the actual `transition-duration` using `window.getComputedStyles()` for the element: https://github.com/twbs/bootstrap/blob/8fa0d3010112dca5dd6dd501173415856001ba8b/js/src/util.js#L92-L114

Ember Bootstrap also supports dynamic transition durations for `<BsAlert>`, `<BsCarousel>`, `<BsModal>`, `<BsTooltip>` but they are controlled using properties on the component not CSS. I think we should align with Bootstrap default behavior again.

Also I'm wondering if the animations are working correctly if `transition-duration` defined in CSS and `transitionDuration` defined as component property are not the same. Didn't investigated it yet.

PR #905, opened 2019-10-03
WIP: support toasts

### LINKED ISSUE #751: Add toast support
Toast support would be awesome!

jelhan:
This implements toasts. It's only provides very basic features so far. But since it includes some new concepts and was stale on my private computer for more than a month, I would like to ship it even if missing some basic features like customization of header and footer in block mode.

We might consider shipping it as an _experimental feature_. We could highlight that it's not covered by our semver commitments. In that case we would keep some flexibility regarding the API as long as we haven't implemented full feature set. I won't expect a breaking change though but you never know.

I will try to finish the implementation soon but shipping a very basic feature set would make it easier for other to help out with some features.

From a technical perspective I would like to highlight `getTransitionDurationFromElement`. This is an implementation of #839. After we have landed it here, we could and should use it for `<BsAlert>`, `<BsCarousel>`, `<BsModal>` and `<BsTooltip>` as well.

Closes #751

### DIFF
--- a/addon/components/bs4/bs-toast.js
+++ b/addon/components/bs4/bs-toast.js
@@ -1,0 +1,315 @@
+import Component from '@ember/component';
+import { layout as templateLayout, tagName } from '@ember-decorators/component';
+import layout from 'ember-bootstrap/templates/components/bs-toast';
+import defaultValue from 'ember-bootstrap/utils/default-decorator';
+import { cancel, later } from '@ember/runloop';
+import { action } from '@ember/object';
+import { getTransitionDurationFromElement } from 'ember-bootstrap/utils/transition';
+import { assert } from '@ember/debug';
+
+const ClassName = {
+  FADE: 'fade',
+  HIDE: 'hide',
+  SHOW: 'show',
+  SHOWING: 'showing'
+}
+
+@templateLayout(layout)
+@tagName('')
+export default class BsToastComponent extends Component {
+  /**
+   * @property animate
+   * @type Boolean
+   * @default true
+   * @public
+   */
+  @defaultValue
+  animate = true;
+
+  /**
+   * Controls if the toast is hidden automatically.
+   *
+   * @property autohide
+   * @type Boolean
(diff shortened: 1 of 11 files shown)


simonihmig:
Thanks @jelhan for working on this! I still need to properly review this, currently lacking time though. Hopefully I can do this within this week...

Tests are failing though for Ember 2.18, have you noticed?

package.json
```diff
@@ -17,6 +17,7 @@
   },
   "devDependencies": {
     "@ember/optional-features": "^0.7.0",
+    "@ember/render-modifiers": "^1.0.1",
```
simonihmig: This must be put to `dependencies`!

package.json
```diff
@@ -51,6 +52,7 @@
     "ember-fetch": "^6.3.1",
     "ember-load-initializers": "^2.0.0",
     "ember-maybe-import-regenerator": "^0.1.6",
+    "ember-on-modifier": "^1.0.0",
```
simonihmig: Same

addon/components/bs4/bs-toast.js
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    // if toast is in showing state the animation must be canceled
+    if (this.state === 'showing') {
+      cancel(this._timers.onShown);
+      this._timers.delete('onShown');
+    }
+
+    if (typeof this.onHide === 'function') {
+      this.onHide();
```
simonihmig: For Modals we allow returning `false` to prevent hiding it, in case the user wants to apply some custom logic. Does it make sense to allow that here also?

jelhan:
I'm going to close this pull request. I didn't found the time to follow up in two years. This initial draft had some architectural issues as highlighted by @simonihmig in his review.

I still would love to add toast support in Ember Bootstrap. But I don't have the time currently.

If someone else wants to follow-up I would recommend to start from scratch. The architectural issues may lead you into the wrong direction otherwise.

PR #920, opened 2019-10-31
Try to mitigate Browserstack tests (IE11) flakiness

simonihmig:


.travis.yml
```diff
@@ -57,7 +57,7 @@ jobs:
   # we recommend new addons test the current and previous LTS
   # as well as latest stable release (bonus points to beta/canary)
   - stage: "Additional Tests"
-    env: EMBER_TRY_SCENARIO=ember-default BABELPOLYFILL=true BEFORE_SCRIPT="ember browserstack:connect" AFTER_SCRIPT="ember browserstack:disconnect" TEST_COMMAND="ember test --filter !FastBoot --config-file testem.browserstack.js"
+    env: EMBER_TRY_SCENARIO=ember-default BABELPOLYFILL=true BEFORE_SCRIPT="ember browserstack:connect" AFTER_SCRIPT="ember browserstack:disconnect" TEST_COMMAND="travis_retry ember test --filter !FastBoot --config-file testem.browserstack.js"
```
jelhan: This would rerun all tests and not only for IE11, isn't it? Would be great if we could rerun IE11 only to speed up tests and not waste that much CI time.

simonihmig: Yes, but I did not want to add Travis-specific stuff into the testem config... However it seems this is not working at all, `travis_retry` seems to be only available for dependency installations... 🤔

PR #921, opened 2019-11-01
Add keyboard control to dropdowns

### LINKED ISSUE #470: Add keyboard support for Dropdowns

simonihmig:
Closes #470

### DIFF
--- a/addon/components/base/bs-dropdown.js
+++ b/addon/components/base/bs-dropdown.js
@@ -1,9 +1,20 @@
 import { classNameBindings, layout as templateLayout } from '@ember-decorators/component';
 import { action, computed } from '@ember/object';
 import Component from '@ember/component';
-import { bind } from '@ember/runloop';
 import layout from 'ember-bootstrap/templates/components/bs-dropdown';
 import defaultValue from 'ember-bootstrap/utils/default-decorator';
+
+const ESCAPE_KEYCODE           = 27 // KeyboardEvent.which value for Escape (Esc) key
+const SPACE_KEYCODE            = 32 // KeyboardEvent.which value for space key
+const TAB_KEYCODE              = 9 // KeyboardEvent.which value for tab key
+const ARROW_UP_KEYCODE         = 38 // KeyboardEvent.which value for up arrow key
+const ARROW_DOWN_KEYCODE       = 40 // KeyboardEvent.which value for down arrow key
+
+const SUPPORTED_KEYCODES = [
+  ESCAPE_KEYCODE,
+  ARROW_DOWN_KEYCODE,
+  ARROW_UP_KEYCODE,
+];
 
 /**
   Bootstrap style [dropdown menus](http://getbootstrap.com/components/#dropdowns), consisting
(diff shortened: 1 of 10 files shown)


simonihmig:
@jelhan do you want to have a look at this, before merging? It could look a bit nicer implementation-wise if we were able to use pure Octane stuff (e.g. modifiers on angle bracket components is not covered by a polyfill), but at least from the user's point of view this should all work now.

addon/components/base/bs-dropdown.js
```diff
@@ -272,61 +283,88 @@ export default class Dropdown extends Component {
...
+        (e.type === 'keyup' && e.which === TAB_KEYCODE && menuElement && !menuElement.contains(target))
+        || (e.type === 'click' && toggleElement && !toggleElement.contains(target) && ((menuElement && !menuElement.contains(target)) || this.get('closeOnMenuClick')))
+      )) {
+      this.closeDropdown();
+    }
+  }
+
+  @action
+  handleKeyEvent(event) {
+    if (/input|textarea/i.test(event.target.tagName)
```
jelhan: IMO `Array.includes()` is easier to read but that's just coding style.

```suggestion
    if (['input', 'textarea'].includes(event.target.tagName)
```

addon/components/base/bs-dropdown.js
```diff
@@ -272,61 +283,88 @@ export default class Dropdown extends Component {
...
+      ? (
+        event.which === SPACE_KEYCODE
+        || event.which !== ESCAPE_KEYCODE && (event.which !== ARROW_DOWN_KEYCODE && event.which !== ARROW_UP_KEYCODE || this.get('menuElement').contains(event.target)))
+      : !SUPPORTED_KEYCODES.includes(event.which)) {
```
jelhan: That's a very complex condition. An inline comment explaining it, would be great.

addon/components/base/bs-dropdown.js
```diff
@@ -272,61 +283,88 @@ export default class Dropdown extends Component {
...
+      return;
     }
+
+    event.preventDefault();
+    event.stopPropagation();
```
jelhan: Is this introducing a breaking change?

addon/components/base/bs-dropdown.js
simonihmig: Yeah, I copy pasted this from bootstrap.js tbh 😆

addon/components/base/bs-dropdown.js
simonihmig: Yeah, some copy pasting basically. And got me confused as well, *multiple times* . during working on it 😂. Will do

addon/components/base/bs-dropdown.js
simonihmig: Yeah, in theory *could* be. But tbh I don't see a way to introduce this *feature* without excluding the possibility to break things for some rare edge cases. Don't know how we should treat this. But honestly if we are very strict about it, anything we introduce could cause things to break in some more or less odd cases. Think of introducing a new private property in a component, but the user has extended from it already and added something with the same name. 🤔

In this case for example opening the dropdown by pressing the space key when the toggle is focused requires you to intercept the event, otherwise the default behaviour of the space key to scroll down the page would happen, which is not what you want here... (some for down key etc.)

Original bootstrap.js does the same (followed their implementation closely), so it is the right thing to do. But yes, could break things. But again I would consider this not a breaking change of our API, but a feature that introduces (potentially breaking) new behaviour. Idk if that's a valid perspective from the point of SemVer, but I think it is a valid one from a practical standpoint!? 😆

jelhan: Agree. But I think we should mention in release notes to give a hint for consumers that are affected by it.

addon/components/base/bs-dropdown.js
jelhan: Sorry, my bad. The regex was case insensitive. `tagName` is uppercase.

```suggestion
    if (/input|textarea/i.test(event.target.tagName.toLowerCase())
```

simonihmig: Fixed.

PR #924, opened 2019-11-02
app.options['ember-bootstrap'] may be undefined

jelhan:
This will hopefully fix Ember Twiddle as well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ module.exports = {
     this.app = app;
 
     let options = Object.assign({}, defaultOptions, app.options['ember-bootstrap']);
-    if (options.bootstrapVersion === 4 && app.options['ember-bootstrap'] && app.options['ember-bootstrap'].importBootstrapFont) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     this.app = app;
 
     let options = Object.assign({}, defaultOptions, app.options['ember-bootstrap']);
-    if (options.bootstrapVersion === 4 && app.options['ember-bootstrap'].importBootstrapFont) {
+    if (options.bootstrapVersion === 4 && app.options['ember-bootstrap'] && app.options['ember-bootstrap'].importBootstrapFont) {
```
simonihmig: Good catch! I just wonder if there is any reason to use `app.options['ember-bootstrap']` here, and not just `options`, which should always be defined, right?

jelhan: Good catch! :+1: Seems like we missed that in #859. Refactored to `options.importBootstrapFont`, which is way more readable.

PR #927, opened 2019-11-02
Add ember-a11y-testing

### LINKED ISSUE #224: Review and update a11y
Bootstrap 4 has done a more thorough job of addressing aria markup and a11y concerns. We should review the current state and bring ember-bootstrap up to the latest.

simonihmig:
This does not replace specific tests for a11y features (e.g. testing for ARIA roles), but serves at least as a sanity check in case something has been forgotten. (in fact it uncovered #926)

Similar check should be added to any new component tests we introduce (hint: toasts). /cc @jelhan  

Closes #224

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "chai": "^4.0.0",
     "chai-things": "^0.2.0",
+    "ember-a11y-testing": "^1.1.1",
     "ember-bootstrap-power-select": "^1.1.0",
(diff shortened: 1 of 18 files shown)


simonihmig:
@jelhan Thanks, yes, was a productive afternoon! 😀I merged some of the smaller PRs right away to not block further work, and figured they were not very large and controversial. Though if you find something in those that you want to comment on, feel free to do so! I can still iterate on stuff if needed! :wink:

Issue #929, opened 2019-11-12
Use a standalone app for docs


simonihmig:
The current setup of using the addon's dummy app for the docs site has caused conflicts more than once, e.g. using latest features for the docs would break ember-try scenarios of older Ember versions for the addon. Also dependencies of the docs can cause other unwanted side-effects in the addon tests.

Having the docs as a separate app with its separate dependencies and build seems better overall!?

Related: https://github.com/ember-learn/ember-cli-addon-docs/pull/409

PR #930, opened 2019-11-12
Test addon build with embroider

### LINKED ISSUE #915: Add Embroider test scenario

simonihmig:
Closes #915

Currently not working, due to some ember-prism related build failure w/ embroider. #929 would fix that...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,7 @@ jobs:
             command: ember test --launch Firefox
           - scenario: fastboot-tests
           - scenario: node-tests
+          - scenario: embroider-tests
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/addon/templates/components/bs-form/element/control/radio.hbs
+++ b/addon/templates/components/bs-form/element/control/radio.hbs
@@ -1,3 +1,4 @@
+{{!-- template-lint-disable no-positive-tabindex --}}
 {{#each @options as |option index|}}
   {{#let (concat @id "-" index) as |id|}}
     {{#if (macroCondition (macroGetOwnConfig "isBS3"))}}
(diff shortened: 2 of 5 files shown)


jelhan:
Not sure if I got it correctly but seems like we can't remove the dependency on `ember-cli-fastboot-testing` but at the same time we can't include it cause it brakes Embroider. If that's the case, did you tried keeping the FastBoot dependency but setting `FASTBOOT_DISABLED=true` while running the embroider test? If that doesn't work removing the FastBoot related tests from the build might be another approach? There is an example how to do it in [this Discourse topic](https://discuss.emberjs.com/t/remove-a-particular-file-folder-from-the-app-tree-depends-upon-the-build-environment/14635/2).

simonihmig:
> Not sure if I got it correctly but seems like we can't remove the dependency on ember-cli-fastboot-testing but at the same time we can't include it cause it brakes Embroider. 

Correct.

> If that's the case, did you tried keeping the FastBoot dependency but setting FASTBOOT_DISABLED=true while running the embroider test? If that doesn't work removing the FastBoot related tests from the build might be another approach? There is an example how to do it in this Discourse topic.

Thanks for the suggestions. I just pushed a commit with a combination of those (and blacklisting ember-cli-fastboot-testing). Which solves the issue of a failing build, and most tests should pass now. With a couple of exceptions, which seem to be caused by an Embroider issue, specifically the `{{hasBlock}}` helper is somehow not working. Already investigated this, will file an issue tomorrow...

simonihmig:
Blocked on https://github.com/embroider-build/embroider/issues/373

simonihmig:
Rebased this, and with the latest dependencies this is passing now! :tada:
Seems the issues have been fixed upstream, through we still have to disable FastBoot due to https://github.com/embroider-build/embroider/issues/184.

PR #933, opened 2019-11-29
Deprecate arguments whose only purpose is supporting HTML attributes

jelhan:
This should deprecate all properties whose only purpose is supporting HTML attributes _and_ which are not forwarded internally.

Can not deprecate the arguments on `FormElementControl` yet cause they are set by Ember Boostrap itself (`FormElement` component) and `{{component}}` helper does not support HTML attributes yet. Seems like we need to delay that _after_ the next major.

Actually I had expected that more arguments are affected. Seems like the pattern was only used on `FormElement`,  `FormElementControl` and `BsButton`.

Part of #826

### DIFF
--- a/addon/components/base/bs-button.js
+++ b/addon/components/base/bs-button.js
@@ -10,6 +10,7 @@ import { observes } from '@ember-decorators/object';
 import { computed } from '@ember/object';
 import { equal, or } from '@ember/object/computed';
 import { scheduleOnce } from '@ember/runloop';
+import { deprecate } from '@ember/debug';
 import Component from '@ember/component';
 import layout from 'ember-bootstrap/templates/components/bs-button';
 import sizeClass from 'ember-bootstrap/utils/cp/size-class';
@@ -148,6 +149,7 @@ export default class Button extends Component {
    * @property disabled
    * @type ?boolean
    * @default null
+   * @deprecated
    * @public
    */
   @defaultValue
@@ -168,6 +170,7 @@ export default class Button extends Component {
    * @property buttonType
    * @type String
    * @default 'button'
+   * @deprecated
    * @public
    */
   @defaultValue
@@ -374,6 +377,7 @@ export default class Button extends Component {
    *
    * @property title
    * @type string
+   * @deprecated
    * @public
    */
   @defaultValue
@@ -448,5 +452,38 @@ export default class Button extends Component {
     }
 
     return this.get('bubble');
+  }
+
+  init() {
+    super.init(...arguments);
+
+    // deprecate arguments used for attribute bindings only
+    [
+      ['buttonType:type', 'submit'],
+      ['title', 'foo'],
+      ['disabled', true],
+    ].forEach(([mapping, value]) => {
+      let argument = mapping.split(':')[0];
+      let attribute = mapping.includes(':') ? mapping.split(':')[1] : argument;
+      let deprecationMessage =
+        `Argument ${argument} of <BsButton> component is deprecated. It's only purpose ` +
+        `was setting the HTML attribute ${attribute} of the control element. You should use ` +
+        `angle bracket  component invocation syntax instead:\n` +
+        `Before:\n` +
+        `  {{bs-button ${attribute}=${typeof value === 'string' ? `"${value}"` : value}}}\n` +
+        `  <BsButton @${attribute}=${typeof value === 'string' ? `"${value}"`: `{{${value}}}`} />\n` +
+        `After:\n` +
+        `  <BsButton ${typeof value === 'boolean' ? ( value ? attribute : '' ) : `${attribute}="${value}"`} />`;
+
+      deprecate(
+        deprecationMessage,
+        // eslint-disable-next-line ember/no-attrs-in-components
+        !Object.keys(this.attrs).includes(argument),
+        {
+          id: `ember-bootstrap.deprecated-argument.button#${argument}`,
+          until: '4.0.0',
+        }
+      );
+    });
   }
 }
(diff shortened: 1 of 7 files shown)


addon/components/base/bs-button.js
```diff
@@ -449,4 +453,37 @@ export default class Button extends Component {
 
     return this.get('bubble');
   }
+
+  init() {
```
jelhan: I tried to write a decorator instead of putting it here directly but wasn't successful. Haven't expected that decorators are that hard. I had some solutions but they were breaking quickly if put in different context...

addon/components/base/bs-button.js
```diff
@@ -449,4 +453,37 @@ export default class Button extends Component {
...
+        `Before:\n` +
+        `  {{bs-button ${attribute}=${typeof value === 'string' ? `"${value}"` : value}}}\n` +
+        `  <BsButton @${attribute}=${typeof value === 'string' ? `"${value}"`: `{{${value}}}`} />\n` +
+        `After:\n` +
+        `  <BsButton ${typeof value === 'boolean' ? ( value ? attribute : '' ) : `${attribute}="${value}"`} />`;
+
+      deprecate(
+        deprecationMessage,
+        // eslint-disable-next-line ember/no-attrs-in-components
+        !Object.keys(this.attrs).includes(argument),
```
jelhan: `this.attrs` is internal API and there is even an ESLint rule to not use it. But there isn't a better API available. Relying on it's keys only is actually a recommended interim solution until `@glimmer/components` introduces `this.args` public API. Pzuraq [calls it an _intimate API_](https://discuss.emberjs.com/t/best-way-to-get-attrs-only/17021). I think it's quite unlikely that it will be removed before `@ember/component` is removed entirely.

tests/helpers/setup-no-deprecations.js
```diff
@@ -36,14 +36,16 @@ export default function setupNoDeprecations({ beforeEach, afterEach }) {
   };
 
   QUnit.assert.deprecationsInclude = function(expected) {
-    let foundIndex = deprecations.indexOf(expected);
+    let found = deprecations.find((deprecation) => deprecation.includes(expected));
```
jelhan: Matching by a substring wasn't possible before but needed due to the long deprecation messages introduced.

jelhan:
Not sure about the labels. It's not a new feature but also not a bug fix and not only internal. It's also not a breaking change. _Enhancement_ seems to fit best but only cause of the list available. :smile:

addon/components/base/bs-button.js
```diff
@@ -449,4 +452,36 @@ export default class Button extends Component {
...
+        `Argument ${argument} of <BsButton> component is deprecated. It's only purpose ` +
+        `was setting the HTML attribute ${attribute} of the control element. You should use ` +
+        `angle bracket  component invocation syntax instead:\n` +
+        `Before:\n` +
+        `  {{bs-button ${attribute}="${value}"}}\n` +
+        `  <BsButton @${attribute}="${value}" />\n` +
+        `After:\n` +
+        `  <BsButton ${attribute}="${value}" />`;
+
+      deprecate(
```
simonihmig: One "problem" here and at the other places where `deprecate()` is called, is that the preceding code add's a lot of unnecessary code to production builds. Ember (I guess ember-cli-babel) will remove any occurrence of `deprecate()` from production, but only that. So either
* put all code into a number of explicit `deprecate()` calls, i.e. duplication the message, the assertion etc., nothing else
* wrap everything related to deprecations in a `runInDebug` call, see https://api.emberjs.com/ember/3.14/functions/@ember%2Fdebug/runInDebug

simonihmig:
> Can not deprecate the arguments on FormElementControl yet cause they are set by Ember Boostrap itself (FormElement component) and {{component}} helper does not support HTML attributes yet.

We could maybe pass them with some new internal-only properties (e.g. `_placeholder` instead of `placeholder`), which don't raise deprecations, and show the deprecations for the official props? This way we can treat the internal props as strictly private API, that we can drop anytime there is a better way to pass HTML attributes using `{{component}}`, but the public ones are immediately deprecated, and as such not requiring any new major version.

addon/components/base/bs-button.js
jelhan: Good catch. Have wrapped the deprecation related code in `runInDebug` in f9df0e7d499710b54ee450fbd4c5a0d1c156d64e.

jelhan:
> We could maybe pass them with some new internal-only properties (e.g. _placeholder instead of placeholder), which don't raise deprecations, and show the deprecations for the official props? This way we can treat the internal props as strictly private API, that we can drop anytime there is a better way to pass HTML attributes using {{component}}, but the public ones are immediately deprecated, and as such not requiring any new major version.

Done so for `disabled` argument of `<BsButton>` in 37d63154d588bc9e1c5618fa1b0c96646c09a51f. But I'm not sure about the arguments passed to `FormElementControl` to be honest. I fear the boilerplate code this would require. We are talking about a lot arguments used in four components. Also not sure about the performance impact. But I get the argument that deprecating everything in the same major avoids confusion and churn.

simonihmig:
> Done so for disabled argument of <BsButton> in 37d6315

That broke some tests in seems!

jelhan:
> > Done so for disabled argument of  in [37d6315](https://github.com/kaliber5/ember-bootstrap/commit/37d63154d588bc9e1c5618fa1b0c96646c09a51f)
> 
> That broke some tests in seems!

Sorry, didn't noticed that CI was failing for Ember 3.4 and 2.18. Should be fixed by 5cec6401c4f6df526ec282085d365b11f226e573. Wasn't aware that `@defaultValue` decorator also provides a work-a-round for missing [RFC 337 Native Class Constructor Update](https://github.com/emberjs/rfcs/blob/master/text/0337-native-class-constructor-update.md) in older Ember versions. Haven't used it cause we don't expect any consumer to override that internal variable.

The [ember-native-class-polyfill](https://github.com/pzuraq/ember-native-class-polyfill) does not support Ember < 3.4. So I think there isn't any better solution than just using `defaultValue` decorator for now.

Issue #939, opened 2019-12-12
Assertion Failed: Backdrop element should be in DOM if optional feature async-observers is enabled


Subaku:
Getting this error in the console with no other information when attempting to open a Modal programmatically. I.e 

`<BsModal @open={{showModal}} as |Modal|>...</BsModal>`

Note that `showModal` is a computed property that starts out false but once the condition for it to turn true happens (which is does) the model fails to open and I get the mentioned error.

Also note that Modals will open fine if I don't specify anything for the `@open` attribute or I'm referencing a simple boolean attribute off my component which I can toggle from the inspector. Only when it's from a computed property do I have any issues.

Ember: 3.14.1
Ember-Bootstrap: 3.1.0

jelhan:
Are you able to provide a reproduction? We don't have a test for programmatically showing a `<BsModal>` but it should be covered by this test for `<BsModalSimple>` implicitly: https://github.com/kaliber5/ember-bootstrap/blob/58887daf2e15e7da07423772ec55510738979a24/tests/integration/components/bs-modal-simple-test.js#L89-L101

Subaku:
Hi @jelhan 

Here's a repo I threw up with my current project I'm working on that exhibits the problem.
https://github.com/Subaku/ember-modal-issue.git

It's bare bones atm. If you click on any of the rendered View buttons in the table you should get some "asset-locations" loaded from mirage. This in turn makes the computed `showModal` inside the `asset-location-summaries/locations` component turn true and then you should get the error in the console.

Let me know if you need anything and thanks!

jelhan:
Seems like some of your code is manipulating the DOM. I started debuging at this warning:

> WARNING: No wormhole destination element found for component <inventory@component:bs-modal-simple::ember196>. If you have set `insertEmberWormholeElementToDom` to false, you should insert a `div#ember-bootstrap-wormhole` manually!

`<div id="ember-bootstrap-wormhole">` is in the `dist/index.html`. But if I put a debugger here it's not in the DOM: https://github.com/kaliber5/ember-bootstrap/blob/58887daf2e15e7da07423772ec55510738979a24/addon/utils/dom.js#L65

It's present in the DOM again when the page is loaded. So it might be related to your loading animation.

[...]

PR #945, opened 2019-12-15
add test coverages for optional feature: async-observers

jelhan:
As reported in #939 Ember Bootstrap does not work with optional feature *async-observers*. This adds another ember try scenario running our test suite with that optional feature enabled to see how worse it is (and provide a starting point to fix it).

We might want to remove scenario as soon as we remove the last observer from our code base. We might want to ensure before removing that all of our dependencies don't use observers anymore or have a similar scenario in their CI pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ jobs:
   - env: EMBER_TRY_SCENARIO=ember-beta
   - env: EMBER_TRY_SCENARIO=ember-canary
   - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
+  - env: EMBER_TRY_SCENARIO=ember-default-with-async-observers
   - env: EMBER_TRY_SCENARIO=ember-lts-2.18 BOOTSTRAPVERSION=3
   - env: EMBER_TRY_SCENARIO=ember-lts-3.4 BOOTSTRAPVERSION=3
   - env: EMBER_TRY_SCENARIO=ember-release BOOTSTRAPVERSION=3
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -86,6 +86,14 @@ module.exports = function() {
           }
         },
         {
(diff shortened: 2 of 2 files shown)


jelhan:
Following components does not seem to work with `async-observers: true`:

- `<BsAlert>`
- `<BsCollapse>`
- `<BsModal>` and `<BsModalSimple>`
- `<BsNavbar>`
- `<BsTab>`

simonihmig:
After the recent Ember update, I did turn on the flag locally to see what happens, and saw that a lot of tests were breaking. But it *looked* as if a lot of that was related to the new asynchrony not being correctly accounted for in async test helpers, e.g. `await click()`. Which kinda makes sense, when you assert that something happened after the click, which was happening before as it was running *synchronously*, but that happens now asynchronously but `await` is not waiting for it.

But I would need to check again more thoroughly to make sure this is indeed what is happening.

simonihmig:
Wondering how to proceed with this PR... maybe configure this scenario as allowed to fail, merge it, tackle the fixes in subsequent PRs?

jelhan:
> But it _looked_ as if a lot of that was related to the new asynchrony not being correctly accounted for in async test helpers, e.g. `await click()`.

Is that behavior mentioned in docs? I would have expected that an application is not considered to be settled (by test helpers) before async observers are processed.

> Wondering how to proceed with this PR... maybe configure this scenario as allowed to fail, merge it, tackle the fixes in subsequent PRs?

Maybe just keep it as a WIP and rebase when ever something changed related to observers? If we merge it, I would tend to explicitly skip tests for components that aren't working with async observers yet. This would allow us to enable them one by one but also ensure that we don't introduce regression. But as I tend to fix this issue by removing observers we might don't need this scenario at all?

jelhan:
Superseded by #951

PR #947, opened 2019-12-18
Fix typo (Pikaday instead of PiCkaday)

gianlucatomasino:


gianlucatomasino:
> Good catch. Thanks a lot.

I lost some hours before understanding that there was a typo 🤣.

PR #948, opened 2019-12-20
Modals ignore clicking backdrop to close, if mousedown happened on dialog element, reflecting original bootstrap.js behaviour

### LINKED ISSUE #934: Using mouse-down and mouse-up for closing the modal dialog
I'm not sure if it's a bug but it's definitely a difference from how Bootstrap's modals are handled. Bootstrap's implementation checks whether mouse button was down and up on the same element (https://github.com/twbs/bootstrap/blob/master/js/src/modal.js#L154). This prevents closing the modal when user holds down the button on a modal, drags the cursor to the backdrop and releases the button.

It's especially annoying if you have a text area inside the modal and want to select / highlight the text quickly. If you slightly move the cursor out of the modal you will close the dialog by mistake.

Is there any specific reason for using `click`?

Output from `ember bootstrap:info`:

```
Npm packages:
ember-bootstrap: ^2.6.1 -> 2.6.1
ember-cli: ~3.8.3 -> 3.8.3
bootstrap: n/a
bootstrap-sass: ^3.4.0 -> 3.4.1
ember-cli-sass: ^5.3.0 -> 5.6.0
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
importBootstrapCSS: false
importBootstrapFont: true
bootstrapVersion: 3
```

eyupatis:
In bootstrap's source code, it tries to not close the modals if the
mouse down event was inside of the modal and mouse up event was outside
of the modal. With these changes we are copying the same behaviour from
it. We are also using similar variables to reflect it's source code.

Here is the click event of bootstrap: https://github.com/twbs/bootstrap/blob/master/js/src/modal.js#L361
Here is the handlers for mouse down and up events: https://github.com/twbs/bootstrap/blob/master/js/src/modal.js#L154

### DIFF
--- a/addon/components/base/bs-modal/dialog.js
+++ b/addon/components/base/bs-modal/dialog.js
@@ -163,8 +163,28 @@ export default class ModalDialog extends Component {
         }
       }
     }
-    this.set('titleId', nodeId)
+    this.set('titleId', nodeId);
   }
+
+  /**
+   * If true clicking on the backdrop will be ignored and will not close modal.
+   *
+   * @property ignoreBackdropClick
+   * @type boolean
+   * @default false
+   * @private
+   */
+  ignoreBackdropClick = false;
+
+  /**
+   * The target DOM element of mouse down event.
+   *
+   * @property mouseDownElement
+   * @type object
+   * @default null
+   * @private
+   */
+  mouseDownElement = null;
 
   /**
    * Update the elements styles using CSSOM.
@@ -204,10 +224,24 @@ export default class ModalDialog extends Component {
   }
 
   _click(e) {
+    if (this.ignoreBackdropClick) {
+      this.set('ignoreBackdropClick', false);
+      return;
+    }
     if (e.target !== this.element || !this.get('backdropClose')) {
       return;
     }
     this.get('onClose')();
+  }
+
+  _mouseDown(e) {
+    this.set('mouseDownElement', e.target);
+  }
+
+  _mouseUp(e) {
+    if (this.mouseDownElement !== this.element && e.target === this.element) {
+      this.set('ignoreBackdropClick', true);
+    }
   }
 
   didInsertElement() {
@@ -216,6 +250,8 @@ export default class ModalDialog extends Component {
     // iOS to allow clicking the div. So a `click(){}` method here won't work, we need to attach an event listener
     // directly to the element
     this.element.onclick = bind(this, this._click);
+    this.element.onmousedown = bind(this, this._mouseDown);
+    this.element.onmouseup = bind(this, this._mouseUp);
     this.getOrSetTitleId();
     this.updateStyles();
   }
--- a/addon/components/base/bs-modal/dialog.js
+++ b/addon/components/base/bs-modal/dialog.js
@@ -234,11 +234,11 @@ export default class ModalDialog extends Component {
     this.get('onClose')();
   }
 
-  _mouseDown(e) {
+  mouseDown(e) {
     this.set('mouseDownElement', e.target);
   }
 
-  _mouseUp(e) {
+  mouseUp(e) {
     if (this.mouseDownElement !== this.element && e.target === this.element) {
       this.set('ignoreBackdropClick', true);
     }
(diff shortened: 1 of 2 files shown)


simonihmig:
@eyupatis thanks for working on this! 

So far this looks good! But I would love to have a test covering that behaviour. Would you be able to add one? Something that basically simulates what is described in #934:

> This prevents closing the modal when user holds down the button on a modal, drags the cursor to the backdrop and releases the button.

eyupatis:
@simonihmig I added two test cases that covers existing behavior and newly added behavior.

tests/integration/components/bs-modal-test.js
```diff
@@ -151,4 +151,28 @@ module('Integration | Component | bs-modal', function(hooks) {
...
+
+  test('it keeps itself visible when mouse click is started from inside of the modal \
+    and dragged outside of the modal', async function(assert) {
+    await render(hbs`
+      <BsModal>
+        template block text
+      </BsModal>
+    `);
+    await triggerEvent('.modal-content', 'mousedown');
+    await triggerEvent('.modal', 'mouseup');
```
simonihmig: I think we should add `await click('.modal');` here, as that is the sequence of events, see https://javascript.info/mouse-events-basics#complex-events. Also only the click event would have triggered the wrong behaviour previously. Could you please try to confirm that the test fails without the changes in the component?

addon/components/base/bs-modal/dialog.js
```diff
@@ -204,18 +224,34 @@ export default class ModalDialog extends Component {
...
+  }
+
   didInsertElement() {
     super.didInsertElement(...arguments);
     // Ember events use event delegation, but we need to add an `onclick` handler directly on the modal element for
     // iOS to allow clicking the div. So a `click(){}` method here won't work, we need to attach an event listener
     // directly to the element
     this.element.onclick = bind(this, this._click);
+    this.element.onmousedown = bind(this, this._mouseDown);
+    this.element.onmouseup = bind(this, this._mouseUp);
```
simonihmig: The explicit `onclick` handler was required for iOS issue described in the comment. But the other ones could use just the built-in event handler methods (https://guides.emberjs.com/v3.14.0/components/handling-events/), i.e. rename `_mouseDown` to `mouseDown` and not have this code in `didInsertElement`. Or do I miss something?

tests/integration/components/bs-modal-test.js
jelhan: `click` test helper provided by `@ember/test-helpers` would trigger another `mousedown` event. So I guess we would need to do:

```js
    await triggerEvent('.modal-content', 'mousedown');
    await triggerEvent('.modal', 'mouseup');
    await triggerEvent('.modal', 'click');
```

eyupatis: @simonihmig Yeah, when I revert my changes locally on the component, the test case with name `it keeps itself visible when mouse click is started from inside of the modal and dragged outside of the modal` is failing locally.

@jelhan Thank you for your help, I updated the test case accordingly.

addon/components/base/bs-modal/dialog.js
eyupatis: Yeah, you are right about the event handlers, I updated pr with ember's default event handlers, thank you for warning me about this.

addon/components/base/bs-modal/dialog.js
```diff
@@ -204,12 +224,26 @@ export default class ModalDialog extends Component {
   }
 
   _click(e) {
+    if (this.ignoreBackdropClick) {
```
jelhan:
```suggestion
    if (this.get('ignoreBackdropClick')) {
```

I think we should use `this.get()` here. It shouldn't make a difference for this specific case but it's easier to reason about if we always use `this.get()` unless support for Ember < 3.1 has been dropped.

simonihmig: Yes, in theory that's right, but as it's a private property and not expected to be changed to a CP (where `.get()` would be needed for old Ember versions), I think we can accept this and not cause any further work! :wink:

simonihmig:
Thank you for your contribution, @eyupatis!

eyupatis:
It was a pleasure to contribute and work with you, thank you for accepting and merging pr :)

PR #951, opened 2020-01-04
support optional feature default-async-observers

### LINKED ISSUE #939: Assertion Failed: Backdrop element should be in DOM if optional feature async-observers is enabled

jelhan:
Adds support for optional feature `default-async-observers: true` by explicitly setting all observers to sync, which must not be run async.

It also enables optional feature `template-only-glimmer-components` for testing. It was already supported. Only changed it to prevent a regression.

Adds an additional ember try scenario that ensures it's working for the opposite optional feature configuration. This is similar to the ember-classic scenario, which is part of Ember CLI's blueprints since 3.15.

Closes #939
Superseds #945

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -95,6 +95,16 @@ module.exports = function() {
           }
         },
         {
+          name: 'ember-classic',
+          env: {
+            EMBER_OPTIONAL_FEATURES: JSON.stringify({
+              'application-template-wrapper': true,
+              'default-async-observers': false,
+              'template-only-glimmer-components': false
+            })
+          }
+        },
+        {
           name: 'fastboot-tests',
           command: 'ember test --filter FastBoot',
           npm: {
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -76,7 +76,7 @@ module.exports = function() {
           name: 'ember-default',
           npm: {
             devDependencies: {
-            'bootstrap': bootstrapVersion
+              'bootstrap': bootstrapVersion
             }
           }
         },
@@ -98,10 +98,16 @@ module.exports = function() {
           name: 'ember-classic',
           env: {
             EMBER_OPTIONAL_FEATURES: JSON.stringify({
-              'application-template-wrapper': true,
+              // TODO: tests fail if optional feature application-template-wrapper is enabled
+              // 'application-template-wrapper': true,
               'default-async-observers': false,
               'template-only-glimmer-components': false
             })
+          },
+          npm: {
+            devDependencies: {
+              'bootstrap': bootstrapVersion
+            }
           }
         },
         {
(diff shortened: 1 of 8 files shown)


config/ember-try.js
```diff
@@ -94,6 +94,22 @@ module.exports = function() {
             }
           }
         },
+        {
+          name: 'ember-classic',
+          env: {
+            EMBER_OPTIONAL_FEATURES: JSON.stringify({
+              // TODO: tests fail if optional feature application-template-wrapper is enabled
+              // 'application-template-wrapper': true,
```
jelhan: I've no idea why this is happening. But I don't think it's related to this PR. It might also not be a real-world problem but only be triggered by our tests. At least I'm not aware of any bug reports that may be related to this.

jelhan:
Seems like I need to take care of some additional observers, which are used for BS3 only.

jelhan:
Changed some additional observer that is required for BS3 to be sync. Ready to review / merge from my side.

config/ember-try.js
simonihmig: Not sure which test failed because of this, but I would expect this to potentially cause failures in tests. Especially for modals or other wormholed stuff. 

The flag also affects tests, so with it enabled `await render(hbs\`<BsButton .../>\`);` would render the button wrapped in a `<div>` which is then the application root element. Something like `await click('*')` (we don't have this in the code base much anymore, but I previously used this to pick the first rendered element, whatever the tag name it has or class...) would not trigger the click on the button, but the wrapper.

For modals [`getDestinationElement()`])(https://github.com/kaliber5/ember-bootstrap/blob/master/addon/utils/dom.js#L63) would pick the test container as the wormhole destination element. But with the flag enabled, the wrapper div would become the application root element, thus the modal would be rendered *outside* of it, causing at least problems with events (see the other issue comment of today)

This is only an issue for tests though!

simonihmig:
@jelhan Oh, just noticed that the new `ember-classic` scenario is not running in CI, as it is missing in `.travis.yml`! Please add that, and then feel free to merge!

jelhan:
> @jelhan Oh, just noticed that the new `ember-classic` scenario is not running in CI, as it is missing in `.travis.yml`! Please add that, and then feel free to merge!

Good catch! Added that scenario to TravisCI in fa5ac69afbaf289028546fc42d925c36cccc85e6.

PR #964, opened 2020-01-10
upgrade ember to 3.15

jelhan:


package.json
```diff
@@ -131,6 +135,7 @@
   },
   "resolutions": {
     "ip-regex": "^2.1.0",
-    "ember-concurrency": "^1.0.0"
+    "ember-concurrency": "^1.0.0",
+    "ember-cli-htmlbars/semver": "~7.0.0"
```
jelhan: Can be removed as soon as node 8 support is dropped. `semver@^7.1.0` does not support Node 8.

PR #965, opened 2020-01-10
drop support for ember < 3.15 and node 8

jelhan:

PR #966, opened 2020-01-10
refactor component invocation in tests to angle brackets syntax

jelhan:
This is based on #965, which should be merged before.


simonihmig:
Seems like a lot of work! Maybe it makes sense to invest some time in https://github.com/ember-codemods/ember-angle-brackets-codemod/issues/167, to let the codemod do most of the work?

simonihmig:
Oh wait, the Github diff was not loading correctly, or took longer... it seams you already migrated most if not all of the code!? 

@jelhan How did you do it, manually, find/replace, or does some codemod exist to do this?

jelhan:
`ember-angle-bracket-codemod` can migrate inline templates as well. I think it just skips all non-handlebars files by default. So `npx ember-angle-brackets-codemod http://localhost:4200 tests/` does not work. But providing a RegExp that matches the JavaScript files works well: `npx ember-angle-brackets-codemod http://localhost:4200 tests/**/*.js`

Most of the work committed so far was done by the codemod. Just needed to fix some template linting stuff. At least my editor (VSCode) was complaining about it. Not sure if `yarn lint:hbs` actually lints the inline templates as well.

But the codemod skipped some files. Will investigate why. Don't want to convert them by hand cause as you said: That would be a lot of work.

jelhan:
Could run the codemod on nearly all files after removing HTML tag like elements from the files. See https://github.com/ember-codemods/ember-angle-brackets-codemod/issues/167#issuecomment-573178357 for details. These four files are failing with hard to debug errors:

- [ ] `bs-form-test.js`: `A block may only be used inside an HTML element or another block.`
- [ ] `bs-form/element-test.js`: `Syntax error at line 394 col 17: ' is not a valid character within attribute names`
- [ ] `bs-popover-test.js`: `Syntax error at line 101 col 16: ' is not a valid character within attribute names`
- [ ] `bs-tooltip-test.js`: `Syntax error at line 390 col 7: ' is not a valid character within attribute names`

Will try to fix them later. Hope I didn't missed one.

tests/integration/components/bs-modal/header-test.js
```diff
@@ -18,37 +18,37 @@ module('Integration | Component | bs-modal/header', function(hooks) {
...
   test('close button can be removed', async function(assert) {
-    await render(hbs`{{bs-modal/header title="Header" closeButton=false}}`);
+    await render(hbs`<BsModal::Header @title="Header" @closeButton={{false}} />`);
 
     assert.dom('.modal-header button.close').doesNotExist('Modal header has no close button.');
   });
 
   test('close button can be removed in yield block form', async function(assert) {
-    await render(hbs`{{#bs-modal/header as |header|}}<div id="custom">Test</div>{{/bs-modal/header}}`);
+    await render(hbs`<BsModal::Header><div id="custom">Test</div></BsModal::Header>`);
```
jelhan: This test is failing now:

```
not ok 653 Chrome 79.0 - [54 ms] - Integration | Component | bs-modal/header: close button can be removed in yield block form

    ---

        actual: >

            Element .modal-header button.close exists once

        expected: >

            Element .modal-header button.close does not exist

        stack: >

                at DOMAssertions.exists (http://localhost:7357/assets/test-support.js:40540:16)

                at DOMAssertions.doesNotExist (http://localhost:7357/assets/test-support.js:40848:18)

                at Object._callee6$ (http://localhost:7357/assets/tests.js:18278:58)

                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)

                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)

                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)

        message: >

            Modal header has no close button.

        negative: >

            false

        browser log: |

    ...
```

I don't see how that should be related to this change. :confused:

jelhan: `Integration | Component | bs-form/element` is also failing on a strange way. Both `supported input attributes propagate` and `supported textarea attributes propagate` are failing for `readonly` attribute only. :dizzy_face:

simonihmig: Not sure what's happening for `readonly`. But as this is going to be removed anyway, as part of the deprecated attributes, maybe we should just delete these tests here, and stop caring!? :)

jelhan: Yeah, lets be pragmatic. Enough work still left to do.

tests/integration/components/bs-form/element-test.js
```diff
@@ -493,34 +485,16 @@ module('Integration | Component | bs-form/element', function(hooks) {
...
-        title=title
-      }}`);
+      await render(hbs`<BsForm::Element @formLayout={{formLayout}} @name={{name}} @required={{required}} @readonly={{readonly}} @placeholder={{placeholder}} @disabled={{disabled}} @autofocus={{autofocus}} @controlSize={{size}} @tabindex={{tabindex}} @minlength={{minlength}} @maxlength={{maxlength}} @min={{min}} @max={{max}} @pattern={{pattern}} @accept={{accept}} @autocomplete={{autocomplete}} @autocapitalize={{autocapitalize}} @autocorrect={{autocorrect}} @autosave={{autosave}} @inputmode={{inputmode}} @multiple={{multiple}} @step={{step}} @form={{form}} @spellcheck={{spellcheck}} @title={{title}} />`);
 
       for (let attribute in supportedInputAttributes) {
+        if (attribute === 'readonly') {
+          // tests are failing for readonly attribute for unknown reason
+          // but as it's deprecated and will be removed in next release
+          // we don't care
+          break;
```
simonihmig: Tests fail for ember-release and canary due to an unexpected deprecation. I guess this is due to the following `assert.deprecationsInclude()` not being called here because of the early return? Btw, why `break` and not `continue`?

What about completely deleting all these `supported ... attributes propagate...` tests? This would have to be done anyway when dropping support for these attributes, right?

jelhan: > Btw, why `break` and not `continue`?

I was confused. Should have put more focus into it. It should have been `continue` not break.

> What about completely deleting all these `supported ... attributes propagate...` tests? This would have to be done anyway when dropping support for these attributes, right?

I like to do it in separate merge requests to have better track of the changes. At least if it's not too much extra work.

jelhan:
Should be ready for _squash and merge_.

PR #968, opened 2020-01-13
test scenario expects closeButton to be false

jelhan:
`closeButton` defaults to `true` and isn't different for block mode or not. Test scenario for inline mode, which is just one test above, passes `closebutton=false` as it should be. I don't have any idea why the test was passing without `closeButton=false` but it was failing after refactoring to angle bracket invocation: https://github.com/kaliber5/ember-bootstrap/pull/966#discussion_r365409274

### DIFF
--- a/tests/integration/components/bs-modal/header-test.js
+++ b/tests/integration/components/bs-modal/header-test.js
@@ -48,7 +48,7 @@ module('Integration | Component | bs-modal/header', function(hooks) {
   });
 
   test('close button can be removed in yield block form', async function(assert) {
-    await render(hbs`{{#bs-modal/header as |header|}}<div id="custom">Test</div>{{/bs-modal/header}}`);
+    await render(hbs`{{#bs-modal/header closeButton=false as |header|}}<div id="custom">Test</div>{{/bs-modal/header}}`);
 
     assert.dom('.modal-header div#custom').exists({ count: 1 }, 'Modal header custom block.');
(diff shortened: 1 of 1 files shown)


tests/integration/components/bs-modal/header-test.js
```diff
@@ -48,7 +48,7 @@ module('Integration | Component | bs-modal/header', function(hooks) {
   });
 
   test('close button can be removed in yield block form', async function(assert) {
-    await render(hbs`{{#bs-modal/header as |header|}}<div id="custom">Test</div>{{/bs-modal/header}}`);
+    await render(hbs`{{#bs-modal/header closeButton=false as |header|}}<div id="custom">Test</div>{{/bs-modal/header}}`);
```
simonihmig: The API for the header is basically that you opt into custom markup if you use the component in block mode and with yielded components. It doesn't make sense to render the close button, if you are yielding the close button component to the user for him to specify how to render. See https://github.com/kaliber5/ember-bootstrap/blob/master/addon/templates/components/bs4/bs-modal/header.hbs#L1-L7

So I think the test has to remain as it is: even without `closeButton=false` the header component should not render anything.

jelhan: Oh got it. Sorry my fault. Now it makes sense why the test was failing after the refactoring. Has removed the block params cause template linter was complaining about the unused variable... Thanks for the hint.

jelhan: I think we need to document that behavior.

PR #980, opened 2020-02-01
Convert Button and Dropdown to tag-less / outerHTML component

simonihmig:
This required a bit more of refactoring, as Dropdowns kinda relied on internals of Buttons (`.element`).  This now uses some more modern approaches:
* `{{ref}}` modifier to pass DOM node reference
* `element` helper for dynamic tag names (based on [RFC389](https://emberjs.github.io/rfcs/0389-dynamic-tag-names.html))

The latter is another example for subtle breaking changes due to tag-less components, as you cannot modify a component's `tagName` anymore. In this case this was required for dropdown in a `.nav` (making them `<li>`s).


addon/components/base/bs-button.js
```diff
@@ -442,16 +432,17 @@ export default class Button extends Component {
    * @method click
    * @private
    */
-  click() {
-    let action = this.get('onClick');
+  @action
+  handleClick(e) {
+    let onClick = this.get('onClick');
```
simonihmig: This was a weird one! This failed with some undefined error when the local variable had the same name ("action") as the (newly added) decorator. Probably the decorator babel transform produced some wrong code!? 🤔

addon/templates/components/common/bs-dropdown/button.hbs
```diff
@@ -0,0 +1,21 @@
+<button
+  disabled={{this.__disabled}}
+  type={{this.buttonType}}
+  title={{this.title}}
+  aria-expanded={{this.aria-expanded}}
+  class="btn dropdown-toggle {{if this.active "active"}} {{if this.block "btn-block"}} {{this.sizeClass}} {{this.typeClass}}"
+  ...attributes
+  {{on "click" this.handleClick}}
+  {{on "keydown" this.handleKeyDown}}
+  {{ref @dropdown "toggleElement"}}
```
jelhan: Yielding the full component and changing it's content looks scary to me. Did you consider yielding a `registerToggleElement` method that is used here with `{{did-insert}}` modifier?

```js
  @action
  registerToggleElement(element) {
    assert(element instanceof HTMLElement);

    this.set('toggleElement', element);
  }
```

```hbs
  {{yield
    registerToggleElement=this.registerToggleElement
  }}
```

```hbs
  {{did-insert @registerToggleElement}}
```

addon/components/base/bs-dropdown/button.js
```diff
@@ -13,27 +13,15 @@ import { computed } from '@ember/object';
...
+@templateLayout(layout)
 export default class DropdownButton extends Button {
-  keyDown(e) {
+  @action
+  handleKeyDown(e) {
     this.get('onKeyDown')(e);
   }
 
   @computed('isOpen')
   get 'aria-expanded'() {
```
jelhan:
```suggestion
   get ariaExpanded() {
```

Not related to this change but noticed when reviewing the template.

addon/templates/components/common/bs-dropdown/button.hbs
```diff
@@ -0,0 +1,21 @@
+<button
+  disabled={{this.__disabled}}
+  type={{this.buttonType}}
+  title={{this.title}}
+  aria-expanded={{this.aria-expanded}}
```
jelhan:
```suggestion
  aria-expanded={{this.ariaExpanded}}
```

Per the suggested change in JavaScript file.

addon/components/base/bs-button.js
```diff
@@ -467,7 +458,9 @@ export default class Button extends Component {
       }
     }
 
-    return this.get('bubble');
+    if (!this.get('bubble')) {
+      e.stopPropagation();
+    }
```
jelhan: Oh, I wasn't aware that return value is not controlling event propagation anymore if used with `{{on}}` listener. Good catch! :clap:

addon/components/base/bs-button.js
jelhan: Wondering if we should report it upstream?

addon/templates/components/common/bs-dropdown/button.hbs
simonihmig: :+1:

We basically did set it directly also before that change, so the new code did not change that. But this certainly looks better! We do use this pattern already in the parent/child mixins, but should probably stop using these mixins for new code...

addon/components/base/bs-button.js
simonihmig: Yeah, that's a feature of the older Ember events / EventDispatcher, not a thing when using native `addEventListener()` that `{{on}}` is using.

Also a thing that the codemod is not doing correctly in this case. Wonder if we can/should add that code automatically!? 🤔

jelhan: > Also a thing that the codemod is not doing correctly in this case. Wonder if we can/should add that code automatically!? thinking

Not sure if automation is worse the time. But it should be at least mentioned in the docs. I think the codemod needs a section in the docs which lists all the possible edge cases that may be broken after conversion. The event bubbling is only of them, isn't it?

addon/templates/components/common/bs-dropdown/button.hbs
simonihmig: Done, including unregistering.

simonihmig:
> <BsDropdown::Menu::Divider> and <BsDropdown::Menu::Item> don't require version specific templates.

Actually, it was required to have separate templates, I just forgot to migrate those menu/* components. So they still had some class decorators. Fixed now!

addon/components/base/bs-button.js
simonihmig: Yeah, probably. But haven't figured out where yet. It's probably an issue upstream for the babel decorator plugin, or babel itself...

simonihmig:
Merging this, as I believe all concerns have been addressed.

Thanks @jelhan for the review, it really helped to improve the code!

jelhan:
> Thanks @jelhan for the review, it really helped to improve the code!

Thanks a lot for doing the conversion!

PR #981, opened 2020-02-01
Convert Modals to tag-less / outerHTML component

simonihmig:


package.json
```diff
@@ -101,6 +102,7 @@
     "ember-maybe-in-element": "^0.4.0",
     "ember-on-helper": "^0.1.0",
     "ember-popper": "^0.10.3",
+    "ember-style-modifier": "^0.5.0",
```
simonihmig: 👍 :wink:

jelhan: There is one downside of it: The style is not applied on server-side rendering / FastBoot. I'm planing to support it but it required more work as expected upstream at ember-cli-content-security-policy. I don't think this is an issue for this specific use case so. But wanted to make sure that we are aware.

simonihmig: Yeah, understood. But as we (you!) were already using CSSOM API, triggered mostly in `didInsertElement`, and not in-template style assignments, we were not applying these things in FastBoot anyway.

addon/templates/components/bs4/bs-modal/dialog.hbs
```diff
@@ -0,0 +1,27 @@
...
+  tabindex="-1"
+  aria-labelledby={{this.ariaLabelledby}}
+  class="modal {{if @fade "fade"}} {{if @showModal "show"}} {{if @inDom "d-block"}}"
+  ...attributes
+  {{on "keydown" this.handleKeyDown}}
+  {{on "mousedown" this.handleMouseDown}}
+  {{on "mouseup" this.handleMouseUp}}
+  {{on "click" this.handleClick}}
+  {{style paddingLeft=@paddingLeft paddingRight=@paddingRight display=(if @inDom "block" "")}}
+  {{ref this "_element"}}
```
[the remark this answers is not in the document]
simonihmig: That's a great point! Will try this out.

package.json
jelhan: Yeah. Server-side rendering, dynamic styles and Content Security Policy is a tricky combination. When doing the changes I fear I missed that this might break styles in FastBoot. Glad that no one reported bugs so far. Cause I fear that there is currently no solution which does not violate a strict CSP but supports server-side rendering.

addon/templates/components/bs4/bs-modal/dialog.hbs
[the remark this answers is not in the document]
jelhan: Similar here: https://github.com/kaliber5/ember-bootstrap/pull/980#pullrequestreview-351936207

simonihmig:
> If I got it right, it's safe cause a) it's an internal component and not part of our public API and b) the default values were always overruled by arguments passed in.

Exactly

> The documentation seems to be out of sync with that change. Especially the documented default values. In some cases the used types have also changed, haven't they?

Which documentation exactly are you referring to? We are talking only about private props, aren't we?

> I haven't expected that the migration to tagless components will cause that much code duplication for us.

I agree. I was kind of in a "tunnel mode" when doing the migration, and happy I was able to finish und make all tests pass again. 😅 But I agree, in hindsight this causes a lot of duplication. I love the approach you suggested. Also would probably allow us to have an easier migration to a native embroider addon some time in the (distant) future, instead of having our broccoli tricks. I'll open up another issue for further discussions!

jelhan:
> Which documentation exactly are you referring to? We are talking only about private props, aren't we?

I meant the API docs of `fade`, `showModal`, `inDom`, `paddingLeft`, `paddingRight`, `keyboard`, `size` and `backdropClose` properties of `<BsModal::Dialog>`. The properties are marked as public even so the full component is private. But for me it's more about avoiding wrong API docs as this could be very confusing later. E.g. `fade` is still documented as having a default value of `true`. Deleting the API docs for the internal component would be fine for me.

> But I agree, in hindsight this causes a lot of duplication. I love the approach you suggested. [...] I'll open up another issue for further discussions!

I'm wondering if we should wait with the transformations until we have a long term strategy for this question or accept the code duplication for now and remove it in a later refactoring? I'm tending towards the second option to not block a v4 release.

simonihmig:
> Deleting the API docs for the internal component would be fine for me.

Will do.

> I'm tending towards the second option to not block a v4 release.

Yeah, me too. I think we can do it even within the beta/RC period. I think/hope it shouldn't be that difficult. But also I would like to not slow down the current tag-less migration process.

addon/templates/components/bs4/bs-modal/dialog.hbs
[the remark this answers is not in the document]
simonihmig: Done. Worked beautifully! 

> Not sure if I missed something but it looked as if it's only used to trigger side-effects in the setter.

We need the element reference elsewhere, but not the setter anymore for side effects by using the render modifiers!

simonihmig:
Beside the code duplication issue we agreed to tackle elsewhere/later, I think your concerns have been addressed now. Thanks again!

PR #983, opened 2020-02-01
Fix progress bar role and ARIA attributes

simonihmig:


simonihmig:
> The changes to angle bracket component invocation in the tests aren't related to the bug, are they?

No, no changes there. I just copied the test with the newly added assertions from the other branch (#982), which already had ABI in that test...

PR #986, opened 2020-02-04
WIP: drop deprecated arguments for html attributes

jelhan:
This drops the *deprecated* arguments for HTML attributes. This affects the following components:

- [x] `<BsButton>`
- [x] `<BsForm>`
- [x] `<BsForm::Element>`

I like to suggest that we remove the other ones later in smaller chunks and decide individually which ones we might need to keep as arguments. E.g. `<BsForm>` has a `@disabled` arguments which sets `disabled` HTML attributes on all controls. I think such shortcuts needs to be discussed case by case and this will be easier if it's not mixed with other cases.

This is blocked by a bug in ember. Setting HTML attribute `type` on component invocation is not supported: https://github.com/emberjs/ember.js/issues/18232 Wondering that this wasn't reported yet by anyone in this repo as we have a deprecation that tells everyone to use `<BsButton type="submit" />`. Seems like no one addressed that deprecation yet.


addon/templates/components/common/bs-form/element.hbs
```diff
@@ -30,35 +30,12 @@
...
-      pattern=pattern
-      accept=accept
-      autocomplete=autocomplete
-      autocapitalize=autocapitalize
-      autocorrect=autocorrect
-      autosave=autosave
-      inputmode=inputmode
-      multiple=multiple
-      step=step
       form=form
```
simonihmig: What about this? Shouldn't this be also removed?

addon/templates/components/common/bs-form/element.hbs
```diff
@@ -30,35 +30,12 @@
...
       value=this.value
       id=this.formElementId
-      name=@name
       type=this.controlType
       label=@label
-      placeholder=placeholder
-      autofocus=autofocus
       disabled=disabled
       readonly=readonly
       required=required
```
simonihmig: Can't we remove `readonly` and `required` as well?

addon/components/base/bs-button.js
```diff
@@ -485,13 +460,12 @@ export default class Button extends Component {
           `After:\n` +
           `  <BsButton ${typeof value === 'boolean' ? attribute : `${attribute}="${value}"`} />`;
 
-        deprecate(
-          deprecationMessage,
+        warn(
```
simonihmig: Just asking myself if we shouldn't throw/assert here rather? The warning can be overseen, and I wouldn't expect people to have such rigid test coverage to detect this kind of missing attributes...

addon/templates/components/common/bs-form/element.hbs
jelhan: You mean the `form`? Is this related to arguments used for HTML attributes only? I thought the instance of `<BsForm>` is passed for some internal communication reasons. But to be honest I haven't had a look at these part of the code for quiet a while.

simonihmig: > You mean the form? Is this related to arguments used for HTML attributes only? 

Yeah. I believe so. It is just an ID, and not passed by default AFAICT. And it has been deprecated! :wink:

addon/components/base/bs-button.js
jelhan: I also considered throwing instead of warning but was afraid that in some weird edge cases a consumer likes to pass that arguments even so it's fine if they don't have any affect. E.g. for dynamic component invocation by `{{component}}` helper and some of the components require an argument with that exact name. Not sure so how realistic such a scenario is.

addon/templates/components/common/bs-form/element.hbs
jelhan: `disabled` and `readonly` are not that simple cases cause they allow a consumer to set that HTML attributes for all controls of a form at once, e.g.

```hbs
<BsForm @disabled={{true}} @readonly={{true}} as |form|>
  ...
</BsForm>
```

I'm tending to also remove that functionality but would prepare to discuss it separately.

`required` argument wasn't deprecated yet. Will open a PR to deprecate it for `v3` and remove it here.

jelhan: > `required` argument wasn't deprecated yet. Will open a PR to deprecate it for `v3`

See #987

addon/templates/components/common/bs-form/element.hbs
jelhan: You are right. Removed in 49868eea220dd2ad269fa56b5dc753782c8e2ba9.

What a confusing naming. Now that you pointed it out I remember that it also was totally confused about that one when deprecating it.

simonihmig:
CI is broken, due to some transitive dependency requiring node 10... 😩

simonihmig:
Replaced by #1004

PR #987, opened 2020-02-06
deprecate required arg of <BsForm::Element>

jelhan:
The `required` argument also adds an `.is-required` class: https://github.com/kaliber5/ember-bootstrap/blob/fd3a1cbd0d9ecf176c26d8de36755d86231ef01d/addon/components/base/bs-form/element.js#L394

To be honest I have no idea why this was added. Wasn't able to find that class in Bootstrap docs. The class name binding is their since nearly ever. It was added in March 2016 by this commit https://github.com/kaliber5/ember-bootstrap/commit/1199dd639aac24e71b123189776f75447160642a, which was part of #775. Not sure if we should mention this behavior explicitly in the deprecation message...

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -1093,6 +1093,7 @@ export default class FormElement extends FormGroup {
         ['name', 'foo'],
         ['pattern', '^[0-9]{5}$'],
         ['placeholder', 'foo'],
+        ['required', true],
         ['rows', '10'],
         ['spellcheck', true],
         ['step', '2'],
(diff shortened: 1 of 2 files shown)


simonihmig:
CI is broken, due to some transitive dependency requiring node 10... 😩

jelhan:
> CI is broken, due to some transitive dependency requiring node 10... weary

It's already reported and tracked here: https://github.com/babel/ember-cli-babel/issues/326 Guess it will be fixed soon.

simonihmig:
The ember-cli-babel issue has been resolved, so I restarted Travis. Now seeing some failed tests due to unexpected deprecations in tests...

jelhan:
> The ember-cli-babel issue has been resolved, so I restarted Travis. Now seeing some failed tests due to unexpected deprecations in tests...

Oh sorry. Seems like I missed half part of the work. Good that we have a CI. :laughing:

Issue #984, opened 2020-02-02
[RFC] Rethink our BS 3/4 build system


simonihmig:
## Motivation 

Based on [discussions](https://github.com/kaliber5/ember-bootstrap/pull/981#pullrequestreview-351928655) related to refactoring component to outerHTML semantics - as a stepping stone to later do a full migration to Glimmer components - it became clear that outerHTML semantics cause even more code duplication as before. We now have to have many more separate BS3/4 templates, as the elements root element cannot be tweaked anymore by the backing class (no more `@classNames`, `@classNameBindings` decorators), for slight difference in class names for example.

Also the often useless, but for our current build system required [re-exports by bs3/4 classes](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs4/bs-tooltip.js) are annoying. 

## Detailed Design

We will introduce macro helpers, that allow us to create (JS/Template) code that differs based on the configured Bootstrap version. These helpers would look like (and initially be) regular helpers, but provide the possibility to be compiled away at build time, so to not have any runtime performance/bundle size overhead. This approach is inspired by established patterns in the Ember ecosystem, like:
* `@ember/debug` / `@glimmer/env`
* [`ember-composability-helpers`](https://github.com/pzuraq/ember-compatibility-helpers)
* [`embroider` macros](https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#macro-system)

### JS

A `hasBootstrapVersion` util function is introduced:

```
import { hasBootstrapVersion } from 'ember-bootstrap/compatibility-helpers';

if (hasBootstrapVersion(3)) {
  // do something BS3 specific
}
```

By a custom Babel plugin, this could be compiled down to this when using BS3:

```
if (true) {
  // do something BS3 specific
}
```

For BS4:

```
if (false) {
  // do something BS3 specific
}
```

This would later be eliminated completely by `terser` as part of Ember's minification process for production builds, as this will be recognized as deed code. Just as `@ember/debug` stuff works.

### Handlebars

A `hasBootstrapVersion` helper is introduced:

```hbs
<div class="alert" {{if this.showAlert (if (has-bootstrap-version 3) "in" "show")}}">
```

With a custom AST transform, this could be compiled down to this when using BS3:

```hbs
<div class="alert" {{if this.showAlert "in"}}">
```

### Broccoli tree magic

[...]

PR #989, opened 2020-02-07
Helpers for Bootstrap version

simonihmig:
Initial basic (runtime) implementation of #984

### DIFF
--- a/tests/integration/helpers/has-bootstrap-version-test.js
+++ b/tests/integration/helpers/has-bootstrap-version-test.js
@@ -1,0 +1,21 @@
+import { module } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { testBS3, testBS4 } from '../../helpers/bootstrap-test';
+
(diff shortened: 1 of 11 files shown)


tests/integration/helpers/has-bootstrap-version-test.js
```diff
@@ -0,0 +1,21 @@
...
+import { testBS3, testBS4 } from '../../helpers/bootstrap-test';
+
+module('Integration | Helper | hasBootstrapVersion', function(hooks) {
+  setupRenderingTest(hooks);
+
+  testBS3('it works', async function(assert) {
+    await render(hbs`{{#if (has-bootstrap-version 3)}}Ok{{/if}}`);
+
+    assert.dom(this.element).hasText('Ok');
+  });
```
jelhan: Not sure if I missed something but shouldn't we also test the opposite?

```suggestion
  });
  
  testBS3('it works', async function(assert) {
    await render(hbs`{{#if (has-bootstrap-version 4)}}Ok{{/if}}`);

    assert.dom(this.element).hasNoText();
  });
```

And the same for BS4.

simonihmig: Added those.

PR #992, opened 2020-02-09
Use build helpers / first set of migration

simonihmig:
Use the build helpers for the already migrated tagless components, according to #984:
* alert
* button
* button group
* dropdown
* modal
* nav
* progress


simonihmig:
@jelhan this was more "mechanical" work (moving existing code around), so as long as tests pass, no thorough review is needed I believe! :)

jelhan:
This looks awesome! Makes the code much easier. If I got it right, you are tending to do the migration as soon as possible? Maybe it would be even easier to do the remaining outerHTML conversion directly with the build helpers rather than duplicating the code for BS3/BS4?

Restarted BrowserStack test, which is flickering as usual. :roll_eyes:

simonihmig:
> Maybe it would be even easier to do the remaining outerHTML conversion directly with the build helpers rather than duplicating the code for BS3/BS4?

Yes, indeed. I focused on these helpers as I realized that often when moving to outerHTML you would have to move `common` templates to `bs3`/`bs4`, only to move them back to `common` when implementing the helpers. So I think makes sense to start using them immediately!

PR #995, opened 2020-02-10
Deprecate readonly and disabled args of bs form element

jelhan:
Open questions:

1. If `@disabled` is `true` a `disabled` class is set on the `.form-group`: https://github.com/kaliber5/ember-bootstrap/blob/fd3a1cbd0d9ecf176c26d8de36755d86231ef01d/addon/components/base/bs-form/element.js#L394 This is not mentioned in the deprecation message. But I'm not sure if it should be. I wasn't able to find the reason for setting that class in the first place. I wasn't able to find a `.disabled` class in Boostrap 4 docs at all and in Bootstrap 3 docs I have only seen it for checkboxes and radios - and in that cases it was not set on the same element as the `.form-group`.

2. Should `@disabled` and `@readonly` arguments of `<BsForm>` are deprecated as well? They allow setting the `disabled` / `readonly` HTML attributes for all controls of a form. I'm tending to remove that feature but would like to get another opinion before doing it.

### DIFF
--- a/addon/components/base/bs-form/element.js
+++ b/addon/components/base/bs-form/element.js
@@ -391,7 +391,7 @@ const nonDefaultLayouts = A([
   @public
 */
 @templateLayout(layout)
-@classNameBindings('disabled:disabled', 'required:is-required', 'isValidating')
+@classNameBindings('__disabled:disabled', 'required:is-required', 'isValidating')
 export default class FormElement extends FormGroup {
   /**
    * Text to display within a `<label>` tag.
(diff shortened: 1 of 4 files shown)


simonihmig:
> If `@disabled` is true a disabled class is set on the .form-group:

I assume this was added a long time ago, to help with styling a disabled form element (not just the input). But I am fine with removing that. Was never documented IIRC.

> Should `@disabled` and `@readonly` arguments of <BsForm> are deprecated as well?

We used that in an app not too long ago (and I believe it was added for that use case) to quickly disable a whole form. Sure, it's not required, more syntactic sugar. But helps with reducing boilerplate and to not forget setting that argument for newly added form elements. So I would vote to not remove that! :wink:

jelhan:
Lets also wait here with merging until node 10 dependency issue is fixed.

simonihmig:
@jelhan some merge conflicts, could you please rebase?

simonihmig:
Thank you, and no worries! :)

PR #996, opened 2020-02-11
Convert Collapse to tag-less / outerHTML component

simonihmig:


addon/templates/components/common/bs-collapse.hbs
```diff
@@ -0,0 +1,11 @@
+<div
+  class="{{if this.collapse "collapse"}} {{if this.collapsing "collapsing"}} {{if this.showContent (if (has-bootstrap-version 4) "show" "in")}}"
```
jelhan: I think we should come up with a convention around `{{has-bootstrap-version}}` helper. I have the feeling that using the else part is confusing with that helper and doesn't help with readability. Even so it would be more verbose I think using it multiple times might be easier to read:

```suggestion
  class="{{if this.collapse "collapse"}} {{if this.collapsing "collapsing"}} {{if this.showContent (if (has-bootstrap-version 4) "show") (if (has-bootstrap-version 3) "in")}}"
```

This would also require less refactoring and would be less error prune if another bootstrap version is added.

addon/components/base/bs-collapse.js
```diff
@@ -123,13 +124,14 @@ export default class Collapse extends Component {
...
+
   setCollapseSize(size) {
     let dimension = this.get('collapseDimension');
 
     assert(`collapseDimension must be either "width" or "height". ${dimension} given.`, ["width", "height"].indexOf(dimension) !== -1);
 
-    this.element.style.width = dimension === 'width' && size ? `${size}px` : '';
-    this.element.style.height = dimension === 'height' && size ? `${size}px` : '';
+    this.set('collapseSize', size);
   }
```
jelhan: Maybe `collapseSize` and `setCollapseSize` should be refactored to native getter and setter for readability?

```js
@computed('_collapseSize')
get collapseSize() {
  return this._collapseSize;
}
set collapseSize(size) {
  let dimension = this.get('collapseDimension'); 

  assert(`collapseDimension must be either "width" or "height". ${dimension} given.`, ["width", "height"].indexOf(dimension) !== -1);

  this.set('_collapseSize', size);
}

_collapseSize: null;
```

Will be even easier to read as soon as we introduce tracked properties.

simonihmig: Tbh there is a ton of code that needs to be refactored, especially in that class (e.g. observers). But I intentionally wanted to make the changes as small as possible, only just as much to enable the main goal here, which is moving to outerHTML semantics. But again yes, the code is ages old and definitely needs some love! 😍

PR #1004, opened 2020-02-16
Drop deprecated arguments for html attributes of form components

simonihmig:
Replaces #986 

@jelhan took your PR, reverted the button#buttonType removal for now, to unblock the PR, and also removed the attributeBindings of control components.


jelhan:
Thanks a lot for pushing this one forward. Need to make sure that we don't forget to add warnings for the removed arguments of the control components. As we don't have deprecations for them, these ones would profit most from such warnings.

PR #1006, opened 2020-02-21
Remove Bootstrap version specific trees

simonihmig:
After having finished the refactor to use the new build helpers, we can drop all version specific trees and make the addon (almost) work like any other addon! :tada:


simonihmig:
Thank you. But you are not supposed to review PRs in your holidays, man! Take the time off and relax! :wink:

Issue #957, opened 2020-01-07
Road to v4


simonihmig:
### Breaking Changes:

- [x] Drop support for Ember < 3.15 (aka Octane) for RC releases / 3.16 (LTS) for stable release [#965] 

- [x] Drop node 8 support [#965] 

- [x] Refactor all components to tag-less/outerHTML components (only breaking in rare cases, e.g. using `ember-test-selectors` with curly invocation syntax)
  - [x] Accordion #1001 
  - [x] Alert #976
  - [x] Button #980 
  - [x] Button Group #985
  - [x] Carousel #1002 
  - [x] Collapse #996 
  - [x] Dropdown #980 
  - [x] Form #1005
  - [x] Modal #981 
  - [x] Navbar #997 
  - [x] Nav #988 
  - [x] ~Popover~
  - [x] Progress #982 
  - [x] Tabs #998 
  - [x] ~Tooltip~

- [x] Drop all arguments for HTML attributes (#826) and add warnings if they are still used

### Other

- [x] Refactor tests to angle brackets [#966] 
- [x] Use new build helpers (#984), remove separate `bs3`/`bs4` trees #1006 
- [x] Implement AST transforms for build helpers - at build time, with proper branch elimination 

### Release blockers

On the way we noticed some potential release blockers related to deprecation of arguments:

- [ ] Ember does not support `type` HTML attribute for `...attributes` if already set on the element: https://github.com/emberjs/ember.js/issues/18232 This breaks `<BsButton type="submit" />`.
- [ ] Ember Power Select does not support replacing some arguments by HTML attributes yet. This breaks `ember-bootstrap-power-select`: https://github.com/kaliber5/ember-bootstrap-power-select/issues/93

### Outlook

This allows refactoring all components step by step to Glimmer components during v4.x releases. This will cause breaking changes for consumers extending Ember Bootstrap classes. We don't consider extending the components as part of our public API but will not do such changes in patch releases.

simonihmig:
I started to work on https://github.com/ember-codemods/tagless-ember-components-codemod/issues/3, let's see how this turns out. Hopefully this will allow us to do the tag-less refactoring without much further effort on this side!

jelhan:
> Drop all (deprecated) properties for HTML attributes (#826)

[...]

PR #1009, opened 2020-02-27
Fix bs-carousel with async observers

jakesjews:
When using async observers the carousel is adding active classes to each visited slide and not removing them after the slide has transitioned.

### DIFF
--- a/tests/integration/components/bs-carousel-test.js
+++ b/tests/integration/components/bs-carousel-test.js
@@ -193,7 +193,7 @@ module('Integration | Component | bs-carousel', function(hooks) {
     );
     await waitTransitionTime();
     assert.notOk(getActivatedSlide(1), 'ltr has correct behavior');
-    assert.notOk(getActivatedSlide(1), 'ltr has correct behavior');
+    assert.notOk(getActivatedSlide(2), 'ltr has correct behavior');
     assert.ok(getActivatedSlide(3), 'ltr has correct behavior');
     this.stopCarousel();
   });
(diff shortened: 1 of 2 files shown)


tests/integration/components/bs-carousel-test.js
```diff
@@ -184,6 +192,8 @@ module('Integration | Component | bs-carousel', function(hooks) {
       hbs`{{#bs-carousel autoPlay=true interval=interval transitionDuration=transitionDuration ltr=false as |car|}}{{car.slide}}{{car.slide}}{{car.slide}}{{/bs-carousel}}`
     );
     await waitTransitionTime();
+    assert.notOk(getActivatedSlide(1), 'ltr has correct behavior');
+    assert.notOk(getActivatedSlide(1), 'ltr has correct behavior');
```
simonihmig: should this be like this?

```suggestion
    assert.notOk(getActivatedSlide(2), 'ltr has correct behavior');
```

simonihmig:
> How do we want to track that something needs to be backported (supported? 😜) to dev-v4? Guess adding it to roadmap issue would be best, wouldn't it?

Agree!

@jakesjews for context: we have a `dev-v4` branch in preparation of the new version, see #957. If you are interested in porting your changes here also to this branch, this would be awesome. The test should probably be integrated easily (e.g. by cherry-picking your commit here), but the component changes might need some manual fixes, as the module location has changed (no more `base` folder).

jakesjews:
@simonihmig no problem! I’ll make another 
PR to that branch as well.

tests/integration/components/bs-carousel-test.js
jakesjews: yup! I'll get that fixed

jakesjews:
PR for version 4 is up at https://github.com/kaliber5/ember-bootstrap/pull/1010

PR #1022, opened 2020-03-28
Refactor Button and ButtonGroup to Glimmer components

simonihmig:
Proof of concept of how easy/difficult it would be to refactor components:
* https://github.com/ember-codemods/ember-tracked-properties-codemod helped a bit
* used https://github.com/jkusa/ember-arg-types at first, but later decided to move most usages into the template (named args, custom helpers instead of CPs), and a custom, very simple `@args` decorator
* took a few hours, but was ok
* ~tests for `<BsButtonGroup> ` and `<BsDropdown>` are *expectedly* failing, as they access private properties of the button, which does not work anymore~
* tests for `<BsButton>` should be passing, without any changes. Does that make us confident enough that we can refactor component by component in minor version updates, without breaking changes? Or should we do a single major version release with all component refactored? 🤔

### DIFF
--- a/addon/components/bs-button-group.js
+++ b/addon/components/bs-button-group.js
@@ -2,6 +2,7 @@ import { action } from '@ember/object';
 import Component from '@glimmer/component';
 import { isArray } from '@ember/array';
 import arg from 'ember-bootstrap/utils/decorators/arg';
+import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
 /**
   Bootstrap-style button group, that visually groups buttons, and optionally adds radio/checkbox like behaviour.
@@ -61,6 +62,7 @@ import arg from 'ember-bootstrap/utils/decorators/arg';
   @extends Glimmer.Component
   @public
 */
+@deprecateSubclassing
 export default class ButtonGroup extends Component {
   /**
    * @property buttonComponent
--- a/addon/components/bs-button-group.js
+++ b/addon/components/bs-button-group.js
@@ -1,11 +1,7 @@
 import { action } from '@ember/object';
-import { tagName } from '@ember-decorators/component';
-import { equal } from '@ember/object/computed';
-import Component from '@ember/component';
-import { A, isArray } from '@ember/array';
-import sizeClass from 'ember-bootstrap/utils/cp/size-class';
-import defaultValue from 'ember-bootstrap/utils/default-decorator';
-import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
+import Component from '@glimmer/component';
+import { isArray } from '@ember/array';
+import arg from 'ember-bootstrap/utils/decorators/arg';
 
 /**
   Bootstrap-style button group, that visually groups buttons, and optionally adds radio/checkbox like behaviour.
@@ -62,20 +58,16 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
   @class ButtonGroup
   @namespace Components
-  @extends Ember.Component
+  @extends Glimmer.Component
   @public
 */
-@tagName('')
-@deprecateSubclassing
 export default class ButtonGroup extends Component {
-  ariaRole = 'group';
-
   /**
    * @property buttonComponent
    * @type {String}
    * @private
    */
-  @defaultValue
+  @arg
   buttonComponent = 'bs-button-group/button';
 
   /**
@@ -86,8 +78,6 @@ export default class ButtonGroup extends Component {
    * @default false
    * @public
    */
-  @defaultValue
-  vertical = false;
 
   /**
    * Set to true for the buttons to stretch at equal sizes to span the entire width of its parent. (BS3 only!)
@@ -107,8 +97,6 @@ export default class ButtonGroup extends Component {
    * @default false
    * @public
    */
-  @defaultValue
-  justified = false;
 
   /**
    * The type of the button group specifies how child buttons behave and how the `value` property will be computed:
@@ -146,14 +134,6 @@ export default class ButtonGroup extends Component {
    */
 
   /**
-   * @property isRadio
-   * @type boolean
-   * @private
-   */
-  @(equal('type', 'radio').readOnly())
-  isRadio;
-
-  /**
    * Property for size styling, set to 'lg', 'sm' or 'xs'
    *
    * Also see the [Bootstrap docs](https://getbootstrap.com/docs/4.3/components/button-group/#sizing)
@@ -162,11 +142,6 @@ export default class ButtonGroup extends Component {
    * @type String
    * @public
    */
-  @defaultValue
-  size = null;
-
-  @sizeClass('btn-group', 'size')
-  sizeClass;
 
   /**
    * This action is called whenever the button group's value should be changed because the user clicked a button.
@@ -177,27 +152,28 @@ export default class ButtonGroup extends Component {
    * @param {*} value
    * @public
    */
-  onChange() {}
 
   @action
   buttonPressed(pressedValue) {
+    if (!this.args.onChange) {
+      return;
+    }
     let newValue;
 
-    if (this.isRadio) {
+    if (this.args.type === 'radio') {
       newValue = pressedValue;
     } else {
-      if (!isArray(this.value)) {
-        newValue = A([pressedValue]);
+      if (!isArray(this.args.value)) {
+        newValue = [pressedValue];
       } else {
-        newValue = A(this.value.slice());
-        if (newValue.includes(pressedValue)) {
-          newValue.removeObject(pressedValue);
+        if (this.args.value.includes(pressedValue)) {
+          newValue = this.args.value.filter((v) => v !== pressedValue);
         } else {
-          newValue.pushObject(pressedValue);
+          newValue = [...this.args.value, pressedValue];
         }
       }
     }
 
-    this.onChange(newValue);
+    this.args.onChange(newValue);
   }
 }
--- a/addon/components/bs-button-group.js
+++ b/addon/components/bs-button-group.js
@@ -1,6 +1,6 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
-import { isArray } from '@ember/array';
+import { A, isArray } from '@ember/array';
 import arg from 'ember-bootstrap/utils/decorators/arg';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
@@ -174,6 +174,10 @@ export default class ButtonGroup extends Component {
           newValue = [...this.args.value, pressedValue];
         }
       }
+
+      // For compatibility we continue to return an EmberArray instance for now
+      // @todo this should be changed for the next major release!
+      newValue = A(newValue);
     }
 
     this.args.onChange(newValue);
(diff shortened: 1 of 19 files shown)


addon/components/bs-button.js
```diff
@@ -201,8 +177,7 @@ export default class Button extends Component {
    * @default false
    * @public
    */
-  @defaultValue
-  block = false;
+  @arg block = false;
```
jelhan: If I didn't missed something, `block` property is not needed. The usage in template could be changed to `{{@block}}`. See above.

addon/components/bs-button.hbs
```diff
@@ -1,9 +1,10 @@
 <button
   disabled={{this.__disabled}}
   type={{this.buttonType}}
-  class="btn {{if this.active "active"}} {{if this.block "btn-block"}} {{this.sizeClass}} {{this.typeClass}}"
+  class="btn {{if @active "active"}} {{if this.block "btn-block"}} {{bs-size-class "btn" @size}} {{bs-type-class "btn" @type default=(if (has-bootstrap-version 3) "default" "secondary") outline=@outline}}"
```
jelhan: As `this.block` is just an alias with default property of `false`, `{{if this.block "btn-block"}}` should be the same as `{{if @block "btn-block}}`.

```suggestion
  class="btn {{if @active "active"}} {{if @block "btn-block"}} {{bs-size-class "btn" @size}} {{bs-type-class "btn" @type default=(if (has-bootstrap-version 3) "default" "secondary") outline=@outline}}"
```

addon/components/bs-button.js
```diff
@@ -432,66 +370,32 @@ export default class Button extends Component {
...
+    if (!this.args.bubble) {
       e.stopPropagation();
     }
   }
-
-  init() {
-    super.init(...arguments);
-
-    // deprecate arguments used for attribute bindings only
-    runInDebug(() => {
```
jelhan: The warning that the arguments have been removed shouldn't be removed, should it?

addon/components/bs-button.hbs
```diff
@@ -1,9 +1,10 @@
 <button
   disabled={{this.__disabled}}
   type={{this.buttonType}}
```
jelhan: This could be also changed to avoid a local variable:

```suggestion
  type={{if @buttonType @buttonType "button"}}
```

It's a little bit more verbose but doesn't require a switch to another file in order to get it. It's an alternative pattern for default values.

addon/components/bs-button.js
simonihmig: Yes, I did that before, but that broke with a build error that `@block` is a reserved keyword! 😝

addon/components/bs-button.hbs
simonihmig: See above.

addon/components/bs-button.js
simonihmig: Well, not sure when this could be released. My initial assumption was as part of v5, so in that case maybe we can remove those warnings as they have been part of the v4.x cycle. But not sure, again this PR is not really ready to be merged, more an experiment to get some learnings...

addon/components/bs-button.hbs
simonihmig: Yes, that would work. I thought about `{{or @buttonType "button"}}` but `or` is not available be default yet.

jelhan: Yeah. That's even shorter syntax for default values in template. I like that one.

addon/components/bs-button.js
jelhan: Maybe that's a good indication that we should rename? E.g. @isBlock` or `@blockLevel`.

addon/components/bs-button.hbs
simonihmig: Yes, but wo don't have this helper available yet. Would need `ember-truth-helpers`, until we have RFC562. But then I would rather add a simple `{{bs-or}}` or better maybe `{{bs-default}}` (I don't really like abusing boolean logic for defaults :wink::)

addon/components/bs-button.js
simonihmig: But it's a public argument, so that would be a breaking change!

simonihmig:
I finally was able to get back to this:
* rebased
* refactored also ButtonGroup and Dropdown/Button
* all tests pass now (without having touched them, yeah!)
* CI is failing for Ember Canary, but for some form tests, so apparently unrelated

If we are confident enough to release this - and other future Glimmer refactorings - as a regular non-breaking update (seems so judging by not having touched any tests, but who knows if we miss some edge cases?) in the 4.x cycle (maybe with a pre-release like 4.1.0-beta?), then I should bring back the subclassing deprecations as discussed [here](https://github.com/kaliber5/ember-bootstrap/pull/1022#discussion_r400466839)

@jelhan thoughts?

addon/components/bs-button.js
simonihmig: I added them again now!

addon/components/bs-button.hbs
simonihmig: I did some more work on refactoring `<BsAlert>`, and there you have both the template and the class read some args with a default value required. In that case `@arg` seems much better, as you have a single source of truth. Same happens when using the same arg multiple times in a template, as you can easily introduce errors by not having the (same) default value used everywhere...

Related: https://github.com/emberjs/rfcs/issues/479

addon/components/bs-button-group/button.js
```diff
@@ -12,33 +10,26 @@ import defaultValue from 'ember-bootstrap/utils/default-decorator';
  @private
  */
 export default class ButtonGroupButton extends Button {
-  '__ember-bootstrap_subclass' = true;
```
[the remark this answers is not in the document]
simonihmig: It is, good catch!

addon/components/bs-button-group/button.js
```diff
@@ -12,33 +10,26 @@ import defaultValue from 'ember-bootstrap/utils/default-decorator';
...
-
   /**
    * @property groupValue
    * @private
    */
```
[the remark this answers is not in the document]
simonihmig: Those arguments are still passed, from the ButtonGroup component that yields the button. They are just not used on the component class, but as named arguments in the template. I thought we can still leave the doc blocks to document what arguments the component expects.

addon/components/bs-button.js
```diff
@@ -436,39 +376,39 @@ export default class Button extends Component {
    */
   @action
   handleClick(e) {
-    let onClick = this.onClick;
-    let preventConcurrency = this.preventConcurrency;
+    let onClick = this.args.onClick;
+    let preventConcurrency = this.args.preventConcurrency;
 
-    if (onClick === null || onClick === undefined) {
+    if (!onClick) {
```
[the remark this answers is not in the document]
simonihmig: :+1:

addon/components/bs-button-group.js
```diff
@@ -179,27 +154,28 @@ export default class ButtonGroup extends Component {
...
     let newValue;
 
-    if (this.isRadio) {
+    if (this.args.type === 'radio') {
       newValue = pressedValue;
     } else {
-      if (!isArray(this.value)) {
-        newValue = A([pressedValue]);
+      if (!isArray(this.args.value)) {
+        newValue = [pressedValue];
```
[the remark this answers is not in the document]
simonihmig: Hm, yeah, you might be right. My first thought was that that does not really belong into the new Octane world. But indeed it could be a breaking change. I'll revert for now, and add a comment to change this for the next major version.

addon/components/bs-button.js
```diff
@@ -493,7 +433,7 @@ export default class Button extends Component {
         warn(
           warningMessage,
           // eslint-disable-next-line ember/no-attrs-in-components
```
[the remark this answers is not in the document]
simonihmig: Good catch!

addon/components/bs-button.js
```diff
@@ -436,39 +376,39 @@ export default class Button extends Component {
...
+    let preventConcurrency = this.args.preventConcurrency;
 
-    if (onClick === null || onClick === undefined) {
+    if (!onClick) {
       return;
     }
 
     if (!preventConcurrency || !this.isPending) {
-      let promise = onClick(this.value);
+      let promise = onClick(this.args.value);
```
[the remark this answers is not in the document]
simonihmig: Yeah, so the ButtonGroup's version of a button needs a value, as that determines the value of the group (e.g. an array of all selected button values). The base button does not really need a value, I agree. I think we can refactor that eventually, but only separately as a breaking change!

simonihmig:
@jelhan thanks for the 2nd review! Some how I missed this...

I think I addressed your feedback, thanks for that! Rebased again, after some merge conflicts (#1221). And fixed the new submitButton feature, which was broken by these changes. Hope everything turns green now! 🤞

simonihmig:
✅  All green! 🚀

I'll do another bugfix release first before merging this, so we get that fix out without coupling that to eventually (unforeseen) breaking changes. 

But after that, I'll like to get this merged, if @jelhan you have nothing left, to not need another rebase! Also I have another branch lined up already that refactored BsAlert, but builds on top of this, so I'll wait to get this merged before opening the PR...

simonihmig:
> I'll do another bugfix release first before merging this

Done! :)

jelhan:
> I'll do another bugfix release first before merging this, so we get that fix out without coupling that to eventually (unforeseen) breaking changes.

I noticed that I missed to update include / exclude feature when implementing form submit button feature. Provided a fix in #1223.

> But after that, I'll like to get this merged, if @jelhan you have nothing left, to not need another rebase! Also I have another branch lined up already that refactored BsAlert, but builds on top of this, so I'll wait to get this merged before opening the PR...

Awesome. 👏 Let's ship it.

Are you planning to bundle multiple refactored to `@glimmer/component` in the same minor? Or do you want to ship each one as soon as it's ready?

addon/components/bs-button.js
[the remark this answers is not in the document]
jelhan: I'm still thinking that this is better done baby curling the arguments using `{{fn}}` helper.

For the `<Button group::Button>` it's not that obvious cause how it extends from `<BsButton>`. Instead of invoking the `<BsButton>` component in a template it extends from its class.  To be honest I don't like that pattern.

I think the `<BsButtonGroup::Button>` should be refactored to something like this:

```hbs
<BsButton
  @active={{this.isActive}}
  @onClick={{fn @onClick @value}}
  @type={{@type}}
  @buttonType={{@buttonType}}
  ... attributes
  as |button|
>
  {{yield button}}
</BsButton>
```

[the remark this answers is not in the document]
jelhan: Don't get me wrong. This is nothing which is blocking this PR. I will see if I have some time in the next days to provide a PR showcasing such a refactoring.

simonihmig:
> Are you planning to bundle multiple refactored to @glimmer/component in the same minor? Or do you want to ship each one as soon as it's ready?

As I said, I have BsAlert already lined up, so would at least add that too. 

Though about maybe releasing a 4.2.0-beta version for people to test this, before a stable v4.2.0? In that case, do you think it makes sense to bundle more refactorings into it, so to have less work testing these kind of prereleases?

jelhan:
> Though about maybe releasing a 4.2.0-beta version for people to test this, before a stable v4.2.0? In that case, do you think it makes sense to bundle more refactorings into it, so to have less work testing these kind of prereleases?

Releasing directly as v4.2.0 would be fine for me as well. We can still revert if needed. But in both cases I would release as soon as possible. Better catch regressions not covered by our testing suite as early as possible to not waste time.

PR #1026, opened 2020-04-08
WIP: transpile object spread syntax

jelhan:
I faced an issue that an app was throwing a syntax error in Microsoft Edge after I removed IE11 from build list. Adding the babel transform for it to the addon fixed it.

Object spread syntax is at least used here: https://github.com/kaliber5/ember-bootstrap/blob/44c393a2cd991060a4dab7e55afe5220cf493f54/addon/utils/default-decorator.js#L13

For some reason I was not able to reproduce the issue if doing `yarn link` or using `file:` or `link:` protocol. That wasn't helpful for debugging...

### DIFF
--- a/index.js
+++ b/index.js
@@ -343,5 +343,11 @@ module.exports = {
 
     whitelist.forEach(_addToWhitelist);
     return list;
+  },
+
+  options: {
+    babel: {
+      plugins: ['transform-object-rest-spread']
+    }
   }
 };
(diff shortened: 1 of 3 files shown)


jelhan:
Actually I'm not sure if it really fixed the issue or if I just had a `yarn link` that I wasn't aware of. Please ignore it for now.

PR #1035, opened 2020-04-14
Merge branch 'origin/master' into dev-v4

simonihmig:
To sync them for following merge of dev-v4 back to master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,38 @@
 
+## 3.1.3 (2020-04-11)
+
+#### Bug Fixes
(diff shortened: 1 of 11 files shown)


jelhan:
Wouldn't it be easier to port the few changes to `master`, which are applicable for v4 since we branched out `dev-v4` over to `dev-v4`? If I recall correctly it should only be very few pull requests. I think some where already ported to `dev-v4`.

simonihmig:
I wanted to make sure all CI tests pass with the merged result. And this way I believe we should have avoided any merge conflicts when merging dev-v4 to master...

Issue #1033, opened 2020-04-11
Add prettier


simonihmig:
* add ESLint config
* run `eslint --fix` automatically pre-commit using `husky` and `lint-staged`

jelhan:
I take this one. Have an automated script to set it up. 🤣

jelhan:
Only talking about using Prettier for JavaScript aren't we? Haven't used it yet for Handlebars / Glimmer templates but heard that it's not that stable yet. Would love to test it out but not sure if this is the best repository to start with. Or do you have good experience with it?

simonihmig:
> Only talking about using Prettier for JavaScript aren't we?

Yes.

jelhan:
> automatically pre-commit using `husky` and `lint-staged`

I'm not convinced by this one:

- Running ESLint and prettier is an expensive task. `yarn lint:js --fix` needs 10 seconds on my machine for this repo.
- Running prettier in a pre-commit hook is not needed for my setup. I have configured my editor (VS Code) to fix linting errors when a file is saved. This is very cheap (read: performant) as it's only run against one file and not all.

I think my setup is a very common one. I would suggest that we don't add the precommit hook but document in `CONTRIBUTING.md` how one could configure the local development environment to run prettier automatically? We can list both options: Do it in editor on save or as a local pre-commit hook.

jelhan:
Oh haven't read carefully enough. `lint-staged` would also only run the linting against the files which will be committed. So there wouldn't be a noticeable performance impact except for commits touching a lot of files. Such commits doesn't happen that often.

Still not seeing a need for it as I expect most developers to have their editors configured to fix linting issues on safe. But it would help with onboarding new developers for sure.

jelhan:
Added husky and lint-staged to #1038. Directly noticed that it's slower even if touching only a few files. Committing a change to only one file takes 1.5 seconds with husky + lint-staged. It's only a few milliseconds without it. It's basically adding a waiting step that wasn't there before. To be honest I don't like it. But I can live with it if you prefer to have it in.

simonihmig:
> Still not seeing a need for it as I expect most developers to have their editors configured to fix linting issues on safe. But it would help with onboarding new developers for sure.

[...]

PR #1038, opened 2020-04-16
setup prettier

jelhan:
Introduces Prettier.

It's integrated as an ESLint plugin. So `yarn lint:js` will lint that all files are formatted with Prettier. `yarn lint:js --fix` will prettify the code.

It uses the same prettier configuration as ember source. I think just following that one helps to avoid discussions about best configuration. Also the source code will look similar to the one of other projects in ember ecosystem.

It adds a `.gitattribute`. This enforce a specific git configuration for this repository. It addresses a common issue if using Prettier on Windows.

It adds a pre-commit hook as listed in the requirements in #1033. But to be honest I don't like it. Have explained reasons in #1033.


.eslintrc.js
```diff
@@ -5,19 +5,13 @@ module.exports = {
...
   },
-  plugins: [
-    'ember',
-    'hbs'
-  ],
-  extends: [
-    'eslint:recommended',
-    'plugin:ember/recommended'
-  ],
+  plugins: ['ember', 'hbs'],
```
simonihmig, line 11: Shouldn't `prettier` be also added as a plugin here? Not really sure though, but that's what we have in the common setup of our apps, and Ember seems to have it also: https://github.com/emberjs/ember.js/blob/master/.eslintrc.js#L14

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Git may be configured to convert the line endings from LF to CRLF globally.
+# This conflicts with ESLint setup, which lints for LF file endings. It causes
+# countless linting errors. As all relevant text editors support LF line
+# endings these days, disabling the feature should be safe.
+core.autocrlf=false
```
simonihmig, line 5: :+1:

.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = {
+  singleQuote: true,
+  trailingComma: 'es5',
+  printWidth: 100,
+};
```
simonihmig, line 7: *Although* the core idea of prettier is to prevent bikeshedding, I have to do it 😝
We have this in our common `.prettierrx.yml`:

```yaml
arrowParens: always # It's convenient that you don't have to add them by hand when you need to add more args or a type
singleQuote: true   # Enabled to align with EmberJS Prettier config
printWidth: 120     # Per team's agreement
trailingComma: es5  # Trailing commas makes editing easier, diffs cleaner. `es5` option aligns with EmberJS Prettier config
```

What do you think? Re: `printWidth` I think everybody has a wide enough monitor to support this. Things often got harder to read from our experience when prettier forced line wraps to eagerly...

jelhan: Prettier changed the default value for arrow function parens to always include them in v2.

Only meaningful difference is print width if I didn't missed something. Prettier explicitly recommends against a print width of 120:

> With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.

See prettier docs for more details: https://prettier.io/docs/en/options.html#arrow-function-parentheses

I try to avoid having a personal opinion on that topics at all. 😜

.eslintrc.js
jelhan: I tried to follow the recommended configuration for ESLint integration in Prettier docs: https://prettier.io/docs/en/integrating-with-linters.html#recommended-configuration

simonihmig: But doesn't the section above it say you should add prettier to the `plugins` array: https://prettier.io/docs/en/integrating-with-linters.html#use-eslint-to-run-prettier. Or is that happening implicitly when using `prettier/recommended`? The docs seem to be not very clear, at least for me...

.prettierrc.js
simonihmig: > Prettier explicitly recommends against a print width of 120:

But also against one of 100! :wink:

IIRC one thing that was for example hard to read was function invocations/definitions with a couple of arguments, with longer (speaking) names. When breaking this up into one line per argument, this became difficult to read. As when you are skimming through code to see what's happening, you want to see which functions are called, but not necessarily with which arguments (at this point). But having the function name and all arguments all split up into separate lines (with the arguments just having those 2 more spaces of indentation) made it hard to immediately see what's a function name and what's an argument. Just one example...

So when you don't have strong feelings, I would suggest to use that width of 120. 🙂

.eslintrc.js
jelhan: The recommended configuration combines both steps mentioned before. At least if I got it right. The config is working fine for a bunch of apps and addons since a few months.

simonihmig: Ok, convinced! :)

.prettierrc.js
jelhan: Changed `printWidth` to `120` in 822b5bb3af41da2d2df2446f85fc1c4884e523f4.

jelhan: I also removed `trailingComma: 'es5'` in 6a717e3aebe483b9183137b279fbf327bec18d9c cause that's the default value for pretter@^2.0. Only had it included due to not diverge from configuration used in Ember.

jelhan:
Addressed outstanding review comments and fixed merge conflicts by a rebase. Ready to be merged.

simonihmig:
@jelhan after merging the other PR, this now has some merge conflicts! Sorry! :)

jelhan:
> @jelhan after merging the other PR, this now has some merge conflicts! Sorry! :)

Rebased again. Luckily not much work involved. :smile:

simonihmig:
Thanks! :tada:

PR #1076, opened 2020-05-13
Remove debug code from production builds

### LINKED ISSUE #1053: Remove use of `runInDebug()`
Replace with `import { DEBUG } from '@glimmer/env';`, see https://github.com/emberjs/ember.js/issues/18912.

simonihmig:
Fixes #1053


jelhan:
Are we using different techniques for other warnings or do we miss warnings for other arguments that we have removed?

simonihmig:
Good that you mention this, there was another warning in the form component, that wasn't wrapped at all. Added that there too!

Issue #1089, opened 2020-05-25
Q: value missing?


basz:
Why is value not defined (but documented) here?

https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form/element.js#L274

jelhan:
What do you mean by "_not defined_"? The property is defined here: https://github.com/kaliber5/ember-bootstrap/blob/235a2433ccee5defce136b8ee9099cde97268e2e/addon/components/bs-form/element.js#L858

basz:
right... thx... seeing some issues element value not being updated, so that's why i'm looking at sources

jelhan:
Actually it seems to be unnecessary complicated. This could be simplified as soon as the component is rewritten as a Glimmer Component to something like this:

```js
get value() {
  return this.args.hasOwnProperty('value') ? this.args.value : this.model[this.property];
}
set value(value) {
  if (this.args.hasOwnProperty('value')) {
    this.args.value = value;
  } else {
    this.model[this.property] = value;
  }
}
```

jelhan:
@simonihmig Just noticed that this two-way binding is not supported anymore after refactoring `<BsForm::Element>` to `@glimmer/component`. `this.args` and all it's values is immutable for glimmer components. We should consider dropping that feature in v4. What do you think?

basz:
I use this.value inside the my custom controls to receive values. And use this.onChange(value) to send values back.

DDAU. So yes, I think we should drop that feature.

simonihmig:
Closing this, as we agreed to land #1097!

PR #1097, opened 2020-06-01
Remove leftover two-way binding of form element's `@value` argument

basz:
I've noticed some issues when I use form in combination with models wrapped in ember-changeset changesets.

When I have a Changeset that has a model of 

```
{
  propertyPath: aNonPrimitiveValue
 }
```

And a custom element that calls this.onUpdate(newValue) the changes appear to propagate into the changeset correctly, (changeset.changes has the changes) but unfortunately the new value is not transmitted back into the control on subsequent edits of the value. (It works only once)

You can see this behavior in ember.bushbaby.nl...

After much trial and error I have found out the element creates a dynamic property and aliases it to `model.${this.property}` and that seems to be the culprit. I think it is the alias that does not work as reliably...

I have created this PR to avoid dynamic assignment and aliasing. In my code base, my forms now work as expected... It is my hope it will make the 4.0 release.

also see #1089 for some discussion about two-way binding.

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -1,6 +1,6 @@
 import { layout as templateLayout } from '@ember-decorators/component';
-import { alias, and, equal, gt, notEmpty, or } from '@ember/object/computed';
-import { action, computed, defineProperty } from '@ember/object';
+import { and, equal, gt, notEmpty, or } from '@ember/object/computed';
+import { action, computed, get } from '@ember/object';
 import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 import { isBlank, typeOf } from '@ember/utils';
@@ -271,6 +271,24 @@ export default class FormElement extends FormGroup {
    * @property value
    * @public
    */
+  get value() {
+    if (this.property && this.model) {
+      return get(this.model, this.property);
+    }
+    return this._value;
+  }
+
+  set value(value) {
+    this._value = value;
+  }
+
+  /**
+   * Cache for value
+   * @type {null}
+   * @private
+   */
+  @defaultValue
+  _value = null;
 
   /**
    The property name of the form element's `model` (by default the `model` of its parent `Components.Form`) that this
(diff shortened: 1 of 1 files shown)


addon/components/bs-form/element.js
```diff
@@ -271,6 +271,24 @@ export default class FormElement extends FormGroup {
...
+  get value() {
+    if (this.property && this.model) {
+      return get(this.model, this.property);
+    }
+    return this._value;
+  }
+
+  set value(value) {
+    this._value = value;
+  }
```
jelhan, line 283: Wondering if Ember Bootstrap sets `value` itself or if the setter is only needed to support `@value` argument until the component is refactored to `@glimmer/component`. In the latter case I would move the assertion that not `@value` and `@property` could not be set at the same time to this setter. It would make it a little bit easier to follow the code in my opinion.

simonihmig: It's the latter, and yes, I agree.

basz:
@jelhan I'm not sure what you want the PI docs to say here... 

I have noticed it's two way bounded when an <Input @value=.../> is used... Which is by default for regular types... (text, email, etc). That is something the build-in input component does. I've pulled my hair a few times before I realised that. That thing is not Data Down Actions Up atm...

jelhan:
> I have noticed it's two way bounded when an <Input @value=.../> is used... Which is by default for regular types... (text, email, etc).

I'm not sure what you are referring to. As far as I'm aware we don't use `<Input>` in Ember Bootstrap. Do we?

To be honest I think `<Input>`, `<Textarea>` and these built-in components should be deprecated cause their two-way bindings is confusing. But there doesn't seem to be a consensus for that yet. If I recall correctly the argument was that we need some built-in `{{set}}` helper before we can deprecate that ones to provide a good upgrade path.

basz:
No we don’t. Sorry for that...

basz:
Agree that should be deprecated or refactored

basz:
Can you write that api doc? I’ll put it in the PR

simonihmig:
I'll merge this now, and file a API docs PR separately. Thanks @basz for your work here!

PR #1098, opened 2020-06-02
Reset style attribute after transition of BsCollapse

basz:
I noticed that the height attribute of the accordion component wasn't removed after the transition was finished, which would lead to issues when the dimensions of the contained content would change.

Looks like we were sending 'px' to the style modifier were is should have been ''.

This PR fixes that.

### DIFF
--- a/addon/components/bs-collapse.js
+++ b/addon/components/bs-collapse.js
@@ -3,12 +3,13 @@ import layout from 'ember-bootstrap/templates/components/bs-collapse';
 import { alias, and, not } from '@ember/object/computed';
 import { addObserver } from '@ember/object/observers';
 import Component from '@ember/component';
-import { isPresent } from '@ember/utils';
+import { isPresent, isNone } from '@ember/utils';
 import { next } from '@ember/runloop';
 import { camelize } from '@ember/string';
 import transitionEnd from 'ember-bootstrap/utils/transition-end';
 import { assert } from '@ember/debug';
 import defaultValue from 'ember-bootstrap/utils/default-decorator';
+import { computed } from '@ember/object';
 
 /**
   An Ember component that mimics the behaviour of [Bootstrap's collapse.js plugin](http://getbootstrap.com/javascript/#collapse)
@@ -91,6 +92,29 @@ export default class Collapse extends Component {
    * @public
    */
   expandedSize = null;
+
+  /**
+   * Calculates a hash for style attribute.
+   */
+  @computed('collapseDimension', 'collapseSize')
+  get cssStyle() {
+    if (isNone(this.collapseSize)) {
+      return {};
+    }
+
+    if (this.transitionDuration) {
+      const transitionDurationS = this.transitionDuration / 1000;
+
+      return {
+        [this.collapseDimension]: `${this.collapseSize}px`,
+        transition: `${this.collapseDimension} ${transitionDurationS}s ease`,
+      };
+    }
+
+    return {
+      [this.collapseDimension]: `${this.collapseSize}px`,
+    };
+  }
 
   /**
    * Usually the size (height) of the element is only set while transitioning, and reseted afterwards. Set to true to always set a size.
--- a/addon/components/bs-collapse.js
+++ b/addon/components/bs-collapse.js
@@ -102,15 +102,6 @@ export default class Collapse extends Component {
       return {};
     }
 
-    if (this.transitionDuration) {
-      const transitionDurationS = this.transitionDuration / 1000;
-
-      return {
-        [this.collapseDimension]: `${this.collapseSize}px`,
-        transition: `${this.collapseDimension} ${transitionDurationS}s ease`,
-      };
-    }
-
     return {
       [this.collapseDimension]: `${this.collapseSize}px`,
     };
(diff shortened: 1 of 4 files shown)


basz:
had a change to look at this one?

jelhan:
I didn't had the time yet to look into that bug. It would be helpful if you can provide a test that is failing on latest master but passing with your fixes.

addon/templates/components/bs-collapse.hbs
```diff
@@ -3,9 +3,9 @@
   ...attributes
   {{ref this "_element"}}
   {{style
-    width=(if (bs-eq this.collapseDimension "width") (concat this.collapseSize "px"))
-    height=(if (bs-eq this.collapseDimension "height") (concat this.collapseSize "px"))
+    width=(if (bs-eq this.collapseDimension "height") (if this.width (concat this.collapseSize "px") ""))
```
simonihmig, line 6: This line *seems* wrong. Shouldn't this be like this?
```suggestion
    width=(if (bs-eq this.collapseDimension "width") (if this.collapseSize (concat this.collapseSize "px") ""))
```

jelhan: I also think we shouldn't pass an empty string.. ember-style-modifier should remove the CSS style if `undefined` is passed. So it would be:

```suggestion
    width=(if (bs-eq this.collapseDimension "width") (if this.collapseSize (concat this.collapseSize "px")))
```

`{{and}}` helper would improve readability a lot:

```hbs
    width=(if (and this.collapseSize (bs-eq this.collapseDimension "width")) (concat this.collapseSize "px"))
```

Sad that the [logical operators RFC](https://github.com/emberjs/rfcs/pull/562) seems to be stale.

basz: i'll fix those mistakes. but I can't rely on the `and` helper can I. That's not build in or is it?

jelhan: > but I can't rely on the `and` helper can I. That's not build in or is it?

That's right. It was more like a side note as I realized how much it would improve readability in this case. It's not available yet in ember. The RFC is not even merged yet. We don't want to introduce another dependency (ember-truth-helpers) for it.

basz:
Ok!

Done for the day. Struggling with getting a test that reads the style.height. Doesn’t seem to be set at all in the test. Which must be wrong... well a bit of rest for now

basz:
I think this is a better approach and it seems to work perfectly in my app. Both on BsCollapse and a   BsAccordion. However I can get the test to work and I don't understand why, so if anyone could give me some help there that would be great...

basz:
The utility function `transitionEnd` inside used in the collapse show/hide method contains the following.

```
if (Ember.testing) {
    duration = 0;
}
```

[see](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/utils/transition-end.js#L11)

So this [block](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-collapse.js#L187-L202) is called in the same `tick` and that prevents my introduced `get cssStyle` with computed properties from being called. Therefore the tests are failing... Commenting that snipped makes my tests pass... What's would be a safe way to work around this without increasing the run time of the tests too much. (eg. setting duration=5 still fails, but 50 works...)

addon/components/bs-collapse.js
```diff
@@ -92,6 +93,29 @@ export default class Collapse extends Component {
...
+    if (isNone(this.collapseSize)) {
+      return {};
+    }
+
+    if (this.transitionDuration) {
+      const transitionDurationS = this.transitionDuration / 1000;
+
+      return {
+        [this.collapseDimension]: `${this.collapseSize}px`,
+        transition: `${this.collapseDimension} ${transitionDurationS}s ease`,
```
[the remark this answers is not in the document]
basz: I did this because the css-duration must/should be kept in sync with components transition-duration and figured this would be an attempt to avoid that. But yes not really optimal because of ease, but also css-property prefixing might be an issue...

tests/integration/components/bs-collapse-test.js
```diff
@@ -71,6 +71,25 @@ module('Integration | Component | bs-collapse', function (hooks) {
...
   });
 
+  test('after collapsing/expanding height/width is set correctly ', async function (assert) {
+    this.set('collapsed', true);
+    await render(
+      hbs`<BsCollapse @expandedSize={{200}} @transitionDuration={{500}} @collapsed={{collapsed}}><p>Just some content</p></BsCollapse>`
+    );
+    this.set('collapsed', false);
+
+    assert.equal(this.element.querySelector('div.collapsing').style.height, '200px', 'should have height of 200px');
```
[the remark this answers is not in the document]
basz: Callbacks *are* being called, and collapseSize is *also* set correctly. But I believe what is  happening is that the collapseSize is set to 200px, then to null. The cssStyle getter is getting a notification only for the latter. Therefore the test never detects '200px'...

addon/components/bs-collapse.js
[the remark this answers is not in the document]
simonihmig: Yeah, you definitely have a valid concern here! @jelhan brought this up a while ago: https://github.com/kaliber5/ember-bootstrap/issues/839

But we should tackle this in a separate PR at least, and for all components where this applies!

simonihmig:
@basz I pushed a commit to you PR that should fix the failing test.

Indeed the issue was the `waitForTransitionEnd` running with a duration of 0 in tests, so I added a way to override that in tests if needed.

Also the suggestion to use `qunit-dom`'s `hasStyle` proved to be not ideal, as it uses `getComputedStyle()`. So when making an assertion in the middle of the transition, it would return a computed height of ~100px, which is not what we want to assert against. So I changed that to look directly at the `style` attribute, using `hasAttribute`.

Once tests pass in CI, I'll merge this. Thanks again!

basz:
thanks!

PR #1104, opened 2020-06-09
Remove support for subclassing component classes 

simonihmig:
Subclassing makes non-breaking refactorings difficult/impossible, e.g. the current move to tagless components, or further to Glimmer components in the future. Therefore we only consider template invocations as part of a components public API.

Addons like `ember-bootstrap-cp-validations` still make use of subclassing, and until we come up with better APIs to support their use cases that don't need subclassing, we offer them an escape hatch to silence the warnings. This seems acceptable to me, as they can be considered "privileged" addons as *we* are in control to make sure they continue to work when changes are introduced that would break subclassing.

This PR could be backported to the `v3` branch to publish a new v3 release that also contains the deprecation messages.

### DIFF
--- a/addon/utils/deprecate-subclassing.js
+++ b/addon/utils/deprecate-subclassing.js
@@ -5,20 +5,20 @@ export default function deprecateSubclassing(target) {
   if (DEBUG) {
     const originalInit = target.prototype.init;
 
-    if (originalInit) {
-      target.prototype.init = function () {
-        deprecate(
-          `Extending from ember-bootstrap component classes is not supported, and might break anytime. Detected subclassing of <Bs${target.name}> component.`,
-          // the `__ember-bootstrap_subclass` flag is an escape hatch for "priviliged" addons like validations addons that currently still have to rely on subclassing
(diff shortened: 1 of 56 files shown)


addon/utils/deprecate-subclassing.js
```diff
@@ -0,0 +1,24 @@
+import { DEBUG } from '@glimmer/env';
+import { deprecate } from '@ember/debug';
+
+export default function deprecateSubclassing(target) {
+  if (DEBUG) {
+    const originalInit = target.prototype.init;
+
+    if (originalInit) {
+      target.prototype.init = function () {
+        deprecate(
```
simonihmig, line 10: Note that I used `deprecate()` here instead of `warn()`, although this is strictly not a real deprecation. But this allows us to test it easily, and might also be easier to process, as the deprecation message are aggregated in the Ember inspector, and can be silenced if needed. Does that make sense?

jelhan:
Fully agree. I expect to see some use cases that aren't supported yet without extending the components and overriding internals. But I think the time is way better invested by improving invocation API and/or extracting more reusable low level primitives than trying to keep that internal APIs stable.

PR #1106, opened 2020-06-09
Refactor BsModalSimple to not extend from BsModal

### LINKED ISSUE #1105: Refactor BsModalSimple to not extend from BsModel
For whatever historical reasons this is the case, although BsModalSimple also *uses* BsModal (in the template).

simonihmig:
The OOP relationship between those two is now of type `has-a` instead of `is-a`, i.e. BsModalSimple *uses* BsModal, instead of *being* one.

Fixes #1105

@lolmaus FYI, IIRC you discovered this implementation a while ago...

### DIFF
--- a/addon/components/bs-modal-simple.js
+++ b/addon/components/bs-modal-simple.js
@@ -1,5 +1,5 @@
-import { layout as templateLayout } from '@ember-decorators/component';
-import Modal from 'ember-bootstrap/components/bs-modal';
+import { layout as templateLayout, tagName } from '@ember-decorators/component';
+import Component from '@ember/component';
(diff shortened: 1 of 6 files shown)


simonihmig:
> Did you considered refactoring it to a template-only component? The JavaScript class seems to be only used for default values.

You are right. Though we still need something to put API docs in. And we would need a new custom helper. I agree that's the way forward, but maybe tackle this later in the course of refactoring all component to Glimmer, as similar issues will probably arise then...

jelhan:
Makes sense.

> Though we still need something to put API docs in.

If I recall correctly template only components still need a JavaScript file, which exports the result of `setComponentTemplate()`. So we would have a place to put API docs.

PR #1108, opened 2020-06-12
Update API docs for FormElement#value

simonihmig:
cc @basz

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -257,16 +257,16 @@ export default class FormElement extends FormGroup {
   controlType = 'text';
 
   /**
-   * The value of the control element is bound to this property. You can bind it to some controller property to
-   * get/set the control element's value:
+   * The value of the control element is bound to this property:
    *
    * ```hbs
    * <form.element @controlType="email" @label="Email" @value={{this.email}} />
    * ```
    *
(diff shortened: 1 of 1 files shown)


basz:
lgtm

addon/components/bs-form/element.js
```diff
@@ -257,17 +257,17 @@ export default class FormElement extends FormGroup {
...
-   * get/set the control element's value:
+   * The value of the control element is bound to this property:
    *
    * ```hbs
    * <form.element @controlType="email" @label="Email" @value={{this.email}} />
    * ```
    *
-   * Note: you lose the ability to validate this form element by directly binding to its value. It is recommended
+   * Note two things:
+   * * the binding is uni-directional (DDAU), so you would have to use the `onChange` action to subscribe to changes.
```
jelhan, line 267: Does `{{on "change" this.updateMyValue}}` work as well? I think / hope so but haven't checked myself. Not sure if we should recommend that one instead of overriding `onChange`.

simonihmig: I don't think this can work, as the element that `...attributes` is passed to is `.form-group` IIRC. Which obviously has no `change` event.

But that's what `onChange` is explicitly for. When using `property` it is already wired up to essentially mutate `model`, but when using `@value` you would want to use it.

PR #1110, opened 2020-06-12
Un-deprecate BsButton's buttonType argument

simonihmig:
Due to https://github.com/emberjs/ember.js/issues/18232 we still support the attribute-argument `buttonType` in current 4.0 RCs, so remove the deprecation here to not cause folks to refactor to anything that will actually not work correctly.

### DIFF
--- a/addon/components/base/bs-button.js
+++ b/addon/components/base/bs-button.js
@@ -476,7 +476,6 @@ export default class Button extends Component {
     // deprecate arguments used for attribute bindings only
     runInDebug(() => {
       [
-        ['buttonType:type', 'submit'],
         ['disabled', true],
         ['title', 'foo'],
       ].forEach(([mapping, value]) => {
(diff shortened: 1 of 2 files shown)


jelhan:
Very sad that this bug hasn't been fixed upstream yet. It's very likely to cause confusion. :disappointed:

rwwagner90:
@simonihmig @jelhan so does this mean `buttonType` is required? We were using `type` and relying on splattributes, but that doesn't seem to work anymore.

jelhan:
Yes. The bug is still not fixed in Glimmer VM.

rwwagner90:
I believe `buttonType` is no longer documented. If that is the case, we should add it back to the docs.

jelhan:
> I believe `buttonType` is no longer documented. If that is the case, we should add it back to the docs.

Good catch. Would appreciate a PR if you have the time. Documentation should be available in git history.

PR #1117, opened 2020-06-15
Fix CI config for docs

simonihmig:
The previous `if` condition did not work correctly, resulting in a skipped job. Note this will only run when a release in Github is published, not merely a git tag. But as we use `release-it` this will happen automatically. Added the `edited` event type just to be able to manually trigger the (previously skipped) job run. Hopefully this works now...

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,15 +1,12 @@
 name: Publish docs
 
 on:
-  push:
-    tags:
-      - 'v*'
+  release:
+    types: [published,edited]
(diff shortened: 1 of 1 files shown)


simonihmig:
cc @lolmaus

simonihmig:
Oh well, somehow the `edited` thing does not work (does not trigger CI run), idk why. And don't want to create another release just to trigger CI... 🤔  

It did work previously when running on a `push` event, but that's not how I wanted it to behave...

PR #1182, opened 2020-08-12
improve: replace ember-ref-modifier to ember-ref-bucket

lifeart:
Since ember-ref-modifier has list of downsides (runloop scheduling), api design, I will suggest to replace it to new implementation.

https://github.com/lifeart/ember-ref-bucket

refs: 
https://github.com/lifeart/ember-ref-modifier/issues/345
https://github.com/lifeart/ember-ref-modifier/issues/203
https://github.com/lifeart/ember-ref-modifier/pull/232


simonihmig:
@lifeart thanks for working on this! I haven't looked deeply into `ref-bucket` yet, but wonder if the error could be caused by having *both* `ember-ref-bucket` and `ember-ref-modifier` as dependencies, as that would cause a name clash, right? This addon also uses `ember-popper`, which still uses `ref-modifier` (refactored away in https://github.com/kybishop/ember-popper/pull/127, but this has not landed here yet). 

Shouldn't both addons be able to co-exist? This would mean to rename the `{{ref}}` modifier in the new one I guess!?

lifeart:
>> Shouldn't both addons be able to co-exist? This would mean to rename the {{ref}} modifier in the new one I guess!?
I whould like to have only one `ref-*` addon, because `ref-modifier` is kinda legacy :)
I'm agree on different naming for `ref-bucket` implementation, do you have any naming suggestions?

`ref-as` -> may confuse developers fabiliar with blocks `{{#let foo as ...}}`;

`ref-to` - already used as helper, and I think we could use it as modifier name

`ref-name` - may be an alternative name for modifier.

`create-ref` - may be an case (to follow react naming https://reactjs.org/docs/refs-and-the-dom.html#creating-refs)

basz:
`{{ref this "_element"}}` is used in 6 places and is currently problamatic...

- bs-form (addressed via #1183)
- bs-form/element
- bs-collapse
- bs-carousel
- bs-modal/dialog
- bs-tab/pane

lifeart:
`ember-ref-bucket` now released under `v1.0.0`, modifier name replaced from `ref` to `create-ref`, to keep backward compatibility with `ember-ref-modifier`.
CI looks green

basz:
Cool! Nice work!

basz:
May I kindly ping for a release?

addon/templates/components/bs-form.hbs
```diff
@@ -4,7 +4,7 @@
   novalidate={{this.hasValidator}}
   class={{this.layoutClass}}
   ...attributes
-  {{ref this "_element"}}
+  {{create-ref "mainNode"}}
```
simonihmig, line 7: There is a conflict here now, after #1183 has been merged, that removes ref-usage all together. @lifeart could you please rebase and remove it from here (also from the backing component).

package.json
```diff
@@ -97,7 +97,7 @@
     "ember-load-initializers": "^2.1.1",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-qunit": "^4.6.0",
-    "ember-ref-modifier": "^1.0.1",
+    "ember-ref-bucket": "^1.0.0",
```
simonihmig, line 100: This can be removed, as we already have it in `dependencies`. I realize we had also `ember-ref-modifier` in both places, which did not make really sense...

simonihmig (changes requested):
Thanks for working on this! 👍 

Looks good, just two little things to fix...

simonihmig (approved):
Thank you!

Issue #1181, opened 2020-08-12
BsForm "Cannot read property 'dispatchEvent' of undefined"


basz:
I'm seeing `Cannot read property 'dispatchEvent' of undefined` for forms with submitOnEnter. 

It comes from here;

```
@action
  handleKeyPress(e) {
    let code = e.keyCode || e.which;
    if (code === 13 && this.submitOnEnter) {
      this.triggerSubmit();
    }
  }

  triggerSubmit() {
    let event = document.createEvent('Event');
    event.initEvent('submit', true, true);
    this._element.dispatchEvent(event);
  }
```

this._element is null, hence the error.
_element is set in the form template with the ref helper...

```
  {{ref this "_element"}}
  {{on "keypress" this.handleKeyPress}}
  {{on "submit" this.handleSubmit}}
```

Since the error happening depends on the path the user takes through the application, sometimes _element is correct. It is consistently erroring for those paths.

I've fiddled a bit with using the `fn` helper. Also, perhaps we can simply use `event.target` as that is suppost to be the element... Dunno, curious to see if anyone else has this problem...

---
Output from `ember bootstrap:info`:

```
Npm packages:
ember-bootstrap: ^4.0.0 -> 4.0.0
ember-cli: ~3.20 -> 3.20.0
bootstrap: ^4.4.1 -> 4.5.0
bootstrap-sass: n/a
ember-cli-sass: ^9.0.0 -> 10.0.1
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 4
importBootstrapCSS: false
importBootstrapTheme: false
importBootstrapFont: false
whitelist: undefined
```

Edit: I do not have `ember-ref-bucket` installed as was suggested.

jelhan:
Is #1182 fixing the issue for you?

jelhan:
Another solution may be to inline `triggerSubmit` in `handleKeyPress`. It doesn't seem to be used anywhere else. And we could dispatch the event on `event.target` of the key press event. No need to use `this._element` at all. Would also simplify implementation, wouldn't it?

basz:
Yes, I believe so...

basz:
> Is #1182 fixing the issue for you?

It won't run here... possibly https://github.com/lifeart/ember-ref-bucket/issues/1

basz:
[...]

PR #1183, opened 2020-08-13
Avoid using ref modifier in form

basz:
addresses #1181 by avoiding the ref helper


simonihmig (approved):

simonihmig:
Thanks @basz! 

> This test is failing:

Yes, also for all other (dependabot) PRs, so unrelated. Needs to be investigated...

Issue #1192, opened 2020-08-21
Documented way to use custom form controls broken in 4.0.0-rc.2 & later


kitsuneyo:
The documented way to use custom form controls, such as [Ember Pikaday](https://github.com/adopted-ember-addons/ember-pikaday), is broken in releases `4.0.0-rc.2` and later. It raises the error `Assertion Failed: You cannot set both property and value on a form element` when the control value is changed.

Maybe this is related to PR #1097 "Remove leftover two-way binding of form element's `@value` argument".

Here is the [example from the documentation](https://www.ember-bootstrap.com/api/classes/Components.FormElement.html):

```
<BsForm @model={{this}} @onSubmit={{action "submit"}} as |form|>
  <form.element @label="Select-2" @property="gender" @useIcons={{false}} as |el|>
    <PikadayInput @value={{el.value}} @onSelection={{action (mut el.value)}} id={{el.id}} />
  </form.element>
</BsForm>
```

Here is my implementation:

```
<form.element
  @label={{t "forms.profile.birthday.label"}}
  @property="birthday"
  as |el|
>
  <PikadayInput
    @value={{el.value}}
    @class={{el.validation}}
    @onSelection={{action (mut el.value)}}
    id={{el.id}}
  />
</form.element>
```

Here are the package versions I'm using. I've also tried `4.0.0`. Everything prior to `rc.2` works.

```
DEBUG: -------------------------------
DEBUG: Ember             : 3.18.1
DEBUG: Ember Data        : 3.18.0
DEBUG: Ember Bootstrap   : 4.0.0-rc.2
DEBUG: Ember Simple Auth : 3.1.0-beta.0
DEBUG: -------------------------------
```

Full error output (in testing):

```
Source: | Error: Assertion Failed: You cannot set both property and value on a form element
     at Object.assert (http://localhost:4202/assets/vendor.js:41817:15)
     at BsFormElementWithChangesetValidationsSupport.set value [as value (http://localhost:4202/assets/vendor.js:102326:57)
     at setWithMandatorySetter (http://localhost:4202/assets/vendor.js:37040:22)
     at set (http://localhost:4202/assets/vendor.js:20653:43)
     at PropertyReference.[UPDATE_REFERENCED_VALUE] (http://localhost:4202/assets/vendor.js:52967:16)
     at MutReference.__INVOKE__ember15980008601611513244063377__ (http://localhost:4202/assets/vendor.js:11850:60)
[...]

PR #1202, opened 2020-08-29
value yielded by FormElement should provide two-way binding

jelhan:
This restores the v3 and documented behavior to update a value yielded by `<form.element>` in a custom control. This would fix the regression reported in #1192. 

It does so by providing a two-way binding of the yielded value. This is a little bit contradictory to the intend of #1097, which removed two-way binding but only discussed it on `FormElement` itself - not on the yielded `value` property.

An alternative solution would be yielded an additional `setValue` this would feel more inline with DDAU. The usage would look like this:

```hbs
<BsForm @model={{this.model}} as |form|>
    <input
      type="text"
      value={{el.value}}
      {{on "change" (action this.setValue value="target.value")}}
    />
  </form.element>
</BsForm>
```
The control component has an `onChange` argument, which provide the same value. But I'm not convinced by that name. I think a `value` property and a `setValue` function to update it, would be a more intuitive API.

To be honest I'm tending toward that API but only noticed after I had the fix nearly finished. So I wanted to provide that one as well to start the discussion.

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -286,9 +286,14 @@ export default class FormElement extends FormGroup {
   }
 
   set value(value) {
-    assert('You cannot set both property and value on a form element', isBlank(this.property));
-
     this._value = value;
+  }
+
+  get twoWayBoundValue() {
+    return this.value;
+  }
+  set twoWayBoundValue(value) {
+    this.doChange(value);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


simonihmig:
Thanks @jelhan for working on this!

> To be honest I'm tending toward that API

I agree.

For the record, the usage would be more like this, wouldn't it? (i.e. `setValue` comes form the yielded value)

```
{{on "change" (action el.setValue value="target.value")}}
```

Btw, there is nothing with which we can replace that `target.value` hack in an Octane world, i.e. something to get rid of `action`?

jelhan:
> > To be honest I'm tending toward that API
> 
> I agree.

I have implemented it in #1203. 

> For the record, the usage would be more like this, wouldn't it? (i.e. `setValue` comes form the yielded value)
> 
> ```
> {{on "change" (action el.setValue value="target.value")}}
> ```

Yes. Having `this.setValue` instead of `el.setValue` was a typo.

> Btw, there is nothing with which we can replace that `target.value` hack in an Octane world, i.e. something to get rid of `action`?

I don't think so. I haven't seen a good pattern yet to replace that feature of `{{action}}` helper.

simonihmig:
Closing in favor of #1203

PR #1203, opened 2020-08-30
yield setValue function from <form.element> to mutate the value in custom controls

### LINKED ISSUE #1192: Documented way to use custom form controls broken in 4.0.0-rc.2 & later

jelhan:
In #1097 the two-way binding of `value` yielded from `<form.element>` was removed. We missed to provide an alternative for custom controls to change the value. This regression was reported in #1192.

This PR adds the missing capability by yielding a `setValue` function from `<form.element>`:

```hbs
<BsForm @model={{hash}} as |form|>
    <input value={{el.value}} {{on "input" (action el.setValue value="target.value")}} />
  </form.element>
</BsForm>
```

We do not consider this a breaking change cause two-way binding of yielded `value` was removed in `v4.0.0-rc2`. It was never working in `v4.0.0`.

We evaluated reintroducing two-way binding of yielded `value` in #1202. But choose to prefer a separated `setValue` function. Providing a separate function feels more inline with the rest of our API and the data down, actions up pattern in general.

Closes #1192 and #1202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@
 
 #### :boom: Breaking Change
 * [#1104](https://github.com/kaliber5/ember-bootstrap/pull/1104) Remove support for subclassing component classes  ([@simonihmig](https://github.com/simonihmig))
-* [#1097](https://github.com/kaliber5/ember-bootstrap/pull/1097) Remove leftover two-way binding of form element's `@value` argument ([@basz](https://github.com/basz))
+* [#1097](https://github.com/kaliber5/ember-bootstrap/pull/1097) Remove leftover two-way binding of form element's `@value` argument and yielded `value` ([@basz](https://github.com/basz))
 * [#1106](https://github.com/kaliber5/ember-bootstrap/pull/1106) Refactor BsModalSimple to not extend from BsModal ([@simonihmig](https://github.com/simonihmig))
 
 #### Features
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -74,7 +74,7 @@ const nonDefaultLayouts = A(['checkbox']);
   ```hbs
   <BsForm @model={{this}} @onSubmit={{action "submit"}} as |form|>
     <form.element @label="Select-2" @property="gender" @useIcons={{false}} as |el|>
-      <PikadayInput @value={{el.value}} @onSelection={{action (mut el.value)}} id={{el.id}} />
+      <PikadayInput @value={{el.value}} @onSelection={{action el.setValue}} id={{el.id}} />
     </form.element>
   </BsForm>
   ```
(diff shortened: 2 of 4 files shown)


jelhan:
@simonihmig I wasn't sure if we should consider this as an _enhancement_ or a _bug_ fix. It's adding a new API. So it could be seen as an enhancement. But at the same time it fixes a regression in v4. :man_shrugging: Feel free to change the label. I changed it twice from _enhancement_ to _bug_ and the other way around when creating this merge request. :innocent:

simonihmig (approved):

PR #1205, opened 2020-08-30
yield submitButton component from <BsForm>

jelhan:
Adds a `submitButton` component yielded by `<BsForm>`. It's a `<BsButton>` with having `@buttonType="submit"` and `@type="primary"`. It's disabled if the form is submitting.

This new API helps to reduce the code needed to implement a common submit button. It also improves readability in my opinion as it directly states that the button belongs to the form and will submit it.

```hbs
{{! new API added by this PR }}
<BsForm as |form|>
  <form.submitButton>Submit</form>
</BsForm>

{{! the same with current API }}
<BsForm as |form|>
  <BsButton @buttonType="submit" @type="primary" @disabled={{form.isSubmitting}}>Submit</BsButton>
</BsForm>

{{! minimal code needed for a submit button with current API }}
<BsForm as |form|>
  <BsButton @buttonType="submit">Submit</BsButton>
</BsForm>
```

Additionally the submission state of the form is bound to the button state. This is helpful for showing submission state on the submit button.

Using it in isolation would look like this:

```hbs
<BsForm as |form|>
  <form.submitButton as |button|>
    {{#if button.isPending}}
      <FaIcon @icon="spinner" />
    {{/if}}
    {{#if button.isFulfilled}}
      <FaIcon @icon="check" />
    {{/if}}
    {{#if button.isRejected}}
      <FaIcon @icon="times" />
    {{/if}}
  </form.submitButton>
</BsForm>
```

This doesn't provide much value over using `{{form.isSubmitting}}`, `{{form.isSubmitted}}` and `{{form.isRejected}}`. But it gets powerful if a user customizes the `<BsButton>` component to show the state the button is currently. In that case the submission state integration comes out of the box if just doing this:

```hbs
<BsForm as |form|>
  <form.submitButton />
</BsForm>
```

Such a customization could be achived by overwriting `<BsButton>` in the application. Similar to this:

```hbs
// app/components/bs-button.hbs

<BsButton::Original as |button|>
  {{#if button.isPending}}
    <FaIcon @icon="spinner" />
  {{/if}}
  {{#if button.isFulfilled}}
    <FaIcon @icon="check" />
  {{/if}}
  {{#if button.isRejected}}
    <FaIcon @icon="times" />
  {{/if}}
</BsButton>
```

```js
// app/components/bs-button/original.js

export { default } from 'ember-bootstrap/components/bs-button';
```

This is one of the two outstanding features from #638.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -196,6 +196,14 @@ export default class Form extends Component {
   groupComponent = 'bs-form/group';
 
   /**
+   * @property submitButtonComponent
+   * @type {String}
+   * @private
+   */
+  @defaultValue
+  submitButtonComponent = 'bs-button';
+
+  /**
    * `isSubmitting` is `true` after `submit` event has been triggered and until Promise returned by `onSubmit` is
    * fulfilled. If `validate` returns a Promise that one is also taken into consideration.
    *
@@ -237,6 +245,29 @@ export default class Form extends Component {
    */
   @defaultValue
   isRejected = false;
+
+  /**
+   * State of the form expressed as state values expected by `<BsButton>`.
+   *
+   * @property submitButtonState
+   * @type {String}
+   * @private
+   */
+  get submitButtonState() {
+    if (this.isSubmitting) {
+      return 'pending';
+    }
+
+    if (this.isSubmitted) {
+      return 'fulfilled';
+    }
+
+    if (this.isRejected) {
+      return 'rejected';
+    }
+
+    return 'default';
+  }
 
   /**
    * Count of pending submissions.
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -32,7 +32,16 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
   ### Submitting the form
 
-  When the form is submitted (e.g. by clicking a submit button), the event will be intercepted and the `onSubmit` action
+  The form yields a `submitButton` component, which is a preconfigured `<BsButton>` with `@type="primary"` and `@buttonType="submit"`.
+  The button is disabled while a form submission is pending. Additionally the button state is bound to the form submission state.
+
+  ```hbs
+  <BsForm as |form|>
+    <form.submitButton>Submit</form.submitButton>
+  </BsForm>
+  ```
+
+  When the form is submitted (e.g. by clicking the submit button), the event will be intercepted and the `onSubmit` action
   will be sent to the controller or parent component.
   In case the form supports validation (see "Form validation" below), the `onBefore` action is called (which allows you to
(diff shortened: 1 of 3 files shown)


jelhan:
We could implement submit for correctly positioning the button in horizontal forms if using Bootstrap 3 as discussed in #92. If we want to we should do it now before shipping the feature in a stable release as it would be a breaking change otherwise. But I'm not sure if it makes sense to still invest time in Bootstrap 3 to be honest.

addon/components/bs-form.js
```diff
@@ -82,29 +91,35 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
...
+  by `onSubmit` is fulfilled. It could be used to show a loading spinner instead of the form while it's submitting for example:
 
   ```hbs
   <BsForm @onSubmit={{action "save"}} as |form|>
-    <BsButton @buttonType="submit" @disabled={{form.isSubmitting}}>
-      Save
-      {{#if form.isSubmitting}} {{fa-icon "spinner"}} {{/if}}
-    </BsButton>
+    {{#if form.isSubmitting}}
+      <FaIcon @icon="spinner" />
```
simonihmig, line 99: Should we use a plain BS spinner example here, like this https://getbootstrap.com/docs/4.5/components/spinners/?

Juts in case people copy and paste this and wonder why it's not working ootb!?

addon/components/bs-form.js
```diff
@@ -195,6 +210,14 @@ export default class Form extends Component {
   @defaultValue
   groupComponent = 'bs-form/group';
 
+  /**
+   * @property submitButtonComponent
+   * @type {String}
+   * @private
+   */
+  @defaultValue
+  submitButtonComponent = 'bs-button';
```
simonihmig, line 219: This reminds me that we should get rid of that pattern sooner than later to make component tree-shakable. It's ok for now, as we have that all over the place, but I just created this issue to not forget this: https://github.com/kaliber5/ember-bootstrap/issues/1206

simonihmig (approved):
Thanks @jelhan, this looks great!

> We could implement submit for correctly positioning the button in horizontal forms if using Bootstrap 3 as discussed in #92.

I would tend to think that it's not really the concern of the button to determine its position. Rather something like a special yielded layout component or so, that *uses* the plain button. But I also think we don't need to add things for BS3 specifically anymore :)

I just wanted to check what breaks CI for beta and canary, so we can get all PRs green...

addon/components/bs-form.js
jelhan: Done in 41ffa5bc93666ae856158c5bd072cca6d40c06e3. Also noticed that it might be good practice to not only show an icon but also have a text - at least for screen readers only (accessibility).

PR #1217, opened 2020-09-01
Fix CSP failures in FastBoot tests

simonihmig:
This was caused by several facts:
* `ember-a11y-testing` adds two styles tags with inline content
* `ember-cli-fastboot-testing` parses the FastBoot rendered HTML string here: https://github.com/embermap/ember-cli-fastboot-testing/blob/47882cebaaf9fb482df3e3b1eae1067807726890/addon-test-support/index.js#L120
* while the inline style violating our CSP was somehow ignored before, it seems the latest Chrome 85 now triggers a CSP violation event on that line above
* which in turn makes the test fail

This adds hash values to our CSP policy to specifically allow these two inline style elements

### DIFF
--- a/tests/dummy/config/content-security-policy.js
+++ b/tests/dummy/config/content-security-policy.js
@@ -11,7 +11,11 @@ module.exports = function () {
         // Bootstrap 4 uses data URL for some SVG images in CSS
         'data:',
       ],
-      'style-src': ["'self'"],
+      'style-src': [
+        "'self'",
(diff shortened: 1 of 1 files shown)


jelhan:
Thanks a lot for fixing this one.

We are planning to add a feature to ember-cli-content-security-policy to support these cases with zero config: https://github.com/rwjblue/ember-cli-content-security-policy/issues/149

PR #1218, opened 2020-09-01
Update docs for new yielded submit button

simonihmig:


jelhan:
Ups. Totally forgot about the demo part of the docs. Thanks a lot for taking care of that one!

PR #1219, opened 2020-09-02
Fix Tooltips/Popovers when used w/ FastBoot and rehydration enabled

simonihmig:
With rehydration the FastBoot rendered empty TextNode (used to find the component's parent element) is reused. So the TextNode instance that the component has created at runtime is not attached to the DOM, thus cannot be used to find the parent. In this case Ember's private API `Ember.ViewUtils.getViewBounds()` is used as a fallback.

Related: https://github.com/emberjs/rfcs/issues/168


lolmaus (approved):
Hmm.

PR #1221, opened 2020-09-03
Colocate component templates

jelhan:
Most work was done by [ember-component-template-colocation-migrator](https://github.com/ember-codemods/ember-component-template-colocation-migrator). Only needed to fix some edge cases, which triggered bugs.

Component template colocation is available since Ember 3.13. As we require Ember 3.16+ and Ember CLI 3.15+ this shouldn't be a breaking change for any consumer.


simonihmig (commented):
Thanks for working on this!

Why has `bs-tooltip/element.hbs` been renamed to `bs-contextual-help/element.hbs`?
This seems odd, as the `bs-contextual-help/element` component is basically an abstract class, extended from both by `bs-tooltip/element` and `bs-popover/element`.

jelhan:
> Why has `bs-tooltip/element.hbs` been renamed to `bs-contextual-help/element.hbs`?
> This seems odd, as the `bs-contextual-help/element` component is basically an abstract class, extended from both by `bs-tooltip/element` and `bs-popover/element`.

Good catch. I got confused by the template import in `bs-contextual-help/element.js` and wrongly assumed that it needs the template. Moved the template back to `bs-tooltip/element`.

simonihmig (approved):

PR #1222, opened 2020-09-04
mention removed default attribute bindings in changelog

### LINKED ISSUE #1091: Unable to set custom element-id on form elements
In v4 setting `@id="my-id"` on a form element does not work as expected.

in v3 you could do

```
<form.element @id="password"
                @controlType="password"
                @property="password"
                @label={{t "model.identity.password"}}  as |element|>
    <element.control autocomplete="current-password"/>
  </form.element>
```
In v4 you need to add `id="..." ` to the element.control
```
<form.element @_elementId="password"
                @controlType="password"
                @property="password"
                @label={{t "model.identity.password"}} as |element|>
    <element.control id={{element.id}} autocomplete="current-password"/>
  </form.element>
```

(the element will get id='password-field', label will also get for='password-field')

However `@id="my-id"` does not work anymore. `@elementId="my-id"` will throw an error (cause tagLess component), so, for now, we will need to do `@_elementId="my-id"`.

jelhan:
Closes #1091

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,7 +95,7 @@
 ## v4.0.0-rc.0 (2020-04-28)
 
 #### :boom: Breaking Change
-* All components are refactored to tag-less Ember components as a first stepping stone to a full Glimmer components rewrite. In most cases this should not affect you, but there are some edge cases where it might break, e.g. using `ember-test-selectors` with the legacy curly invocation syntax.
+* All components are refactored to tag-less Ember components as a first stepping stone to a full Glimmer components rewrite. In most cases this should not affect you, but there are some edge cases where it might break, e.g. using `ember-test-selectors` with the legacy curly invocation syntax or using `attributeBindings` provided by `@ember/component` like `@id`. In all cases setting them as HTML attributes using angle bracket component invocation is the recommended path forward..
 * [#965](https://github.com/kaliber5/ember-bootstrap/pull/965), [#1037](https://github.com/kaliber5/ember-bootstrap/pull/1037) Drop support for Ember < 3.16 and node 8 ([@jelhan](https://github.com/jelhan), [@simonihmig](https://github.com/simonihmig))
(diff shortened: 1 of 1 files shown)


jelhan:
@simonihmig Do you think we should update the GitHub release messages as well?

CHANGELOG.md
```diff
@@ -95,7 +95,7 @@
 ## v4.0.0-rc.0 (2020-04-28)
 
 #### :boom: Breaking Change
-* All components are refactored to tag-less Ember components as a first stepping stone to a full Glimmer components rewrite. In most cases this should not affect you, but there are some edge cases where it might break, e.g. using `ember-test-selectors` with the legacy curly invocation syntax.
+* All components are refactored to tag-less Ember components as a first stepping stone to a full Glimmer components rewrite. In most cases this should not affect you, but there are some edge cases where it might break, e.g. using `ember-test-selectors` with the legacy curly invocation syntax or using `attributeBindings` provided by `@ember/component` like `@id`. In all cases setting them as HTML attributes using angle bracket component invocation is the recommended path forward..
```
simonihmig, line 98:
```suggestion
* All components are refactored to tag-less Ember components as a first stepping stone to a full Glimmer components rewrite. In most cases this should not affect you, but there are some edge cases where it might break, e.g. using `ember-test-selectors` with the legacy curly invocation syntax or using `attributeBindings` provided by `@ember/component` like `@id`. In all cases setting them as HTML attributes using angle bracket component invocation is the recommended path forward.
```

double period :wink:

simonihmig (approved):

simonihmig:
> Do you think we should update the GitHub release messages as well?

Yeah, why not. 

Thanks!

CHANGELOG.md
jelhan: I should have known that writing pull requests from your smartphone is maybe not the best idea. 🤪

PR #1223, opened 2020-09-05
Manual include/exclude of components: <BsForm> requires <BsButton> due to yielded submitButton

jelhan:
Missed this in #1205


jelhan:
I haven't had the time yet to verify that it actually works. I don't think we have test coverage for this feature, have we? Please review carefully.

simonihmig:
Looks good, thank you!

> I don't think we have test coverage for this feature, have we? 

Nope :(

simonihmig:
Released in v4.1.2

jelhan:
Thanks a lot!

PR #1226, opened 2020-09-07
Refactor BsAccordion to Glimmer component

simonihmig:


addon/components/bs-accordion/item/_body.js
```diff
@@ -0,0 +1,16 @@
+/**
```
jelhan, line 1: Why is the file prefixed with an underscore? Is it breaking if we have a `.js` file without a default export for a template-only component? Is this affecting all template-only components or is there something special about `<BsAccordion::Item::Body>`?

addon/components/bs-accordion/item/title.js
```diff
@@ -32,13 +30,12 @@ export default class AccordionItemTitle extends Component {
...
 
   @action
   handleClick(e) {
     e.preventDefault();
-    if (!this.disabled) {
-      this.onClick();
+    if (!this.args.disabled) {
+      this.args.onClick?.();
     }
   }
```
jelhan, line 40: Wondering how hard we should try to push for template-only components. Template-only components should provide a better performance. It should be possible in this case if introducing a `bs-noop` template helper and using `{{action}}` helper. It would look like this:

```hbs
{{on "click" (if @disabled (bs-noop) (action @onClick preventDefault=true))}}
```

jelhan (approved):
Looks good to me. I guess you are already looking at the failed tests for Ember 3.16?

addon/components/bs-accordion/item/_body.js
simonihmig: > Is it breaking if we have a .js file without a default export for a template-only component? 

Exactly. 

> Is this affecting all template-only components

Yes.

addon/components/bs-accordion/item/title.js
simonihmig: > Template-only components should provide a better performance

I don't think that will have any meaningful effect here, unless you are rendering hundreds or thousands of accordion items!? :) Also you would have to account for the additional cycles that the template helpers need, which also occur even when the item is never clicked.

addon/components/bs-accordion/item/_body.js
jelhan: Thanks for explaining! Seems to be another reason to revisit the tools we use to generate our API docs.

addon/components/bs-accordion/item/title.js
jelhan: Agree.

We may want to revisit as soon as [Contextual Helper and Modifier]() are available in all supported versions. If I got https://github.com/emberjs/rfcs/issues/660#issuecomment-689013203 right, it will make it both more readable and performant:

```hbs
{{unless @disabled (on "click" (action @onClick preventDefault=true))}}
```

simonihmig:
Was unable to fix the 3.16 issue, which is related to the use of `@localCopy`. Reported in https://github.com/pzuraq/tracked-toolbox/issues/17

simonihmig:
@jelhan I wanted to get this in before working in BS5 support for accordions, and somehow managed to find an ugly workaround for this `@localCopy` issue with Ember 3.16. All passing now! :tada:

addon/components/bs-accordion.js
```diff
@@ -52,23 +49,44 @@ export default class Accordion extends Component {
...
+    // ideally we would have used @localCopy here, but unfortunately this fails for Ember 3.16 in this special case
+    // see https://github.com/pzuraq/tracked-toolbox/issues/17
+    // So don't look at this, it is going to get ugly...
+    this._isSelected; // just consume this to invalidate the getter when this changes
+
+    if (this.args.selected && this._prevSelected !== this.args.selected) {
+      // eslint-disable-next-line ember/no-side-effects
+      this._isSelectedNonTracked = this._prevSelected = this.args.selected;
+
+      return this._isSelectedNonTracked;
```
jelhan, line 81: This return statement could be removed to simplify the code a little bit.

```suggestion
```

jelhan (approved):
Implementing `@localCopy` yourself. To be honest it was very challenging to follow that code. I assume it must have been way harder to write it. :clap:

simonihmig:
> To be honest it was very challenging to follow that code. I assume it must have been way harder to write it.

Yeah. I had a version that was more straightforward, with just one private tracked property. But that suffered from the same problem, same assertion only in Ember 3.16. From then on it was mainly trial and error. I just had the vague intuition to have two synced internal props, one of which is non-tracked, which I could read from without triggering that same computation thingy. Somehow it eventually worked out, I was the one most surprised by that! 😅

PR #1248, opened 2020-09-17
Introduce useDefaultValueIfUndefined optional feature to ease passing arguments through

jelhan:
We are wrapping components provided by Ember Bootstrap in custom addons. So we have many code like this:

```hbs
// addon/components/my-form.hbs

<BsForm
  @formLayout={{@formLayout}}
  @model={{@model}}
  @onSubmit={{@onSubmit}}
>
  {{yield form}}
</BsForm>
```

We noticed that the current implementation of `defaultValue` decorator does not use the defaults if the argument is passed but being `undefined`. So we would need to redeclare all default value:

```hbs
// addon/components/my-form.hbs

<BsForm
  @formLayout={{if @formLayout @formLayout "vertical"}}
  @model={{if @model @model (hash)}}
  @onSubmit={{if @onSubmit @onSubmit this.noop}}
>
  {{yield form}}
</BsForm>
```

Not great solution. So we thought we might want to change the `defaultValue` decorator.

I also noticed that `arg` decorator used for the components, which have been refactored to `@glimmer/component` seem to fall back to default value if `undefined` is passed: https://github.com/kaliber5/ember-bootstrap/blob/c740df9f818a8081e523e265a8e438418b78648c/addon/utils/decorators/arg.js#L5-L8 So I think we can consider the current implementation of `defaultValue` to be a bug.

I wasn't sure how to test this as it affects so many places. Just added one test as a showcase.

Disclaimer: Decorators are still scary to me. I'm definitely not an expert on that topic. Came up with that fix by try and error rather than by understanding what I'm doing. Please let me know if it's totally wrong.

### DIFF
--- a/tests/helpers/bootstrap.js
+++ b/tests/helpers/bootstrap.js
@@ -1,5 +1,6 @@
 import config from 'dummy/config/environment';
-import { skip, test } from 'qunit';
+import { getOwnConfig } from '@embroider/macros';
+import { module, skip, test } from 'qunit';
 import { Promise } from 'rsvp';
 
 const currentBootstrapVersion = parseInt(config.bootstrapVersion);
@@ -26,6 +27,23 @@ export function versionDependent(v3, v4) {
   }
 
   return v4;
+}
+
+// eslint-disable-next-line ember/no-test-module-for
+export function moduleForOptionalFeature(optionalFeature, fn) {
+  let ownConfig = getOwnConfig();
+
+  if (!Object.keys(ownConfig).includes(optionalFeature)) {
+    throw new Error(`Optional feature ${optionalFeature} does not exist.`);
+  }
+
+  let optionalFeatureEnabled = ownConfig[optionalFeature];
+  let name = `Optional feature: ${optionalFeature}`;
+  if (optionalFeatureEnabled) {
+    module(name, fn);
+  } else {
+    skip(name, fn);
+  }
 }
 
 export function visibilityClass() {
--- a/tests/helpers/bootstrap.js
+++ b/tests/helpers/bootstrap.js
@@ -1,5 +1,5 @@
 import config from 'dummy/config/environment';
-import { getOwnConfig } from '@embroider/macros';
+import { dependencySatisfies, getConfig, getOwnConfig, macroCondition } from '@embroider/macros';
 import { module, skip, test } from 'qunit';
 import { Promise } from 'rsvp';
 
@@ -31,13 +31,27 @@ export function versionDependent(v3, v4) {
 
 // eslint-disable-next-line ember/no-test-module-for
 export function moduleForOptionalFeature(optionalFeature, fn) {
-  let ownConfig = getOwnConfig();
+  let optionalFeatureEnabled;
 
-  if (!Object.keys(ownConfig).includes(optionalFeature)) {
+  // `getConfig('ember-bootstrap')` is the correct way to retrieve the
+  // configuration of the addon it it's dummy app. But the `getConfig`
+  // macro is broken for dummy app of an addon in classic builds.
+  // `getOwnConfig` macro also has a bug in classic builds. It resolves
+  // with the configuration of the addon not of it's dummy app. We can
+  // use that bug as a work-a-round for classic builds until the but of
+  // `getConfig` macro has been fixed.
+  // Both bugs are tracked in this issue:
+  // https://github.com/embroider-build/embroider/issues/537
+  if (macroCondition(dependencySatisfies('@embroider/webpack', '*'))) {
+    optionalFeatureEnabled = getConfig('ember-bootstrap')[optionalFeature];
+  } else {
+    optionalFeatureEnabled = getOwnConfig()[optionalFeature];
+  }
+
+  if (optionalFeatureEnabled === undefined) {
     throw new Error(`Optional feature ${optionalFeature} does not exist.`);
   }
 
-  let optionalFeatureEnabled = ownConfig[optionalFeature];
   let name = `Optional feature: ${optionalFeature}`;
   if (optionalFeatureEnabled) {
     module(name, fn);
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -161,6 +161,8 @@ export default class Form extends Component {
    * @type Ember.Object
    * @public
    */
+  @defaultValue
+  model = macroCondition(getOwnConfig().useDefaultValueIfUndefined) ? {} : undefined;
 
   /**
    * Set the layout of the form to either "vertical", "horizontal" or "inline". See http://getbootstrap.com/css/#forms-inline and http://getbootstrap.com/css/#forms-horizontal
--- a/addon/utils/decorators/arg.js
+++ b/addon/utils/decorators/arg.js
@@ -1,10 +1,16 @@
+import { macroCondition, getOwnConfig } from '@embroider/macros';
+
 export default function arg(target, key, descriptor) {
   let defaultValue = descriptor.initializer ? descriptor.initializer() : undefined;
 
   return {
     get() {
-      const argValue = this.args[key];
-      return argValue !== undefined ? argValue : defaultValue;
+      if (macroCondition(getOwnConfig().useDefaultValueIfUndefined)) {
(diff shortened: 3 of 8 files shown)


jelhan:
This does not resolve the issue for `onBefore`, `onSubmit` and `onInvalid` as they do not use the `defaultValue` decorator. Need to address them differently.

simonihmig:
> So we would need to redeclare all default value:

We needed to do similar things, so yes, I see you!

> So I think we can consider the current implementation of defaultValue to be a bug.

I don't think so actually. At least depending from which angle you see it...

It was introduced to not cause breaking changes when refactoring to native classes, where we define default values als class fields. But those always live on the instance, and are not part of the prototype chain, as was the case for properties in Ember's object model (`.extend()`). So `@defaultValue` makes these live on the prototype again (as CPs), as they were before. Maybe it could have been named differently as the primary intent again was to keep the behavior the same, so you could (jokingly) call the decorator `@makeComputedOnPrototypeWithAdditionalDefaultValueSupport`.

The intent was *not* to change the treatment of `undefined` when passed as an argument. That said, I am pretty sure that any 2.x version (i.e. without native classes and no `@defaultValue`) shows the exact same behavior you are "complaining" about here! See this quick demo: https://ember-twiddle.com/3dca6060033641818c24965a7468bd57

So, as `@defaultValue` is an implementation detail, and not used by any users, the implementation seems to do exactly what it is meant to do (preserve compatibility)!

That doesn't mean that it wouldn't be desirable to change that. However that seems to be a potentially breaking change. Though not sure if anybody would really intentionally pass `@someArg={{undefined}}` to "unset" some property - which would not work with the new semantics anymore.

Having said that...

> I also noticed that arg decorator used for the components, which have been refactored to @glimmer/component seem to fall back to default value if undefined is passed

This discussions seems to have uncovered that the replacement of `@defaultValue` (for Ember.Components) with `@arg` (for Glimmer.Components) could be seen - again in a very edge case scenario - as a breaking change! Not sure if this means we have to rethink this, e.g. change the `@arg` behavior to match what we had before, and only do what we do now as a later explicitly breaking change?

FWIW, I just had a very similar topic in a PR to embroider yesterday, when I realized that `Object.assign()` treats a missing property differently than one with an undefined value:

```
Object.assign({}, { foo: 'bar' }, { baz: true }); // {foo: "bar", baz: true}
Object.assign({}, { foo: 'bar' }, { foo: undefined, baz: true }); // {foo: undefined, baz: true}
```

That's pretty much the same...

So TL;DR, I am in favor of changing this, but I am not sure if we should do that ignoring any potential (but assumingly very rare) breaking changes, or be totally correct Semver-wise and only introduce this change in a major bump?

jelhan:
> So TL;DR, I am in favor of changing this, but I am not sure if we should do that ignoring any potential (but assumingly very rare) breaking changes, or be totally correct Semver-wise and only introduce this change in a major bump?

I see two potential path forwards:

1. Explore for which arguments `undefined` is a supported value. There are three cases I currently see:
    1. In some cases (e.g. `formLayout` of `<BsForm>`) Ember Bootstrap currently throws if `undefined` is passed. Changing these should be safe.
    2. In some other cases it may not throw but passing `undefined` is not documented as a supported value and the component does not behave as expected if doing so. Not sure how we should handle these case. Strictly speaking it may not be considered as a breaking change if `undefined` is not documented as a supported value.
    3. And of course there might be cases in which `undefined` is a supported value but not the default one. If we have these cases we would need to drop `undefined` as a supported value and introduce `null` as a replacement (if it's not yet supported). These will be a breaking change for sure.
2. Introduce an optional feature to change the behavior from compatibility mode to apply default if `undefined` is passed. This will allow us to move forward without waiting for a new major. `@embroider/macros` should provide a great infrastructure for that one. It should only affect `@defaultValue` and `@arg` decorators. So the increased complexity should be fine.

The optional feature also has the benefit that we can provide a path forward for `@arg` decorator. As we already released the breaking change in v4.1 rolling it back would resolve the breaking change if compared to v4.0 but introduce a breaking change if compared to v4.1. Supporting both through an optional feature seems to be the only path forward here in my opinion.

simonihmig:
> As we already released the breaking change in v4.1

We didn't release anything related to this at all, did we?
The Glimmer component stuff is still unreleased!

But I like the idea of that optional feature anyway! Indeed exposing this flag as our [`ownConfig`](https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L98-L107) enables to easily support different builds using  embroiders babel macros!

jelhan:
> > As we already released the breaking change in v4.1
> 
> We didn't release anything related to this at all, did we?
> The Glimmer component stuff is still unreleased!

Oh. You are right. I thought it was released as part of v4.1. Good that we catched this one before releasing!

> But I like the idea of that optional feature anyway! Indeed exposing this flag as our [`ownConfig`](https://github.com/kaliber5/ember-bootstrap/blob/master/index.js#L98-L107) enables to easily support different builds using embroiders babel macros!

Will implement in this pull request.

jelhan:
@simonihmig I implemented a `useDefaultValueIfUndefined` optional feature. Changed `@defaultValue` and `@arg` decorators to use it. I also added an additional ember try scenario to test it. 

Will writing some tests just to have one I directly run into another case, which wasn't covered by changing one of the decorators: `model` argument of `<BsForm>`. So #1250 wasn't a rare case and I expect to face more issues like that. But I'm a little bit lazy to write tests for all of our public API. Would it be fine for you if we merge this one without additional tests and than fix bug by bug?

tests/helpers/bootstrap.js
```diff
@@ -28,6 +29,23 @@ export function versionDependent(v3, v4) {
   return v4;
 }
 
+// eslint-disable-next-line ember/no-test-module-for
+export function moduleForOptionalFeature(optionalFeature, fn) {
+  let ownConfig = getOwnConfig();
```
jelhan, line 34: Something goes wrong here in an Embroider build.

This is the build output:

```js
  let ownConfig = Object(_home_jelhan_Code_ember_ecosystem_ember_bootstrap_node_modules_embroider_macros_src_addon_runtime__WEBPACK_IMPORTED_MODULE_2__["config"])("/tmp/embroider/6e5bc6/tests/dummy");
```

It's working fine if changing it manually to this:

```js
  let ownConfig = Object(_home_main_Code_ember_ecosystem_ember_bootstrap_node_modules_embroider_macros_src_addon_runtime__WEBPACK_IMPORTED_MODULE_2__["config"])("/tmp/embroider/6e5bc6")
```

That variable is called `packageRoot`. So it seems as if `@embroider/macros` uses a wrong package root for dummy app in an embroider build.

Will report upstream.

simonihmig (approved):
LGTM!

simonihmig:
> Would it be fine for you if we merge this one without additional tests and than fix bug by bug?

Yes.

> I would recommend that we consider every difference between not passing an argument or passing it with undefined value as a bug if useDefaultValueIfUndefined optional feature is enabled.

Makes sense.

> Not sure if we should document that optional feature just right now. Maybe we can treat it as unstable API for right now to gather some more experience without blocking a new minor release?

Agree.

Hope the embroider issue get's sorted out. Already spent most of last Friday on another one. I *may* be able to look at tis new one *some time*, but no guarantees...

Thanks for working on this!

tests/helpers/bootstrap.js
simonihmig: @jelhan you already have the changes applied you mentioned in the embroider issue (`getConfig('ember-bootstrap')`), right?  Could you please push this, so we have this most current version available here (that fails for a classic build instead of embroider if I correctly understood), as a reproduction in case I find the time to look at this!

jelhan: I pushed a working solution for both classic and embroider builds in f23191f3cfba12d43196a12090eee75be207c025. It uses the correct code for embroider builds and uses the wrong behavior of `getOwnConfig` in classic builds. Would you be okay with using that work-a-round for now to unblock that one? We can remove it by deleting a few lines of code as soon as the bug is fixed upstream. It also only affects our own tests. So no risk for consumers.

simonihmig: :+1:

PR #1250, opened 2020-09-17
<BsForm> should support @onBefore, @onSubmit and @onInvalid arguments to be undefined

jelhan:
Before it was throwing if value of `@onBefore`, `@onSubmit` or `@onInvalid` is `undefined`. With this changes it supports all of them being `undefined`.

This makes usage of `<BsForm>` in addons, which forward the arguments difficult. A wrapping addon needed to pass a noop function itself if the consumer hasn't passed that argument. The code looked like this:

```hbs
<BsForm
  @onBefore={{if @onBefore @onBefore this.noop}}
  @onSubmit={{if @onSubmit @onSubmit this.noop}}
  @onInvalid={{if @onInvalid @onInvalid this.noop}}
  as |form|
>
  {{yield form}}
</BsForm>
```

After this change a wrapping addon can do this:

```hbs
<BsForm
  @onBefore={{@onBefore}}
  @onSubmit={{@onSubmit}}
  @onInvalid={{@onInvalid}}
  as |form|
>
  {{yield form}}
</BsForm>
```

This is the companion to #1248.

While fixing that issue I noticed that one of the tests was not testing what it claimed to do. I fixed that test on the go with the minimum changes needed. The test needs a refactoring for sure. But didn't wanted to touch it too much in a bug fix PR.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -381,7 +381,6 @@ export default class Form extends Component {
    * @param { Object } model  The form's `model`
    * @public
    */
-  onBefore(model) {} // eslint-disable-line no-unused-vars
 
   /**
    * Action is called when submit has been triggered and the model has passed all validations (if present).
@@ -391,7 +390,6 @@ export default class Form extends Component {
    * @param { Object } result The returned result from the validate method, if validation is available
    * @public
    */
-  onSubmit(model, result) {} // eslint-disable-line no-unused-vars
 
   /**
    * Action is called when validation of the model has failed.
@@ -401,7 +399,6 @@ export default class Form extends Component {
    * @param { Object } error
    * @public
    */
-  onInvalid(model, error) {} // eslint-disable-line no-unused-vars
 
   /**
    * Submit handler that will send the default action ("action") to the controller when submitting the form.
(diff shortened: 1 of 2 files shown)


tests/integration/components/bs-form-test.js
```diff
@@ -499,8 +499,8 @@ module('Integration | Component | bs-form', function (hooks) {
 
   test('yielded submit action rejects for expected scenarios', async function (assert) {
     let scenarios = [
-      { validate: sinon.fake.rejects('rejected value') },
-      { onSubmit: sinon.fake.rejects('rejected value') },
+      { validate: sinon.fake.rejects('rejected value'), onSubmit: sinon.fake.resolves() },
```
jelhan, line 502: `validate` function was not executed at all before cause `hasValidator` was false. Ember Bootstrap was rejecting cause `onSubmit` was undefined. Explicitly passing both seems to be more clear. Especially as a property from one scenario leaks to the next scenario if not being overwritten.

addon/components/bs-form.js
```diff
@@ -426,7 +423,10 @@ export default class Form extends Component {
     let model = this.model;
 
     this.incrementProperty('pendingSubmissions');
-    this.onBefore(model);
+
+    if (typeof this.onBefore === 'function') {
+      this.onBefore(model);
+    }
```
jelhan, line 429: I think this is also the path forward for glimmer component refactoring. After refactoring to glimmer components it would look like this:

```js
    if (typeof this.args.onBefore === 'function') {
      this.args.onBefore(model);
    }
```

simonihmig (approved):
Regardless of the outcome of #1248, these changes make sense, as overriding an action with `undefined` would never constituted a valid value, as this would break badly anytime trying to call the action (in the old style without that guard).

So thanks! :+1:

PR #1125, opened 2020-06-16
WIP: Test embroider build with optimized config

simonihmig:
Trying to follow the recommendations at https://github.com/embroider-build/embroider/blob/master/ADDON-AUTHOR-GUIDE.md#support-level-optimized-embroider-safe

PR #1251, opened 2020-09-18
Support Embroider with staticComponents for full tree-shaking

### LINKED ISSUE #1206: Support staticComponents with Embroider
Currently we use string properties defined on a component class and the `{{component}}` helper to refer to other used components like here: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form/element.js#L816, or even "worse" this: https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form/element.js#L797-L808.

This is done so we can override components from the caller side, so the actually used components are not hardcoded in the template. This pattern however makes it basically impossible to statically analyze component usage, therefore the more aggressive `staticComponents` of Embroider to tree-shake unused components would certainly not work currently.

We should support that in the future. Which AFAIU means getting rid of component strings defined on the class. Importing a component class and passing that to the component helper *should* work.

Related:
* https://github.com/emberjs/ember.js/issues/19099
* https://github.com/kaliber5/ember-bootstrap/pull/1125

simonihmig:
Replaces #1125, fixes #1206 

~Forms haven't been updated yet, other than that all components pass locally with strict embroider settings.~ Only locally, because I needed to patch some thing, which still have to be fixed upstream or they are already but nor released yet:

* https://github.com/embroider-build/embroider/issues/539
* ~https://github.com/embroider-build/embroider/pull/536~
* ~https://github.com/embroider-build/embroider/pull/521~

@jelhan As you can see, this adds quite a bit of boilerplate, but this seems the only way for now to make it work. See https://github.com/embroider-build/embroider/issues/487

### DIFF
--- a/addon/components/bs-navbar.hbs
+++ b/addon/components/bs-navbar.hbs
@@ -1,15 +1,16 @@
 {{#let
   (hash
-    toggle=(component this.toggleComponent onClick=this.toggleNavbar collapsed=this._collapsed)
-    content=(component this.contentComponent collapsed=this._collapsed onHidden=this.onCollapsed onShown=this.onExpanded)
+    toggle=(component (bs-default @toggleComponent (component "bs-navbar/toggle")) onClick=this.toggleNavbar collapsed=this._collapsed)
+    content=(component (bs-default @contentComponent (component "bs-navbar/content")) collapsed=this._collapsed onHidden=this.onCollapsed onShown=this.onExpanded)
     nav=(component
-      this.navComponent
-      linkToComponent=(component this.linkToComponent onCollapse=this.collapse)
+      "bs-navbar/nav"
+      linkToComponent=(component "bs-navbar/link-to" onCollapse=this.collapse)
     )
     collapse=this.collapse
     expand=this.expand
     toggleNavbar=this.toggleNavbar
   ) as |yieldedHash|}}
+  {{log yieldedHash}}
   {{#if (macroCondition (macroGetOwnConfig "isBS4"))}}
     <nav class="navbar {{this.positionClass}} {{this.typeClass}} {{this.breakpointClass}} {{this.backgroundClass}}" ...attributes>
       {{#if this.fluid}}
--- a/addon/components/bs-navbar.hbs
+++ b/addon/components/bs-navbar.hbs
@@ -3,14 +3,13 @@
     toggle=(component (bs-default @toggleComponent (component "bs-navbar/toggle")) onClick=this.toggleNavbar collapsed=this._collapsed)
     content=(component (bs-default @contentComponent (component "bs-navbar/content")) collapsed=this._collapsed onHidden=this.onCollapsed onShown=this.onExpanded)
     nav=(component
-      "bs-navbar/nav"
+      (bs-default @navComponent (component "bs-navbar/nav"))
       linkToComponent=(component "bs-navbar/link-to" onCollapse=this.collapse)
     )
     collapse=this.collapse
     expand=this.expand
     toggleNavbar=this.toggleNavbar
   ) as |yieldedHash|}}
-  {{log yieldedHash}}
   {{#if (macroCondition (macroGetOwnConfig "isBS4"))}}
     <nav class="navbar {{this.positionClass}} {{this.typeClass}} {{this.breakpointClass}} {{this.backgroundClass}}" ...attributes>
       {{#if this.fluid}}
--- a/tests/integration/components/bs-modal-test.js
+++ b/tests/integration/components/bs-modal-test.js
@@ -49,7 +49,7 @@ module('Integration | Component | bs-modal', function (hooks) {
 
   test('clicking ok button closes modal when autoClose=true with custom component hierarchy', async function (assert) {
     this.owner.register(
-      'component:my-component',
+      'component:test-component',
       class extends Component {
         layout = hbs`{{yield}}`;
       }
@@ -57,10 +57,10 @@ module('Integration | Component | bs-modal', function (hooks) {
 
     await render(hbs`
       <BsModal @title="Simple Dialog" @body={{false}} @footer={{false}} as |modal|>
-        {{#my-component}}
+        {{#test-component}}
           <modal.body>Hello world!</modal.body>
           <modal.footer />
-        {{/my-component}}
+        {{/test-component}}
       </BsModal>
 
     `);
--- a/tests/integration/components/bs-tooltip-test.js
+++ b/tests/integration/components/bs-tooltip-test.js
@@ -138,10 +138,10 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     class DummyComponent extends Component {
       layout = hbs`<div>{{yield}}</div>`;
     }
-    this.owner.register('component:dum-my', DummyComponent);
+    this.owner.register('component:test-component', DummyComponent);
 
     await render(
-      hbs`{{#dum-my id="target"}}<BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" />{{/dum-my}}`
+      hbs`{{#test-component id="target"}}<BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" />{{/test-component}}`
     );
 
     await triggerEvent('#target', 'mouseenter');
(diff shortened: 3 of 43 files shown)


addon/components/bs-navbar.hbs
```diff
@@ -1,15 +1,16 @@
...
-      this.navComponent
-      linkToComponent=(component this.linkToComponent onCollapse=this.collapse)
+      "bs-navbar/nav"
+      linkToComponent=(component "bs-navbar/link-to" onCollapse=this.collapse)
     )
     collapse=this.collapse
     expand=this.expand
     toggleNavbar=this.toggleNavbar
   ) as |yieldedHash|}}
+  {{log yieldedHash}}
```
jelhan, line 13: Leftover from debugging.

```suggestion
```

tests/integration/components/bs-modal-test.js
```diff
@@ -49,18 +49,18 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+      'component:test-component',
       class extends Component {
         layout = hbs`{{yield}}`;
       }
     );
 
     await render(hbs`
       <BsModal @title="Simple Dialog" @body={{false}} @footer={{false}} as |modal|>
-        {{#my-component}}
+        {{#test-component}}
```
jelhan, line 60: Seems as we have missed that one when refactoring to angle bracket component invocation. As we are touching it now, let's fix it:

```suggestion
        <TestComponent>
```

tests/integration/components/bs-modal-test.js
```diff
@@ -49,18 +49,18 @@ module('Integration | Component | bs-modal', function (hooks) {
...
           <modal.body>Hello world!</modal.body>
           <modal.footer />
-        {{/my-component}}
+        {{/test-component}}
```
jelhan, line 63:
```suggestion
        </TestComponent>
```

tests/integration/components/bs-tooltip-test.js
```diff
@@ -138,10 +138,10 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     class DummyComponent extends Component {
       layout = hbs`<div>{{yield}}</div>`;
     }
-    this.owner.register('component:dum-my', DummyComponent);
+    this.owner.register('component:test-component', DummyComponent);
 
     await render(
-      hbs`{{#dum-my id="target"}}<BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" />{{/dum-my}}`
+      hbs`{{#test-component id="target"}}<BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" />{{/test-component}}`
```
jelhan, line 144:
```suggestion
      hbs`<TestComponent id="target"><BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" /></TestComponent>`
```

jelhan (commented):
Thanks a lot for working on that one! :clap: I'm sorry that I needed so long for the review.

Sometimes `(bs-default @passedIncomponent (component "default-component"))` is wrapped in another `{{component}}` helper even if no arguments are passed to it. An example is `addon/components/bs-dropdown/menu.hbs`, which has:

```hbs
      {{yield
        (hash
          item=(component (bs-default @itemComponent (component "bs-dropdown/menu/item")))
          link-to=(component (bs-default @linkToComponent (component "bs-dropdown/menu/link-to")))
          linkTo=(component (bs-default @linkToComponent (component "bs-dropdown/menu/link-to")))
          divider=(component (bs-default @dividerComponent (component "bs-dropdown/menu/divider")))
        )
      }}
```

This pattern is not used everywhere. An example for the opposite is yielded `title` in `addon/components/bs-modal/header.hbs`:

```hbs
         title=(bs-default @titleComponent (component "bs-modal/header/title")) 
```

I would prefer to only use a wrapping `{{component}}` helper if required to apply additional arguments.

> Forms haven't been updated yet

Any specific reason you left them out for know? Or just due to limited time?

> As you can see, this adds quite a bit of boilerplate, but this seems the only way for now to make it work.

To be honest I don't see it as a big issue. You are talking about requiring `{{bs-default}}`, aren't you? Some more code in the template but one property less in the corresponding JavaScript component. And I like that I can see the default component yielded in the template directly.

simonihmig:
> I would prefer to only use a wrapping {{component}} helper if required to apply additional arguments.

Absolutely. This was probably lazy copy pasting, will fix it!

> Any specific reason you left them out for know? Or just due to limited time?

Yes, the latter. And the fact that I expect some trouble with the way we dynamically determine the control component [here](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-form/element.js#L796-L805)

> To be honest I don't see it as a big issue. You are talking about requiring {{bs-default}}, aren't you?

Yes. And that double wrapping of the component helper. And we have some duplicated default values, like [here](https://github.com/kaliber5/ember-bootstrap/pull/1251/files#diff-aa0d518fb94f8351df46f7626de5a02eR19-R51) vs. the single source of truth we had before when defined on the component class.

But I was more worried that you might object here! :wink: Although I see the drawbacks, I still think the benefit of enabling tree shaking outweighs this.

addon/components/bs-navbar.hbs
```diff
@@ -1,15 +1,16 @@
...
   (hash
-    toggle=(component this.toggleComponent onClick=this.toggleNavbar collapsed=this._collapsed)
-    content=(component this.contentComponent collapsed=this._collapsed onHidden=this.onCollapsed onShown=this.onExpanded)
+    toggle=(component (bs-default @toggleComponent (component "bs-navbar/toggle")) onClick=this.toggleNavbar collapsed=this._collapsed)
+    content=(component (bs-default @contentComponent (component "bs-navbar/content")) collapsed=this._collapsed onHidden=this.onCollapsed onShown=this.onExpanded)
     nav=(component
-      this.navComponent
-      linkToComponent=(component this.linkToComponent onCollapse=this.collapse)
+      "bs-navbar/nav"
+      linkToComponent=(component "bs-navbar/link-to" onCollapse=this.collapse)
```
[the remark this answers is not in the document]
simonihmig: Hm, good catch! Idk honestly, maybe I just tried something out and forget to revert/cleanup. Will fix!

jelhan:
>> Any specific reason you left them out for know? Or just due to limited time?
> 
> Yes, the latter. And the fact that I expect some trouble with the way we dynamically determine the control component here

I see. I'm totally fine with doing that in a follow-up merge request. It's your call.

> And we have some duplicated default values, like here vs. the single source of truth we had before when defined on the component class.

Being able to see it in the template directly outweights the duplication for me. Being able to be treeshake with Embroider is the icing on the cake. 🙃

addon/components/bs-navbar.hbs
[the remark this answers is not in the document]
simonihmig: Done

simonihmig:
Also refactored all form related components. Everything passing now locally, still waiting for a release of https://github.com/embroider-build/embroider/pull/540

jelhan (approved):
Awesome work! :clap: 

One Bootstrap 3 test is failing. Have had a quick look but don't see how it could be related.

```
not ok 228 Chrome 85.0 - [69 ms] - Integration | Component | bs-form/element: adjusts validation icon position if there is an input group [BS3]
    ---
        actual: >
            null
        stack: >
            TypeError: Cannot read property 'style' of null
                at Object._callee25$ (http://localhost:7357/assets/tests.js:11132:89)
                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:9966:105)
                at _next (http://localhost:7357/assets/tests.js:9968:196)
        message: >
            Promise rejected during "adjusts validation icon position if there is an input group [BS3]": Cannot read property 'style' of null
        negative: >
            false
        browser log: |
    ...
```

simonihmig:
> Have had a quick look but don't see how it could be related.

It was, but took my a while to debug this. Should be fixed now.

simonihmig:
With the newly released embroider v0.26.0 this is now passing! :tada:

PR #1257, opened 2020-09-23
fix usage example in accordion docs

jelhan:
- The yielded object was named `Acc` but used as `acc.change`.
- Most of our documentation and ember ecosystem is using lowercase variables only.
- Replaced `onclick` property with `{{on}}` modifier.
- Replaced `{{action}}` helper with `{{fn}}` helper.


simonihmig (approved):
:+1:

PR #1259, opened 2020-09-24
Support both ember-concurrency 1.x and 2.x

maxfierke:
We've just released ember-concurrency 2.0.0-beta.1. It's a largely API-compatible re-engineering of ember-concurrency's internals to support tracked properties and decouple from Ember (shedding EmberObject, etc.). As a popular library which depends on ember-concurrency, it would be nice for `ember-power-select` to support both 1.x and 2.x versions concurrently so that folks can upgrade the version of ember-concurrency in their applications at their leisure without waiting for all the add-ons they use to switch over to a newer version (and risk leaving folks who cannot upgrade yet behind.)

This PR adds some scenarios to ember-try to enable testing against both major releases and relaxes the version restriction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,6 @@ jobs:
           - ember-canary
           - ember-default-with-jquery
           - ember-classic
-          - ember-concurrency-1.x
           - ember-concurrency-2.x
         bootstrap:
           - 3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
           - ember-canary
           - ember-default-with-jquery
           - ember-classic
+          - ember-concurrency-1.x
+          - ember-concurrency-2.x
         bootstrap:
           - 3
           - 4
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -108,6 +108,22 @@ module.exports = async function () {
         command: 'yarn run nodetest',
       },
       {
+        name: 'ember-concurrency-1.x',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^1.3.0',
+          },
+        },
+      },
+      {
+        name: 'ember-concurrency-2.x',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^2.0.0-beta.1',
+          },
+        },
+      },
+      {
         name: 'embroider-tests',
         npm: {
           devDependencies: {
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -108,20 +108,14 @@ module.exports = async function () {
         command: 'yarn run nodetest',
       },
       {
-        name: 'ember-concurrency-1.x',
-        npm: {
-          dependencies: {
-            'ember-concurrency': '^1.3.0',
-            bootstrap: bootstrapVersion,
-          },
-        },
-      },
-      {
         name: 'ember-concurrency-2.x',
         npm: {
           dependencies: {
             'ember-concurrency': '^2.0.0-beta.1',
             bootstrap: bootstrapVersion,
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -64,6 +64,8 @@ jobs:
           - ember-canary
           - ember-default-with-jquery
           - ember-classic
+          - ember-concurrency-1.x
+          - ember-concurrency-2.x
```
simonihmig, line 68: Do we really need two separate scenarios here? Like all the other scenarios are already using one of the versions, so we really just need one additional scenario to test with the other version, don't we?

Our test matrix is already quite big, and in this case two additional scenarios cause 4 additional test jobs, as each is running both for Bootstrap 3 and 4.

config/ember-try.js
```diff
@@ -107,6 +107,22 @@ module.exports = async function () {
...
+          dependencies: {
+            'ember-concurrency': '^1.3.0',
+          },
+        },
+      },
+      {
+        name: 'ember-concurrency-2.x',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^2.0.0-beta.1',
```
simonihmig, line 122:
```suggestion
            'ember-concurrency': '^2.0.0-beta.1',
            bootstrap: bootstrapVersion,
```

config/ember-try.js
```diff
@@ -107,6 +107,22 @@ module.exports = async function () {
         name: 'node-tests',
         command: 'yarn run nodetest',
       },
+      {
+        name: 'ember-concurrency-1.x',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^1.3.0',
```
simonihmig, line 114: The two scenarios are failing for Bootstrap 3, probably due to this one missing (see other scenarios):

```suggestion
            'ember-concurrency': '^1.3.0',
            bootstrap: bootstrapVersion,
```

simonihmig (commented):
Thanks for pro-actively working on this, @maxfierke! 

This is a good change, I remember the churn it caused when e-c was updating its 0.x minor, and causing conflicts everywhere...

Some tests are failing, but I think easy to fix, see me comments.

package.json
```diff
@@ -55,7 +54,7 @@
     "ember-cli-build-config-editor": "0.5.1",
     "ember-cli-htmlbars": "^5.1.2",
     "ember-cli-version-checker": "^5.1.1",
-    "ember-concurrency": "^1.3.0",
+    "ember-concurrency": "^1.3.0 <3",
```
maxfierke, line 57: messed up the version constraint here... should be `>=1.3.0 <3`. will update

.github/workflows/ci.yml
maxfierke: Not really, but it does protect against losing coverage if the version in `yarn.lock` changes. I can limit it to just the `ember-concurrency-2.x` add, though.

jelhan: I think we should enforce a default version explicitly using yarn resolutions and only cover the other one with an explicit scenario.

I would be fine with using v2 as the default one starting from now. Than we can drop support for v1 in the next major.

simonihmig: With the current version specifier in package.json, e-c 2.0.0 *would* get used by default if released, but not 2.0.0-beta.x, right? Or at least we need to update `yarn.lock`.

So can we just wait for the stable release, and then drop the `ember-concurrency-2.x` scenario?

maxfierke: I'll keep the resolved version pinned at 1.x and remove the specific test setting for 1.x, and leave it up to your discretion on when to switch to 2.x

simonihmig (approved):
Thanks again! :+1:

Issue #1258, opened 2020-09-23
Overriding templates does not work anymore w/ colocated addon templates


simonihmig:
In #1221 we refactored the ember-bootstrap components internally to use template colocation. 

While this first seemed like a non-breaking change (all tests passing), it seems that the usual way of overriding template of an addon in an app does *not* work anymore. Idk exactly how colocation is implemented, but looking at the RFC, specifically the [build-time section](https://github.com/emberjs/rfcs/blob/master/text/0481-component-templates-co-location.md#build-time-transformations), I read this as a way to tightly bundle class and template together. So putting a template with the same name as the addon's template into the app does not work anymore to override it. Seems the only way to do it is by explicitly calling `setComponentTemplate` as in this example:

```js
import BaseAlert from 'ember-bootstrap/components/bs-alert';
import layout from './template';

export default setComponentTemplate(layout, class Alert extends BaseAlert {});
```

In ember-bootstrap we explicitly told users that extending from classes is not supported anymore (to better allow internal refactorings, e.g. to Glimmer components), but we did not mention anything specific with regards to templates. So not sure if this can/must be considered an (accidental) breaking change? 🤔

/cc @jelhan

jelhan:
To be honest I would consider overwriting a template of a component as a bad practice. It's even worse than extending it's class. I wouldn't consider either as supported. I have used that pattern myself in the past but it was a maintenance nightmare. I would love to support all needed customization through public API so that a consumer can customize as they need by a wrapping component instead.

Regarding the public API I would only consider invoking the component in a template as our public API. So neither extending a class nor overwriting a template.

And I would also push for that pattern in our internal composition of components. This is what I had in mind here: https://github.com/kaliber5/ember-bootstrap/pull/1022#discussion_r483985828

simonihmig:
> Regarding the public API I would only consider invoking the component in a template as our public API. So neither extending a class nor overwriting a template.

Yeah, I agree. Everything else makes maintenance extremely difficult. I just wanted to bring this up, as people (including me and you) have done so in the past, and we haven't explicitly stated anything about the template part. Maybe we should mention that somewhere, maybe alongside the don't-extend-classes warning!?

fran-worley:
Whilst I totally agree with your conclusions on this, We've just run into this issue as we had overridden the element and label templates to provide the required attribute to the label component so we can include a visual indicator for required fields.

[...]

PR #1263, opened 2020-09-25
render header for <BsAlert> using named blocks

jelhan:
Provide a declarative API to render an alert with a header:

```hbs
<BsAlert>
  <:header>Warning</:header>
  <:body><p>Named blocks are getting reality.</p></:body>
</BsAlert>
```

It's the same as:

```hbs
<BsAlert>
  <h4 class="alert-heading">Warning</h4>
  <p>Named blocks are getting reality.</p>
</BsAlert>
```

So actually it's already possible with existing APIs. It even requires less code with existing APIs. We decided to prefer it in our current project nevertheless for three reasons:

1. It provides a better developer experience as the developer does not need to care about Bootstrap markup and class namings. `<:header>` is easier to read and remember than `<h4 class="alert-heading">`.
2. `<h4 class="alert-heading">` invites for derivation. `<:header>` makes that less likely and therefore helps to ensure same look across the application (or suite of applications).
3. If needed a wrapper around the body or a different class on the header can be applied. This may help with upgrades, e.g. if that part of the markup changes in Bootstrap 5.

I'm not very much opinionated if this belongs in Ember Bootstrap or not. If you don't see the value, we can also keep it in our UI library build on top of Bootstrap.

### DIFF
--- a/addon/components/bs-alert.hbs
+++ b/addon/components/bs-alert.hbs
@@ -9,6 +9,19 @@
         <span aria-hidden="true">&times;</span>
       </button>
     {{/if}}
-    {{yield}}
+
+    {{#if (has-block "header")}}
+      {{#let (element (bs-default @headerTag "h4")) as |Tag|}}
+        <Tag class="alert-heading">
+          {{yield to="header"}}
+        </Tag>
+      {{/let}}
+    {{/if}}
+
+    {{#if (has-block "body")}}
+      {{yield to="body"}}
+    {{else}}
+      {{yield}}
+    {{/if}}
   {{/unless}}
 </div>
--- a/addon/helpers/bs-default.js
+++ b/addon/helpers/bs-default.js
@@ -1,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+export function bsDefault(params) {
+  return params[0] ?? params[1];
+}
+
+export default helper(bsDefault);
--- a/app/helpers/bs-default.js
+++ b/app/helpers/bs-default.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-bootstrap/helpers/bs-default';
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "ember-element-helper": "^0.3.1",
     "ember-focus-trap": "^0.4.0",
     "ember-in-element-polyfill": "^1.0.0",
+    "ember-named-blocks-polyfill": "^0.2.3",
     "ember-on-helper": "^0.1.0",
     "ember-popper": "^0.11.3",
     "ember-ref-bucket": "^1.0.1",
--- a/tests/dummy/app/templates/application.hbs
+++ b/tests/dummy/app/templates/application.hbs
@@ -1,1 +1,18 @@
+<BsAlert>Plain</BsAlert>
+
+{{!--
+<BsAlert>
+  <:header>Header only</:header>
+</BsAlert>
+
+<BsAlert>
+  <:body>Body only</:body>
+</BsAlert>
+--}}
+
+<BsAlert>
+  <:header>Header</:header>
+  <:body>Body</:body>
+</BsAlert>
+
 {{outlet}}
--- a/tests/integration/components/bs-alert-test.js
+++ b/tests/integration/components/bs-alert-test.js
@@ -119,4 +119,28 @@ module('Integration | Component | bs-alert', function (hooks) {
     await a11yAudit();
     assert.ok(true, 'A11y audit passed');
   });
+
+  test('it allows to set a header using named blocks', async function (assert) {
+    await render(hbs`
+      <BsAlert>
+        <:header>Warning</:header>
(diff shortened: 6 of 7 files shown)


jelhan:
Tests are failing in FastBoot. Haven't invested the time yet to look into. I expect it to be a bug in the named blocks polyfill. Wanted to get some feedback first if we want to merge it or not. We aren't using FastBoot. So I wouldn't need to fix if it stays in our UI library. I'm lazy you know. :stuck_out_tongue_winking_eye:

simonihmig:
I like that! 

Yeah, the failing test is most likely due to the polyfill. I hope we don't run into more issues in case the polyfill is just not stable enough yet!? 🤔

Also this new API unlocks a few API improvements, eventually breaking changes. Like we could convert some usage of contextual components to named blocks where this makes sense. For example yielding header, footer and body from `<BsModal>` does not make that much sense anymore, as you would want to use one of these components more than one time, or in a different order than header > body > footer. Still have to get used to that API though, to fully grasp it...

jelhan:
> I like that!

Great to hear!

> Yeah, the failing test is most likely due to the polyfill. I hope we don't run into more issues in case the polyfill is just not stable enough yet!? :thinking:

To be honest I'm a little bit worried to face a similar early adopter situation as with `@embroider/macros`. But yieldable named blocks nearly landed in 3.20. So I hope that it will be stable sooner.

> Also this new API unlocks a few API improvements, eventually breaking changes. Like we could convert some usage of contextual components to named blocks where this makes sense. For example yielding header, footer and body from `<BsModal>` does not make that much sense anymore, as you would want to use one of these components more than one time, or in a different order than header > body > footer.

Yes. But I think / hope that it won't be breaking changes. E.g. for `<BsModal>` we should be able to use named blocks while still yielding the components as we do currently if invoked without named blocks.

I think we get most value out of it for cases in which it allows us to provide access to low-level primitives without requiring us to introduce breaking changes. The `label` rendered by `<BsForm::element>` is such an example as explored here: https://github.com/kaliber5/ember-bootstrap/issues/1258#issuecomment-698773260

> Still have to get used to that API though, to fully grasp it...

Same here. I especially learned that it isn't a good fit for all cases. I came up with three question so far which helps me to distinguish the cases, in which yieldable named blocks may be a good fit from the once which are better served with the custom solution:

1. Is the order of the elements fixed?
2. Should the consumer be able to wrap the rendered elements in custom markup?
3. May an element of the same type be rendered multiple times?

Only if all three questions could be answered with yes, yieldable named blocks are an option.

jelhan:
Fixed the FastBoot test. It wasn't an issue of the polyfil. Neither of the test. Nor of anything else meaningful. I just tricked myself by leaving some debugging stuff in application template of dummy app, which influenced the test. :facepalm: I better don't tell you how much time I wasted on debugging. :grimacing:

jelhan:
@simonihmig Ready to be merged from my side.

simonihmig:
@jelhan what's missing is some documentation of this new feature! :) I think adding an example with some explanation to the API docs, but maybe also a snippet to the demo app would be nice!

Speaking of that, do you know if the polyfill is automatically working for the host app when we have it as a dependency? Or does the host app also have to have it, when *they* use named blocks features? I assume it transpiles some template code using AST transforms, so does it do that only for our *own* usage, or globally? In the former case, we would have to tell the user they need to add the polyfill as well to make use of this new API.

jelhan:
> @jelhan what's missing is some documentation of this new feature! :) I think adding an example with some explanation to the API docs, but maybe also a snippet to the demo app would be nice!

Of course. Totally forgot about it. Didn't had the time to pick it up this week. Being in vacation next one. Will pick it up afterwards.

> Speaking of that, do you know if the polyfill is automatically working for the host app when we have it as a dependency? Or does the host app also have to have it, when _they_ use named blocks features? I assume it transpiles some template code using AST transforms, so does it do that only for our _own_ usage, or globally? In the former case, we would have to tell the user they need to add the polyfill as well to make use of this new API.

A consuming application must install the polyfill explicitly if using named blocks. Otherwise a build-time assertion is thrown. At least for reason Ember versions:

> Assertion Failed: Named blocks are not currently available, attempted to use the <:header> named block. ('application-consuming-example-addon/templates/application.hbs' @ L6:C2) 

Will mention that one in the documentation.

simonihmig:
> Didn't had the time to pick it up this week. Being in vacation next one. Will pick it up afterwards.

Ok, no problem. However I wanted to cut a release asap, at least with the new embroider compatibility. Should we merge this as-is, and add the docs later?

jelhan:
> @jelhan what's missing is some documentation of this new feature! :) I think adding an example with some explanation to the API docs, but maybe also a snippet to the demo app would be nice!

Done so. Ready to be merged from my side.

simonihmig (approved):

simonihmig:
Sorry for the delay, approved and merged! Thanks again, looking forward to see more usage of named blocks in the future! :)

PR #1271, opened 2020-09-30
Add Embroider compatibility note

simonihmig:
@jelhan you agree with that "policy"?


jelhan (approved):
I like being that honest. I don't think we can give more guarantees at this point and trying so will just make maintaining harder.

PR #1275, opened 2020-10-05
Enable splattributes on dropdown menu divider and add tests

### LINKED ISSUE #1262: Unable to set html attributes for dropdown-divider
Up until version v4.1.2, we had a `data-test` attribute set on the dropdown divider. This attribute is no longer rendered. Is this expected behaviour? If so I can change our tests to pickup the `.dropdown-divider` class but I thought I'd raise it incase this was an unexpected consequence of the move to co-located templates.

I'd love to pinpoint exactly what version this started (I know it works in v4.0.0), but I can't load v4.0.1 or v4.1.0 due to issues with the modifier changes that raise this error:

`Uncaught (in promise) Error: Compile Error: Unexpected Modifier ref @ 0..0`

fran-worley:
Fixes #1262


simonihmig (approved):
Thank you! 👍

PR #1279, opened 2020-10-06
Refactor Tooltips and Popovers to Glimmer components

simonihmig:


simonihmig:
Blocked on https://github.com/embroider-build/embroider/pull/560

addon/components/bs-contextual-help.js
```diff
@@ -191,10 +207,7 @@ export default class ContextualHelp extends Component {
    * @readonly
    * @private
    */
-  @computed
-  get overlayId() {
-    return `overlay-${guidFor(this)}`;
-  }
+  overlayId = `overlay-${guidFor(this)}`;
```
jelhan, line 210: Why aren't you using a getter here?

```suggestion
  get overlayId() {
    return `overlay-${guidFor(this)}`;
  }
```

The current syntax allows `overlayId` to be set. But it should be read-only, shouldn't it? Are you afraid of performance issues if not caching it? I would have assumed that `guidFor()` util uses a cache and therefore doesn't have a relevant performance impact.

addon/components/bs-contextual-help/element.js
```diff
@@ -113,16 +98,15 @@ export default class ContextualHelpElement extends Component {
    * @type {object}
    * @private
    */
-  @computed('arrowClass', 'autoPlacement', 'viewportElement', 'viewportPadding')
   get popperModifiers() {
-    let self = this;
+    let id = this.args.id;
```
jelhan, line 102:
```suggestion
    let { id } = this.args;
```

Just coding style. Feel free to ignore.

tests/integration/components/bs-tooltip-test.js
```diff
@@ -141,7 +141,7 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     this.owner.register('component:test-component', DummyComponent);
 
     await render(
-      hbs`{{#test-component id="target"}}<BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" />{{/test-component}}`
+      hbs`{{#test-component id="target"}}<div><BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" /></div>{{/test-component}}`
```
jelhan, line 144: Noticed a left-over from angle bracket component invocation refactoring:

```suggestion
      hbs`<TestComponent id="target"><div><BsTooltip @title="Dummy" @fade={{false}} @triggerElement="parentView" /></div></TestComponent>`
```

jelhan (commented):
Awesome work. :clap: I'm still impressed how much easier it gets to read and reason about after refactoring to `@glimmer/component`. Haven't expected it to make such a big difference.

I think the question on default values if argument passed in is `undefined` also impacts these one. We would need to add some `macroCondition` on `useDefaultValueIfUndefined` optional feature if we decide to continue supporting `useDefaultValueIfUndefined = false`. But let's have that discussion at https://github.com/kaliber5/ember-bootstrap/pull/1282#discussion_r502948017 to not split it up over multiple places.

Additionally I have some questions about implementation of `InState` class.

Beside that everything looks good to me.

addon/components/bs-contextual-help.js
```diff
@@ -1,37 +1,48 @@
...
+class InState {
   hover = false;
   focus = false;
   click = false;
 
-  @or('hover', 'focus', 'click')
-  showHelp;
+  set(key, value) {
+    this[key] = value;
+  }
```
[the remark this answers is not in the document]
simonihmig: > Does this method exists only to ease refactoring or did I missed something?

Yeah, I basically just tried to keep this PR as minimal as possible, and only focus on the Glimmer related changes. There is still room for further refactorings, like eventually moving this `InState` object into the component itself.

addon/components/bs-contextual-help.js
```diff
@@ -1,37 +1,48 @@
...
+
+  get showHelp() {
+    return this.hover || this.focus || this.click;
+  }
 }
 
+const HOVERSTATE_NONE = 0;
+const HOVERSTATE_IN = 1;
+const HOVERSTATE_OUT = 2;
```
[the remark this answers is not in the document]
simonihmig: Makes sense, will update this.

addon/components/bs-contextual-help.js
simonihmig: It's basically a constant value, as it never changes after component instantiation, so I don't see a value in having a getter (that *could* react to changes over time) instead of a more simple initializer. 🤷‍♂️

jelhan: Seems we are missing a const property per class instances. So both styles have trade-offs. Maybe it would be good to document that these property should be considered read-only? I got confused by that syntax cause I assumed that you aren't using a getter cause it could be set somewhere.

simonihmig: > Seems we are missing a const property per class instances

What do you mean by this?

> Maybe it would be good to document that these property should be considered read-only?

It is! 😀  Line 207. And it's private anyway, so shouldn't be used outside of the class.

jelhan: > > Seems we are missing a const property per class instances
> 
> What do you mean by this?

I imagine a language feature here, which does not exist as far as I know. Maybe something like:

```js
class Foo {
  const id = guidFor(this);
}
```

A property defined as `const` could not be changed. Trying to set it would throw an exception. So it would be similar to a native getter without a setter in that regard.

But other than a native getter it should be guaranteed to not change over time. So it would be similar to a const variable in that regard.

Other than a static class property the value would be assigned per instance. So it would not be the same for all instances of that class. Additionally it could depend on that instance. As shown in the example.

But I'm getting off topic here. 🙈 As it's already documented as read-only (thanks for pointing it out) I will close this thread as resolved.

addon/components/bs-contextual-help.js
```diff
@@ -124,8 +146,8 @@ export default class ContextualHelp extends Component {
    * @default 0
    * @public
    */
-  @reads('delay')
-  delayShow;
+  @arg
+  delayShow = this.args.delay ?? 0;
```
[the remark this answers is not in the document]
simonihmig: Obsolete, due to #1283

addon/components/bs-contextual-help.js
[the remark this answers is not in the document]
simonihmig: Done.

addon/components/bs-contextual-help.js
```diff
@@ -1,37 +1,48 @@
...
 import { deprecate } from '@ember/application/deprecations';
+import arg from '../utils/decorators/arg';
+import { tracked } from '@glimmer/tracking';
+import { getParentView } from '../utils/dom';
 
-class InState extends EmberObject {
+class InState {
   hover = false;
   focus = false;
   click = false;
```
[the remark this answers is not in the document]
simonihmig: No, they don't need to, as they are only read from class methods, there are not used in templates in any reactive way.

addon/components/bs-contextual-help.js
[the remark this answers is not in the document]
simonihmig: I refactored that now a bit, to remove that class and keep that part of state directly on the component.

PR #1281, opened 2020-10-11
Fixed onDismissed being called twice when visible changes value

### LINKED ISSUE #1264: @onDismissed event of <BsAlert> is called twice
The action bound to `@onDismissed` event of `<BsAlert>` is called twice. Both if dismissed using the rendered button and if the property bound to `@visible` argument is set to `false`.

## Reproduction

1. Create a new ember application and install latest version of Ember Bootstrap (v4.1.2).
2. Create this application template:
    ```hbs
    <button {{on "click" this.hideAlert}}>
      hide
    </button>

    <BsAlert
      @onDismissed={{this.hideAlert}}
      @visible={{this.showAlert}}
    >
      an alert
    </BsAlert>
    ```
3. Create this application controller:
    ```js
    import Controller from '@ember/controller';
    import { action } from '@ember/object';
    import { tracked } from '@glimmer/tracking';

    export default class ApplicationController extends Controller {
      @tracked showAlert = true;

      @action
      hideAlert() {
        console.log('hideAlert');
        this.showAlert = false;
      }
    }
    ```

I expected to see `hideAlert` console message only once if alert is dismissed. But it's shown twice in console.

ijlee2:
## Description

This pull request addresses and closes #1264.

## How to Review

I recommend reviewing commits one by one. In the 1st commit, I introduced failing tests that should pass once the bug is fixed.

### DIFF
--- a/addon/components/bs-alert.js
+++ b/addon/components/bs-alert.js
@@ -163,6 +163,10 @@ export default class Alert extends Component {
    * @private
    */
   hide() {
+    if (this.hidden) {
+      return;
+    }
+
     if (this.usesTransition) {
       later(
         this,
--- a/tests/integration/components/bs-alert-test.js
+++ b/tests/integration/components/bs-alert-test.js
@@ -1,6 +1,5 @@
 import { module } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { set } from '@ember/object';
 import { click, render, settled, waitFor } from '@ember/test-helpers';
 import { test, testRequiringTransitions } from '../../helpers/bootstrap';
 import hbs from 'htmlbars-inline-precompile';
@@ -98,7 +97,7 @@ module('Integration | Component | bs-alert', function (hooks) {
       hbs`<BsAlert @fade={{true}} @onDismissed={{this.dismissed}} @visible={{this.isAlertShown}}>Test</BsAlert>`
     );
 
-    set(this, 'isAlertShown', false);
+    this.set('isAlertShown', false);
     await settled();
 
     assert.ok(action.calledOnce, 'onDismissed is called once.');
@@ -113,7 +112,7 @@ module('Integration | Component | bs-alert', function (hooks) {
       hbs`<BsAlert @fade={{false}} @onDismissed={{this.dismissed}} @visible={{this.isAlertShown}}>Test</BsAlert>`
     );
 
-    set(this, 'isAlertShown', false);
+    this.set('isAlertShown', false);
     await settled();
 
     assert.ok(action.calledOnce, 'onDismissed is called once.');
--- a/tests/integration/components/bs-alert-test.js
+++ b/tests/integration/components/bs-alert-test.js
@@ -118,6 +118,21 @@ module('Integration | Component | bs-alert', function (hooks) {
     assert.ok(action.calledOnce, 'onDismissed is called once.');
   });
 
+  test('When fade is false, onDismissed is not called when visible changes from false to false', async function (assert) {
+    let action = sinon.spy();
+    this.dismissed = action;
+    this.isAlertShown = false;
+
+    await render(
+      hbs`<BsAlert @fade={{false}} @onDismissed={{this.dismissed}} @visible={{this.isAlertShown}}>Test</BsAlert>`
+    );
+
+    this.set('isAlertShown', false);
+    await settled();
+
+    assert.ok(action.notCalled, 'onDismissed is not called.');
+  });
+
   test('alert is initially hidden when visible=false', async function (assert) {
     await render(hbs`<BsAlert @type="success" @fade={{false}} @visible={{false}}>Test</BsAlert>`);
 
--- a/tests/integration/components/bs-alert-test.js
+++ b/tests/integration/components/bs-alert-test.js
@@ -1,6 +1,7 @@
 import { module } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, click, waitFor } from '@ember/test-helpers';
+import { set } from '@ember/object';
+import { click, render, settled, waitFor } from '@ember/test-helpers';
 import { test, testRequiringTransitions } from '../../helpers/bootstrap';
 import hbs from 'htmlbars-inline-precompile';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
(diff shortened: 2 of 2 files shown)


addon/components/bs-alert.js
```diff
@@ -163,6 +163,10 @@ export default class Alert extends Component {
    * @private
    */
   hide() {
+    if (this.hidden) {
+      return;
+    }
```
ijlee2, line 168: Not a fan of adding a guard for an easy way out... I'd prefer if we can fix the underlying issue, but I wasn't sure what that issue is due to `_visible`, `@visible`, `hidden` seemingly tracking the same thing (I didn't investigate how these are different).

If we do go with adding the guard, I think I'd recommend adding an inline comment to explain its purpose.

tests/integration/components/bs-alert-test.js
```diff
@@ -88,6 +89,36 @@ module('Integration | Component | bs-alert', function (hooks) {
...
+    let action = sinon.spy();
+    this.actions.dismissed = action;
+    this.isAlertShown = true;
+
+    await render(
+      hbs`<BsAlert @fade={{true}} @onDismissed={{action "dismissed"}} @visible={{this.isAlertShown}}>Test</BsAlert>`
+    );
+
+    set(this, 'isAlertShown', false);
+    await settled();
```
ijlee2, line 102: I think these lines of code mimick clicking `<button>hide</button>` to trigger `this.hideAlert` in @jelhan's [bug reproduction](https://github.com/kaliber5/ember-bootstrap/issues/1264#issue-710100120)

ijlee2:
@jelhan @simonihmig I noticed that the canary tests had failed during a [scheduled CI](https://github.com/kaliber5/ember-bootstrap/actions/runs/296777197). The same canary tests seem to be failing in this pull request. Could you have a look?

My current solution to the issue to add a guard to control the state machine. I feel like there can be a better solution, however. What would you suggest that we do?

addon/components/bs-alert.js
simonihmig: Yeah, I'll also prefer finding the root cause for this... maybe putting a break point into the `hide` method to see why it get's called twice? I am unsure why that is, given the [did-update](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-alert.hbs#L4) modifier should call it only once.

tests/integration/components/bs-alert-test.js
```diff
@@ -1,6 +1,7 @@
 import { module } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, click, waitFor } from '@ember/test-helpers';
+import { set } from '@ember/object';
```
simonihmig, line 3: Any reason to use the imported `set()` here? I understand using it helps for objects that you don't know if they extend from `Ember.Object`, i.e. have a `.set()`. But tbh I am not a fan of `set(this, ...)` when we *do know*  what `this` is. So I'd prefer just using `this.set()` below, as this would be also more consistent with the rest of the codebase.

tests/integration/components/bs-alert-test.js
simonihmig: Is this test indeed failing without your guard?

simonihmig (commented):
Thanks for working on this!

addon/components/bs-alert.js
ijlee2: I'm not sure yet why either. Yep, I can look into it further this week.

tests/integration/components/bs-alert-test.js
ijlee2: I added the tests first to check that they fail. When I added the guard, I saw that these tests passed. I haven't verified if adding the guard works by linking the addon to a new app, however.

tests/integration/components/bs-alert-test.js
ijlee2: I added `set()` by habit, heh.

In the tests, I initialized `this.isAlertShown` to `true`. I think, if we subsequently write `this.isAlertShown = false` instead of `set(this, 'isAlertShown', false)`, we get an error?

I haven't checked if we still get the error nowadays.

tests/integration/components/bs-alert-test.js
```diff
@@ -88,6 +89,36 @@ module('Integration | Component | bs-alert', function (hooks) {
     assert.ok(action.calledOnce, 'onDismissed has been called.');
   });
 
+  test('When fade is true, onDismissed is called once when visible changes from true to false', async function (assert) {
+    let action = sinon.spy();
+    this.actions.dismissed = action;
+    this.isAlertShown = true;
+
+    await render(
+      hbs`<BsAlert @fade={{true}} @onDismissed={{action "dismissed"}} @visible={{this.isAlertShown}}>Test</BsAlert>`
```
jelhan, line 98: For readability and as it's more future proof I prefer this syntax to assign actions in tests:

```suggestion
    this.dismissed = action;
    this.isAlertShown = true;

    await render(
      hbs`<BsAlert @fade={{true}} @onDismissed={{this.dismissed}} @visible={{this.isAlertShown}}>Test</BsAlert>`
```

But that's only about coding style. Feel free to ignore.

addon/components/bs-alert.js
jelhan: I expected that the bug is caused by setting the property bound to `@visible` to `false` twice.

In the reproduction given in #1264 the `hideAlert` method is called twice. First by the event listener attached to the button and afterwards by `@onDismissed` event. `hideAlert` method sets `showAlert`, which is bound to `@visible` to `false` even if it's `false` before. This invalidates `_visible` property of the component, which triggers `{{did-update}}` modifier.

A simplified reproduction of that might look like this:

```js
this.visible = true;
await render(hbs`<BsAlert @visible={{this.visible}}>`);

// this is mocking the event listener on button
this.visible = false;

// await at least one runloop cause {{did-update}} modifier is not fired
// multiple times for the same runloop
await new Promise((resolve) => next(resolve));

// this is mocking the action bound to @onDismissed event
this.visible = false;
```

If I recall correctly the bug could not only be reproduced by clicking the button but also with the dismiss button rendered by the component itself. In that case it's a little bit more complex as two different states are changing: clicking the dismiss button rendered by alert component changes the internal state of `<BsAlert>`, `@onDismissed` event changes the state passed in from outside (`this.showAlert` bound to `@visible` argument). But as in the example before `_visible` property of `<BsAlert>` is invalidated twice, which triggers the `{{did-update}}` modifier twice.

To summarize: I assumed that setting property bound to `@visible` argument is required to reproduce the bug. I'm a little bit confused cause the tests added doesn't seem to do so. So maybe my assumption is wrong. Haven't verified.

If my assumption is right, adding a guard as you have done here, seems to be the correct fix in my opinion.

simonihmig:
> I noticed that the canary tests had failed

should be fixed by #1284

tests/integration/components/bs-alert-test.js
ijlee2: I like the suggested syntax and will update it, thanks.

tests/integration/components/bs-alert-test.js
ijlee2: Changed syntax to `this.set()`

addon/components/bs-alert.js
ijlee2: Apologies for my slow progress with this pull request.

> // this is mocking the event listener on button
> this.visible = false;
>
> // await at least one runloop cause {{did-update}} modifier is not fired
> // multiple times for the same runloop
> await new Promise((resolve) => next(resolve));
>
> // this is mocking the action bound to @onDismissed event
> this.visible = false;

After the `await render()` call in my tests, I tried including these lines of code.

The line `await new Promise((resolve) => next(resolve));` didn't trigger the `this.dismissed` function that I passed to the component (i.e. `@onDismissed` was called 0 times).

If I continued to use `await settled()` instead, `@onDismissed` was called 1 time. Not sure yet what to make of this observation.

jelhan (approved):
Looks good to me. I have verified that I can not reproduce the bug manually. Additionally it contains two additional tests, which fail on `master` but not with this fix.

If you still have some time to work on this one @ijlee2 an additional test case that `onDismissed` is *not* called when `@visible` changes from `false` to `false` would be awesome. Having `onDismissed` called every time when `visible` is dirtied even and it's `false` afterwards seems to be the root cause for the endless loop I'm seeing in https://github.com/kaliber5/ember-bootstrap/issues/1264#issuecomment-717056248.

ijlee2:
@jelhan Thanks for reviewing my code. Yep, I can work on adding an extra test case.

simonihmig:
I would like to get this merged asap, to cut a new release. My understanding is that this is basically ready, given that it fixes the original issue, with the some possible "bonus points" for adding that additional test, right? 

@ijlee2 we can remove the draft status and merge it from your point of view, right?

ijlee2:
@simonihmig Yes, you are correct, thanks for confirming with me. Sorry for the delays. 😓

simonihmig (approved):
Thank you @ijlee2 for your contribution! :+1:

PR #1283, opened 2020-10-12
Apply default values even if undefined is passed, allowing for easy wrapping of components

simonihmig:
As discussed in https://github.com/kaliber5/ember-bootstrap/pull/1282#discussion_r502948017, we remove the optional feature introduced in #1248 again, and make its behavior the default one.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -158,11 +158,11 @@ export default class Form extends Component {
    * Moreover the form's `submit` event handler will validate the model and deny submitting if the model is not validated successfully.
    *
    * @property model
-   * @type Ember.Object
+   * @type object
    * @public
    */
   @defaultValue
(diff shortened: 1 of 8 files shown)


simonihmig:
Only failing for Canary, which should be fixed by #1284

jelhan (approved):

jelhan:
I think this should be mentioned in the API docs. Not that sure about the title so. Technically it's about removing the optional feature. But from a consumer point of view it's about supporting passing arguments through without needing to worry about default values. I guess for better changelog the title should focus on that one.

simonihmig:
Changed the title. Feel free to change again if you have something better, at least until it's release time again! :wink:

simonihmig:
Just reverted that, and added the title and label to *your* PR, so you are correctly credited for that change! 😆

Issue #613, opened 2018-06-20
Touch gesture support Carousel


BHuys:
Good afternoon, is there or will there be any touch gesture support for the Carousel component? I would like to use gestures, swipe left/right to change the current slide etc.

simonihmig:
No, it has not, as even the original Bootstrap implementation does not support this AFAIK. I won't have the time to implement it by myself, but I am certainly happy for a PR!

musaffa:
As of v4.2, Bootstrap supports touch gesture. twbs/bootstrap#25776

sergchernata:
This would be pretty nice.

jelhan:
Happy to review and merge a pull request if someone has time to work on it.

PR #1300, opened 2020-10-20
add handle to prev/next slide

nolman:
This exposes the `toNextSlide` and `toPrevSlide` functions so that they can be called. I am doing this so that I can mixing ember-gestures (hammer.js) to add support for left/right swipe events. This can enable the requested functionality in
https://github.com/kaliber5/ember-bootstrap/issues/613.

### DIFF
--- a/addon/components/bs-carousel.hbs
+++ b/addon/components/bs-carousel.hbs
@@ -25,6 +25,8 @@
           orderClassName=this.orderClassName
           presentationState=this.presentationState
         )
+        toNextSlide=this.toNextSlide
+        toPrevSlide=this.toPrevSlide
       )
     }}
   </div>


jelhan:
Thanks a lot for working on this! Can you please add documentation and tests as well?

nolman:
Happy to add documentation however sure where/what to add for documentation.

Here is how I am using it (requires a third party library).
```template.hbs
<BsCarousel as |carousel|>
  {{#each @model.imageUrls as |imageUrl|}}
    <carousel.slide>
      <img
        src={{imageUrl}}
        {{recognize-gesture "swipe"}}
        {{on "swipe" (fn this.swipe carousel)}} />
    </carousel.slide>
  {{/each}}
</BsCarousel>
```

with a recognizer `app/ember-gestures/recognizers/swipe.js`
```swipe.js
export default {
  include: [],
  exclude: [],
  options: { threshold: 25, direction: typeof Hammer === 'undefined' ? '' : Hammer.DIRECTION_HORIZONTAL },
  recognizer: 'swipe'
};
```

jelhan:
The yielded functions should be mentioned in the API docs: https://github.com/kaliber5/ember-bootstrap/blob/c35a790696961d1666099262518664027104f923/addon/components/bs-carousel.js#L15-L64 The documentation should cover what they are meant for and how they work. Additionally a simple usage example should be given.

Ember Bootstrap's API documentation currently does not provide first-class support for documenting arguments and yielded values / functions. Yielded values and function are currently documented in the overview text at the beginning. Please have a look at [documentation for `<Form::Element>`](https://www.ember-bootstrap.com/api/classes/Components.FormElement.html) and [`<Modal>`](https://www.ember-bootstrap.com/api/classes/Components.Modal.html) for examples how that is done.

Please don't get confused that the component is missing in deployed API docs. This is a bug, which I just noticed. Opened #1301 to track that one.

jelhan:
Thanks for adding the documentation at @nolman. Looks good to me. Only tests are missing to get this merged from my side.

@simonihmig Do you have any concerns with introducing this new public API?

simonihmig:
> Do you have any concerns with introducing this new public API?

No, that addition looks good. Thanks @nolman! 

Yes, only some test similar to existing ones would be required to merge this.

jelhan:
@nolman Any chance you can add tests for this one? Feel free to reach out to me on Discord if you have any questions.

jelhan:
@nolman Would love to get this in. Do you think you will have some time to add the missing tests?

addon/components/bs-carousel.js
```diff
@@ -30,6 +30,11 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
   Default settings are the same as the original so you don't have to worry about changing parameters.
 
+  The component yields a hash with the following properties:
+  * `slide`: the component that would be used for rendering the carousel slide.
+  * `toPrevSlide`: function to change change the carousel to the previous slide.
+  * `toNextSlide`: function to change change the carousel to the next slide.
```
jelhan, line 36:
```suggestion
  * `toPrevSlide`: function to change the carousel to the previous slide.
  * `toNextSlide`: function to change the carousel to the next slide.
```

PR #1302, opened 2020-10-20
Update release setup.

rwjblue:
The result of running `npx create-rwjblue-release-it-setup --update`.

This will prevent an issue that you will run into if you attempt to publish a new version while your `release-it` version is 14.2.0 and your `release-it-lerna-changelog` version is 2.4.0.

See https://github.com/release-it/release-it/issues/713 for more information.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,9 +1,8 @@
-# Release
+# Release Process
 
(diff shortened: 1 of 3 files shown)


jelhan:
Thanks a lot @rwjblue!

PR #1305, opened 2020-10-25
adding support for vertical dialog layout

marcemira:


jelhan:
Is this an official Bootstrap feature? Wasn't able to find it in Bootstrap docs.

marcemira:
Ohh, you're right. It isn't.. had a similar name with `modal-dialog-center` but it was actually coming from a theme. I'm sorry, let me close this.

Issue #1338, opened 2020-11-19
Tests for validation support in <BsForm> do not test validation plugin interface


jelhan:
The tests for validation support in `<BsForm>` and `<BsForm::Element>` rely on private properties. They don't test the interface for validation plugins but set arguments from outside.

Let's take this test as an example: https://github.com/kaliber5/ember-bootstrap/blob/50937097944ad7c8a0504088628fc1972606bea4/tests/integration/components/bs-form-test.js#L138-L170

It sets the properties `hasValidator` and `validate` by setting `@hasValidator` and `@validate` arguments. This is only working due to characteristics of `@ember/component`.

This does not tests validation provided by plugins as they exist. Both ember-bootstrap-changeset-validations and ember-bootstrap-cp-validations add validation support by subclassing `<BsForm::Element>`. This interface should be used in the tests as well.

PR #1339, opened 2020-11-19
Refactor <BsForm::Group> and <BsForm::Element> to glimmer component and tracked properties

jelhan:
This continues the journey of Octane refactoring. `<BsForm::Group>` and `<BsForm::Element>` are refactored to `@glimmer/component` and tracked properties.

I picked this two components because I hope that it also resolves two issues:

1. The current integration of Ember Changeset as validation provider is broken for recent Ember versions. The root problem is the integration of tracked properties used by Ember Changeset and computed properties used by `<BsForm::Element>`. Already before we have struggled a lot with integrating both worlds in `ember-bootstrap-changeset-validations`. But so far we haven't found any solution for an assertion added in Ember 3.21.2 and backported to 3.20.6. See https://github.com/kaliber5/ember-bootstrap-changeset-validations/issues/33 for details. Using tracked properties on both sides will remove the root of the problem.
2. We started using a custom `@elementComponent` with `<BsForm>`. Our custom element component is build on top of `<BsForm::Element>` by invoking it in the template. Doing so we uncovered some bugs related to default property handling. E.g. `@showValidationOn` argument was not using it's default value if the argument's value is `undefined`. Refactoring to `@glimmer/component` seemed to be the most valuable way to fix these bugs. As it's very unlikely to introduce these kind of bugs with `@glimmer/component` I avoided adding a test for it.

I noticed that some internal properties used for validation were overwritten by passing arguments in our test suite. `hasValidator`, `validate` and `errors` only need to be arguments cause they are used as such in existing tests. I didn't wanted to touch tests while doing the glimmer component refactoring. So I kept it as is but created an issue to not loose track of that one: https://github.com/kaliber5/ember-bootstrap/issues/1338

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -295,8 +295,6 @@ export default class FormElement extends FormGroup {
    * @property model
    * @public
    */
-  @arg
-  model = null;
 
   /**
    * Show a help text next to the control
@@ -327,18 +325,14 @@ export default class FormElement extends FormGroup {
    * @type {Array}
    * @public
    */
-  @arg
-  options = null;
 
   /**
-   * Property path (e.g. 'title' or 'related.name') to render the label of an selection option. See `options`.s
+   * Property path (e.g. 'title' or 'related.name') to render the label of an selection option. See `options`.
    *
    * @property optionLabelPath
    * @type {String}
    * @public
    */
-  @arg
-  optionLabelPath = null;
 
   /**
    * The array of error messages from the `model`'s validation.
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -213,8 +213,6 @@ export default class FormElement extends FormGroup {
    * @type string
    * @public
    */
-  @arg
-  label = null;
 
   /**
    * Controls label visibility by adding 'sr-only' class.
@@ -224,16 +222,6 @@ export default class FormElement extends FormGroup {
    * @default false
    * @public
    */
-
-  /**
-   * @property hasLabel
-   * @type boolean
-   * @readonly
-   * @private
-   */
-  get hasLabel() {
-    return isPresent(this.label);
-  }
 
   /**
    * The type of the control widget.
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -422,8 +422,6 @@ export default class FormElement extends FormGroup {
    * @type String
    * @public
    */
-  @arg
-  size = null;
 
   /**
    * The array of validation messages (either errors or warnings) from either custom error/warnings or , if we are showing model validation messages, the model's validation
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -341,6 +341,9 @@ export default class FormElement extends FormGroup {
    * @type array
    * @protected
    */
+  // This shouldn't be an argument. It's only an argument because tests rely on
+  // setting it as an argument. See https://github.com/kaliber5/ember-bootstrap/issues/1338
+  // for details.
   @arg errors;
 
   /**
@@ -360,6 +363,9 @@ export default class FormElement extends FormGroup {
    * @type array
    * @protected
    */
+  // This shouldn't be an argument. It's only an argument because tests rely on
+  // setting it as an argument. See https://github.com/kaliber5/ember-bootstrap/issues/1338
+  // for details.
   @arg warnings;
 
   /**
@@ -380,8 +386,6 @@ export default class FormElement extends FormGroup {
    * @type string
    * @public
    */
-  @arg
-  customError = null;
 
   /**
    * @property hasCustomError
@@ -390,7 +394,7 @@ export default class FormElement extends FormGroup {
    * @private
    */
   get hasCustomError() {
-    return isPresent(this.customError);
+    return isPresent(this.args.customError);
   }
 
   /**
@@ -402,8 +406,6 @@ export default class FormElement extends FormGroup {
    * @type string
    * @public
    */
-  @arg
-  customWarning = null;
 
   /**
    * @property hasCustomWarning
@@ -412,7 +414,7 @@ export default class FormElement extends FormGroup {
    * @private
    */
   get hasCustomWarning() {
-    return isPresent(this.customWarning);
+    return isPresent(this.args.customWarning);
   }
 
   /**
@@ -432,13 +434,13 @@ export default class FormElement extends FormGroup {
    */
   get validationMessages() {
     if (this.hasCustomError) {
-      return A([this.customError]);
+      return A([this.args.customError]);
     }
     if (this.hasErrors && this.showModelValidation) {
       return A(this.errors);
     }
     if (this.hasCustomWarning) {
-      return A([this.customWarning]);
+      return A([this.args.customWarning]);
     }
     if (this.hasWarnings && this.showModelValidation) {
       return A(this.warnings);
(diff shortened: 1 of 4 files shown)


jelhan:
This pull requests contain required changes in Ember Bootstrap Changeset Validations: https://github.com/kaliber5/ember-bootstrap-changeset-validations/pull/34

jelhan:
FastBoot tests are failing for some scenarios:

- Floating Dependencies
-  Scenario Tests (ember-default, ember test --launch Firefox) 
-  Scenario Tests (fastboot-tests)

Additionally the _Scenario Tests (embroider-optimized)_ is failing with a build-time error related to unsafe dynamic component usage. But it's for `bs-accordion.hbs`:

```
  - broccoliBuilderErrorStack: ModuleError: Module Error (from /home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/@embroider/webpack/src/webpack-hbs-loader.js):
Unsafe dynamic component: cannot statically analyze this expression in $TMPDIR/embroider/591ed8/components/bs-accordion.hbs
    at Object.emitError (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/webpack/lib/NormalModule.js:173:6)
    at Object.hbsLoader (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/@embroider/webpack/src/webpack-hbs-loader.js:15:14)
    at LOADER_EXECUTION (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/loader-runner/lib/LoaderRunner.js:119:14)
    at runSyncOrAsync (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/loader-runner/lib/LoaderRunner.js:120:4)
    at iterateNormalLoaders (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/loader-runner/lib/LoaderRunner.js:232:2)
    at Array.<anonymous> (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/loader-runner/lib/LoaderRunner.js:205:4)
    at Storage.finished (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:55:16)
    at provider (/home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:91:9)
    at /home/runner/work/ember-bootstrap/ember-bootstrap/node_modules/graceful-fs/graceful-fs.js:123:16
    at FSReqWrap.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:53:3)
```

I don't see how either of these error can be caused by my changes.

simonihmig:
> FastBoot tests are failing for some scenarios:

Yes, just fixed this by #1340. Also Embroider tests are failing in optimized mode. Working on it, but it will take a bit more effort. In the meantime, could you rebase your PR, then at least all scenarios except for optimized Embroider should be passing again.

I hope I can review this tomorrow, thanks for working on this!

jelhan:
> Yes, just fixed this by #1340. Also Embroider tests are failing in optimized mode. Working on it, but it will take a bit more effort. In the meantime, could you rebase your PR, then at least all scenarios except for optimized Embroider should be passing again.

Rebased this pull requests. As expected all scenarios except optimized Embroider are passing now. I saw that you have started working on that one as well in #1341. Thanks a lot for fixing the tests!

addon/components/bs-form/group.js
```diff
@@ -61,6 +58,7 @@ export default class FormGroup extends Component {
    * @type string
    * @public
    */
+  @localCopy('args.validation') validation;
```
simonihmig, line 61: We use `@localCopy` because we know that *subclasses* (like that from `e-b-c-v`) will need to mutate that (given our current validation API), right? Have you tested `e-b-c-v` using this branch, so we know this *really* works?

addon/components/bs-form/element.js
```diff
@@ -218,7 +213,7 @@ export default class FormElement extends FormGroup {
    * @type string
    * @public
    */
-  @defaultValue
+  @arg
```
simonihmig, line 216: Can't we change all arguments, that have either no initializer or `null`, to named args, i.e. remove them here from the class (keeping just the DocBlock) and use them as `@...` in the template? I think `null` (used here) and `undefined` (which they would be as named args when not passed) are used interchangeably here?

addon/components/bs-form/group.js
jelhan: To be honest I have no idea why I used `@localCopy` decorator here. It doesn't seem to be needed at all. I didn't found any place in which it's overwritten locally. Neither in ember-bootstrap itself nor in ember-bootstrap-changeset-validations nor in ember-bootstrap-cp-validations. I guess it's just a leftover from an early iteration? I removed it in c8d51ffdc662d1aecbdc531011972540b9242d9a.

> Have you tested e-b-c-v using this branch, so we know this really works?

Not sure if `e-b-c-v` refers to `ember-bootstrap-changeset-validations` or `ember-bootstrap-cp-validations`. I have tested with the first one. Pull request showing required changes to ember-bootstrap-changeset-validations and passing CI pipeline could be found here: https://github.com/kaliber5/ember-bootstrap-changeset-validations/pull/34/files I haven't tested with ember-bootstrap-cp-validations.

addon/components/bs-form/element.js
jelhan: Good catch. Tried to remove most of these cases. But seems as I missed something in the iterations. Removed some more properties that weren't needed anymore.

addon/components/bs-form/element.js
```diff
@@ -398,17 +360,17 @@ export default class FormElement extends FormGroup {
    * @type array
    * @protected
    */
-  @defaultValue
-  warnings = null;
+  @arg warnings;
```
simonihmig, line 363: We should be able to remove this?

addon/components/bs-form/element.js
```diff
@@ -418,7 +380,7 @@ export default class FormElement extends FormGroup {
    * @type string
    * @public
    */
-  @defaultValue
+  @arg
```
simonihmig, line 383: same?

addon/components/bs-form/element.js
```diff
@@ -439,7 +402,7 @@ export default class FormElement extends FormGroup {
...
```
simonihmig, line 405: same?

addon/components/bs-form/element.js
simonihmig: Looks good. I inline commented on a few more, that maybe could be removed. Feel free to ignore if that's not the case

addon/components/bs-form/group.js
simonihmig: > Not sure if e-b-c-v refers to ember-bootstrap-changeset-validations or ember-bootstrap-cp-validations. 

To the former, indeed! 😆. So all good...

simonihmig (approved):
I approved. Feel free to merge with either my new inline comments addressed, or ignored if you think that wouldn't work.

Thanks for tackling this! I guess that was the harder part of migrating the form components. A lot of the others are "just" about rendering markup, not so much behavior and logic. So those should hopefully be easy to migrate, mostly as template-only. 

Do you plan to continue working on this? Just asking so our work doesn't overlap eventually...

addon/components/bs-form/element.js
jelhan: I thought you prefer that alias if the argument is accessed in the JavaScript file several times. Happy to remove it. Did I got you wrong and you only prefer `@arg` usage if the argument has a default value, which is neither `undefined` nor `null`?

simonihmig: > you only prefer @arg usage if the argument has a default value, which is neither undefined nor null?

Yes, that would be my thinking (i.e. don't have duplicated default values). 

Hope this is consistent with any of my previous PRs, but not totally sure 😆

Issue #1386, opened 2021-01-04
Modal open property only updates once after update to 4.6.0


kboucher:
I recently updated from `ember-bootstrap@4.4.0` to `ember-bootstrap@4.6.0` in order to pick up the fix for [issue 1362](https://github.com/kaliber5/ember-bootstrap/issues/1362).

I notice that in addition to that fix, the updates include converting the Modal component to a Glimmer component.

In any case, it seems that the `@open` property is now only updating one time. I have an event handler for the `onHidden` event that toggles the property I am passing to `@open`, however, after opening and closing the modal one time, the modal no longer displays even though I can see the toggle property and the `@open` property in the `BsModalComponent` updating as expected in the Ember Inspector.

This is also my first Octane project, so I'm not entirely confident that I haven't done something wrong, but my setup worked in 4.4.0, and I have gone back to that version for now.

I will try to return to this ticket with any more useful information I can discern but wanted to raise the issue now.

**UPDATE**

I ran a `git bisect` and have identified the following commit as the _first bad commit_:

```
1bf4cf7aa2fe9aefa4e618b601877e1d4b0b5843 is the first bad commit
commit 1bf4cf7aa2fe9aefa4e618b601877e1d4b0b5843
Author: simonihmig <simon.ihmig@kaliber5.de>
Date:   Sun Dec 20 22:00:33 2020 +0100

    Call handleVisibilityChanges on public @open change

 addon/components/bs-modal.hbs | 2 +-
 addon/components/bs-modal.js  | 6 +++---
 2 files changed, 4 insertions(+), 4 deletions(-)
```

It also seems to be the case that when the modal fails to open on subsequent updates to the `@open` property, the following error is observed:

```
Uncaught Error: Assertion Failed: Backdrop element should be in DOM
```

Additionally, I've noticed that there is some undesired behavior related to the backdrop. (The commit where this arises is: `56ab45ed7f3ec1f01993421fd227d1feacbd5319`)

Specifically, the `.show` class is removed from the backdrop before the modal is de-rendered -- causing a solid black backdrop for a few seconds until the modal is de-rendered.

My implementation is pretty straight-forward:

`NavigationDrawerComponent`:
```
<BsModal
    class="navigation-drawer {{@side}}"
    @fade={{false}}
    @onHidden={{@onNavigationDrawerHidden}}
    @onShown={{this.onNavigationDrawerShown}}
    @open={{@showNavigationDrawer}}
    @scrollable={{true}}
[...]

Issue #830, opened 2019-06-15
BsModal is closed regarding of open property which causes strange edge cases


jelhan:
Current interaction of `<BsModal>`'s `open` property and closing functionality is confusing in my opinion.

1. The modal could be closed by clicking backdrop/close button even if `open` property stays `true`. Seeing `<BsModal @open={{true}}>...</BsModal>` I would expect that these modal can't be closed at all. But actually it could by closed by a click on the backdrop. One must return `false` from `onHide` action to prevent a modal from being closeable or disable all possibilities to close that modal.

2. If a modal is closed but `open` property is still `true` the modal could not be reopened by setting it to `true` again. It must be changed to some other value before. This is likely to cause bugs that are very difficult to debug.
E.g. imagine a modal being used to show an error message if some action fails. This is most likely been implemented by setting a property bound to `<BsModal>`'s `@open` to `true` in that action. If the developer isn't aware of the behavior described here, he will likely not set the property to `false` again when user closes the modal (using `<BsModal>`'s `onHide` or `onHidden` properties). In that case the modal (and with it the error message) won't be shown again if the error reoccurs...

It would be much easier to reason about `<BsModal>` and less error prune if a modal can't be closed / will always be shown if it's `@open` property is `true`. A typical usage would then be `<BsModal @open={{showModal}} @onHide={{action (mut showModal) false}}>`, which is much more inline with the other APIs of Ember Bootstrap and the rest of the ecosystem.

This would be a breaking change. So we might want to consider it for v3.

simonihmig:
This stems from the idea that the component should support both an *uncontrolled* as well as a *controlled* mode of operation. With that I basically mean who is responsible for owning the open-state of the component, see e.g. https://stackoverflow.com/questions/42522515/what-are-controlled-components-and-uncontrolled-components/42522792#42522792.

I think in most cases (like 99%+) you want the uncontrolled version, where the modal shows up by default, and will be closed when user does the corresponding interaction (close button, backdrop click). So no need the prevent the closing. In that case I would like `<BsModal>` to *just work*(TM), without the boilerplate of the controlled version.

In the usually rare case that you want to control the state, the component should allow you to do that, which is the `<BsModal @open={{showModal}} @onHide={{action (mut showModal) false}}>` example.

[...]

PR #1393, opened 2021-01-11
Fix modals not being able to open after being closed once

SanderKnauff:
Currently, modals can only be opened once. After this, it gets stuck in a state on which the BsModal thinks it is open already, causing the `show()` method to do nothing at all.

While the issue I encountered did not exactly fit issue #1386 (I did not receive any errors on my side), it might resolve that issue as well. I however do not know how to test that issue.

### DIFF
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -128,11 +128,6 @@ export default class Modal extends Component {
    * @default true
    * @public
    */
-
-  /**
-   * @property open
-   * @private
-   */
   @arg
   open = true;
 
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -473,7 +473,7 @@ export default class Modal extends Component {
     if (!this._isOpen) {
       return;
     }
-    this.isOpen = false;
+    this._isOpen = false;
 
     this.resize();
     this.showModal = false;
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -11,7 +11,6 @@ import isFastBoot from 'ember-bootstrap/utils/is-fastboot';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 import arg from '../utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
-import { localCopy } from 'tracked-toolbox';
 
 /**
   Component for creating [Bootstrap modals](http://getbootstrap.com/javascript/#modals) with custom markup.
@@ -61,25 +60,6 @@ export default class Modal extends Component {
   document;
 
   /**
-   * Visibility of the modal. Toggle to show/hide with CSS transitions.
-   *
-   * When the modal is closed by user interaction this property will not update by using two-way bindings in order
-   * to follow DDAU best practices. If you want to react to such changes, subscribe to the `onHide` action
-   *
-   * @property open
-   * @type boolean
-   * @default true
-   * @public
-   */
-
-  /**
-   * @property isOpen
-   * @private
-   */
-  @localCopy('args.open', true)
-  isOpen;
-
-  /**
    * @property _isOpen
    * @private
    */
@@ -108,7 +88,7 @@ export default class Modal extends Component {
    * @private
    */
   @tracked
-  showModal = this.isOpen && (!this._fade || isFastBoot(this));
+  showModal = this.open && (!this._fade || isFastBoot(this));
 
   /**
    * Render modal markup?
@@ -119,7 +99,7 @@ export default class Modal extends Component {
    * @private
    */
   @tracked
-  inDom = this.isOpen;
+  inDom = this.open;
 
   /**
    * @property paddingLeft
@@ -138,6 +118,25 @@ export default class Modal extends Component {
   paddingRight;
 
   /**
+   * Visibility of the modal. Toggle to show/hide with CSS transitions.
+   *
+   * When the modal is closed by user interaction this property will not update by using two-way bindings in order
+   * to follow DDAU best practices. If you want to react to such changes, subscribe to the `onHide` action
+   *
+   * @property open
+   * @type boolean
+   * @default true
+   * @public
+   */
+
+  /**
+   * @property open
+   * @private
+   */
+  @arg
+  open = true;
+
+  /**
    * Use a semi-transparent modal background to hide the rest of the page.
    *
    * @property backdrop
@@ -154,7 +153,7 @@ export default class Modal extends Component {
    * @private
    */
   @tracked
-  showBackdrop = this.isOpen && this.backdrop;
+  showBackdrop = this.open && this.backdrop;
 
   /**
    * Closes the modal when escape key is pressed.
@@ -515,7 +514,7 @@ export default class Modal extends Component {
   handleBackdrop(callback) {
     let doAnimate = this.usesTransition;
 
-    if (this.isOpen && this.backdrop) {
+    if (this.open && this.backdrop) {
       this.showBackdrop = true;
 
       if (!callback) {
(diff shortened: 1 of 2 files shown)


addon/components/bs-modal.js
```diff
@@ -473,7 +473,7 @@ export default class Modal extends Component {
     if (!this._isOpen) {
       return;
     }
-    this.isOpen = false;
+    this._isOpen = false;
```
jelhan, line 476: If I didn't missed something, this was the only place, in which `isOpen` was actually set. So we wouldn't need `@localCopy` on it anymore. It would be just an alias to `this.args.isOpen`.

If the big could be resolved by _not_ setting `isOpen`, would actually be inline with my findings in https://github.com/kaliber5/ember-bootstrap/issues/1386#issuecomment-756933816.

jelhan:
Thanks a lot for working on this. I _think_ you are on the right track here
 
> While the issue I encountered did not exactly fit issue #1386, it might resolve that issue as well. I however do not know how to test that issue.

Such a test would look very similar to the test you have. But instead closing the model by setting `@open` to `false`, you would close it with a click on backdrop (or close button). Additionally you would provide an action handler to `@onHidden`, which sets `@open` to false. So invokation would look like this:

```js
this.set('open', false);
this.set('close', () => this.open = false);

await render(hbs`
  <BsModal
    @open={{this.open}}
    @onHidden={{action this.close}}
  />
`);
```

addon/components/bs-modal.js
SanderKnauff: I've refactored out the @localCopy. I've tried to set the value to `this.args.open`, however it seems like that breaks more than it fixes. Accepting open as an argument with a default value of `true` does not seem to break any tests. Any calls to `this.isOpen` without the underscore has been changed to open.

SanderKnauff:
> Thanks a lot for working on this. I _think_ you are on the right track here
> 
> > While the issue I encountered did not exactly fit issue #1386, it might resolve that issue as well. I however do not know how to test that issue.
> 
> Such a test would look very similar to the test you have. But instead closing the model by setting `@open` to `false`, you would close it with a click on backdrop (or close button). Additionally you would provide an action handler to `@onHidden`, which sets `@open` to false. So invokation would look like this:
> 
> ```js
> this.set('open', false);
> this.set('close', () => this.open = false);
> 
> await render(hbs`
>   <BsModal
>     @open={{this.open}}
>     @onHidden={{action this.close}}
>   />
> `);
> ```

SanderKnauff:
(Apologies for closing, I misclicked)

I've updated the test as well as per your suggestions. The test is a bit closer to the case that was presented but is still in the same spirit as it was. 

As I've added more commits I'm wondering about the workflow. Should I squash these after everything has been cleared or will that be handled when merging?

addon/components/bs-modal.js
```diff
@@ -137,6 +117,25 @@ export default class Modal extends Component {
...
+   * @default true
+   * @public
+   */
+
+  /**
+   * @property open
+   * @private
+   */
+  @arg
+  open = true;
```
jelhan, line 137: Any reason you renamed the variable from `isOpen` to `open`? The `is` prefix indicates that it's a boolean. I fear `open` can be confused with a method to open the modal. I would prefer to keep `isOpen` to be honest. Would it be okay for you to change back?

addon/components/bs-modal.js
jelhan: Glad to see that we could actually drop `@localCopy` decorator. Less complexity is always good. :smile: 

Your changes look very good. I'm just a little bit confused why you haven't stuck with `isOpen` name. The `is` prefix is a helpful indicator that a property is a boolean in my opinion. In this special case I also fear that a property `open` might be confused with the `show` method.

Would it be okay for you to change the name back to `isOpen`? Not the implementation. That one is totally fine. I'm only talking about the naming.

jelhan (changes requested):
> I've updated the test as well as per your suggestions. The test is a bit closer to the case that was presented but is still in the same spirit as it was.

The new test looks very good. But I meant it as an addition not as a replacement of your original test. Both tests - the original one in your first commit and the current one - are testing two different cases. And it seems as we have missed test coverage for both.

I'm sorry for that confusion. Would it be okay for you to add the test again?

> As I've added more commits I'm wondering about the workflow. Should I squash these after everything has been cleared or will that be handled when merging?

You don't need to care about squashing. We can squash when merging.

addon/components/bs-modal.js
simonihmig: I haven't reviewed this PR yet, but I am a bit confused at this point: `@open` is the public argument, which we cannot rename. The JS docs above are wrong somehow, as there is `@property open` twice!

SanderKnauff: I've tried to rename `isOpen` to `open` to remove the dependency on the @localCopy and to clarify that it is an argument to the `BsModal` component. I must admit that I do not understand why there is both a `isOpen` and a `_isOpen`. The JSDoc seems indeed to be a bit odd, as `isOpen` had two JSDocs above it as if something between the docs was missing. 

While I do agree that `isOpen` would be a better name for a boolean, it does not reflect the actual args of the component well, causing confusion. I've tried a few ways of getting it to be a proper alias of open, but none of them were succesful.

SanderKnauff:
I've added back the original test case as well now

addon/components/bs-modal.js
jelhan: Oh. I missed that this has `@arg` decorator pointing to the argument. Sorry. Should have taken more time to review. Can not could how often I already got confused by `@open` argument and the two internal properties `isOpen` and `_isOpen`.

To be honest I thought this PR decouples `@open` from the internal state and manages the state updates through `{{did-update}}` modifier in the template only. I think I need to reserve some time to fully understand state flow in this component.

SanderKnauff: I'm not fully clear about the lifecycle of this component. I mostly noticed that the `_isOpen` state did not get put back to false, causing the modal to never reopen. It does seem to fix the odd behaviour people have been experiencing without breaking the existing test conditions.

addon/components/bs-modal.js
```diff
@@ -137,6 +117,25 @@ export default class Modal extends Component {
...
+   * @property open
+   * @type boolean
+   * @default true
+   * @public
+   */
+
+  /**
+   * @property open
+   * @private
+   */
```
simonihmig, line 135:
```suggestion
```

This JSDoc block should be removed, as it is wrongly stating that `@open` is private, and is in direct conflict with the doc block (referring to the same `@property open`) just a few lines above.

simonihmig (commented):
Apart from the minor JSDoc issue (see inline comment) this seems good to me!

Tbh I was a bit puzzled that we were able to simply remove that `@localCopy` without any breaking changes, but as all tests are passing (including the newly added ones :+1:), this seems to be the case.

Apparently I was also getting confused with those different states of the component. 🤦‍♂️
Initially that *mutable* `isOpen` copy of  `@open` was used to be able to directly close the modal by user interaction (ESC, backdrop click) without mutating (and thus violating DDAU) the original `@open` (which was possible in the Ember.Component days, but luckily became impossible with Glimmer components). Whether you like that API or not (see discussion in #830), we can only change that as a breaking change API overhaul in a major version, not now. Anyways, we seem to already have mutable component state like `showModal` and `inDom` that is enough for this purpose, with requiring `@localCopy`...

Thanks a lot for this work here @SanderKnauff! 🙏

addon/components/bs-modal.js
SanderKnauff: Good point! I've removed it now.

simonihmig (approved):

simonihmig:
This looks good from my point of view. @jelhan I believe your feedback has been addressed, or is not relevant anymore (isOpen vs open), right? If so, I would be happy to merge this!

jelhan (approved):
Thanks a lot @SanderKnauff for working on this. Removing `@localCopy` decorator simplifies the code a lot. The two tests will help us to catch regressions earlier next time.

I still feel that we can (and should) improve the state flow in this component. But that can be done in a follow-up pull request.

simonihmig:
Released: https://github.com/kaliber5/ember-bootstrap/releases/tag/v4.6.1 :tada:

tbartelmess:
Thank you!

PR #1401, opened 2021-01-19
remove unnecessary intermediate method in Modal class

jelhan:
Noticed that there is an unnecessary intermediate method in `<BsModal>` when digging deeper into that component to see how we can simplify the state.


simonihmig (approved):
:+1:

PR #1402, opened 2021-01-19
WIP: decouple internal and external open state of <BsModal>

jelhan:
This merge request decouples the internal and the external open state of `<BsModal>`.

It does not use the `@open` argument directly. All that argument does is triggering a `handleVisibilityChange` method.

In addition the class itself has a `state` property, which controls the internal visibility state of the modal. It may be `'opening'`, `'open'`, `'closing'` or `'closed'`.

This is what I meant in #1393.

It is still a WIP:

- Most of the tests are passing. But the FastBoot test is failing. I haven't had the time yet to look into why.
- Also I'm not sure if all the timings are still correct.  Don't want to rely only on our test coverage. Additionally I want to double check against Bootstrap reference implementation manually.

But I wanted to open it anyways to clarify what I meant and put it up for discussion.


addon/components/bs-modal.js
```diff
@@ -679,10 +686,12 @@ export default class Modal extends Component {
       return;
     }
 
-    if (this.args.open !== false) {
-      this.show();
-    } else {
-      this.hide();
-    }
+    next(() => {
```
simonihmig, line 689: Why is the next scheduling necessary? Might this be responsible for the failure in FastBoot?

simonihmig (commented):
The direction looks very good! :+1:

addon/components/bs-modal.js
jelhan: This prevents `state` being modified after it has been used already. Without scheduling the work for the next runloop we run into this deprecation:

> DEPRECATION: You attempted to update `state` on `wrapperClass`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

I think this is a common issue if a method called by a `ember-render-helpers` or `ember-render-modifiers` sets a property. And the reason why ember-render-helpers and ember-render-modifiers should be considered a footgun.

I'm not 100%ly sure why this is happening. But I _think_ the helper / modifier is not execute as part of the `actions` queue. So it must delay all changes to state for the next runloop.

Fastboot test is still failing without scheduling it for the next runloop. I _guess_ some FastBoot guards prevent the modal from being rendered. But I haven't fully looked into. Debugging FastBoot issues is way more painful than other stuff. :disappointed:

addon/components/bs-modal.js
```diff
@@ -90,17 +84,6 @@ export default class Modal extends Component {
   @tracked
   showModal = this.open && (!this._fade || isFastBoot(this));
```
simonihmig, line 85: Seems unused now?

But looking at this, this could also explain the FastBoot issue, as this previous property would make the modal be visible in FastBoot without having to rely on the `handleVisibilityChanges` method, which probably does not even run in FastBoot?

We might apply a similar initializer to the `state` property now (instead of setting to `closed` and again relying on `handleVisibilityChanges()`? 🤔

jelhan: Yes. That's also what I guess.

ember-render-helpers should be executed in FastBoot. Helpers are executed in FastBoot. And I haven't noticed any FastBoot logic in ember-render-helpers itself.

I suspect a FastBoot guard in the `show` method to be to strict and prevent needed work. But need to dig deeper into that one to verify.

addon/components/bs-modal.js
jelhan: You were right.

1. `getOwner(this).isDestroyed` was `true` in `next()` callback. Seems as if application gets destroyed before.
2. The next runloop had only hidden the assertion but not resolved the issue. Tracked property `state` was read in `show` and `hide` method before it was set.

Refactored to have `state` property not tracked as it must be read by `show` and `hide` modal. Otherwise we can not exit `show` method if the modal is already shown. Trying to show a modal with all animations, which is already shown, would lead to strange bugs.

Introduced another tracked property `isModalShown`, which is managed by `state` through a setter.

Actually this duplicated the state. But it seems that this is required by tracked properties if you must read the state before setting it.

jelhan:
I'm sorry for the many changes in one merge request. I may have refactored to much parts at once. Let me try to summarize to ease review:

1. Decoupled internal and external open state. The external `@open` state is observed using `{{did-insert}}` and `{{did-update}}` helpers. It triggers a `handleVisibilityChange` method. Beside that two helpers and the `handleVisibilityChange` method it is not read anywhere in the code.
2. Introduced a non-tracked property `state`, which reflects the internal modal state.
3. Introduced a tracked property `isModalShown`, which reflects the `state`. It is used in the template to control if modal is rendered or not.
4. Renamed `showBackdrop` property to `isBackdropShown`.
5. Split `handleBackdrop` in one method to show and one method to hide the backdrop.
6. Split `resize` method in one method to add event listeners and one method to remove event listeners.
7. Refactored callbacks to `async` / `await`.
8. Moved FastBoot checks as much as possible to the risky operations.

Tests are passing. Not sure if we feel safe enough to just trust our test suite. To be honest I have not tested it much manually.

addon/components/bs-modal.js
```diff
@@ -485,81 +515,101 @@ export default class Modal extends Component {
...
+
+  /**
+   * Attach resize event listeners
+   */
+  addResizeEventListeners() {
+    if (isFastBoot(this)) {
+      return;
+    }
+
+    window.addEventListener('resize', this.adjustDialog, false);
```
simonihmig, line 601: we might also use `{{on-window ...}}` from https://github.com/buschtoens/ember-on-helper to attach listener more declaratively in the template!?

addon/components/bs-modal.hbs
```diff
@@ -1,14 +1,15 @@
...
   {{#let (component (bs-default @dialogComponent (component "bs-modal/dialog"))
     onClose=this.close
     fade=this._fade
-    showModal=this.showModal
+    showModal=this.isModalShown
     keyboard=this.keyboard
     size=@size
     backdropClose=this.backdropClose
-    inDom=this.inDom
+    inDom=this.isModalShown
```
simonihmig, line 12: This seems suspicious, as `@showModal` and `@inDom` receive the same value, thus seem redundant. But I don't think they are. IIRC having those separate args might be related to having to put the modal into DOM first, and then add the `in` class to trigger the CSS transition. Doing it at the same time might prevent the transition to happen correctly. Which is likely not covered by tests.

simonihmig (commented):
Thanks for working on this!

This looks good, but I think we do need to test this well manually also, to make sure fade animations all work correctly, see the one related comment!

addon/components/bs-modal.hbs
jelhan: Good catch. In current implementation `show` method sets `inDom` to `true` immediately while `showModal` is set to `true` _after_ the backdrop is shown in `afterRender` runloop.

Hard to follow the existing code. But if I did not missed something it's simplified like this:

```js
async show() {
  this.inDom = true;
  
  await this.showBackdrop();

  schedule('afterRender', () => {
    this.showModal = true;
  });
}
```

Will try to manual check if I see a visual difference before and after. If so I will try to add test coverage for it.

jelhan: Looking at the `<BsModal::Dialog>` component the `@showModal` and `@inDom` arguments are consumed in https://github.com/kaliber5/ember-bootstrap/blob/6aeead2d17604afbdccc4295a4dffba621771239/addon/components/bs-modal/dialog.hbs#L5 and https://github.com/kaliber5/ember-bootstrap/blob/6aeead2d17604afbdccc4295a4dffba621771239/addon/components/bs-modal/dialog.hbs#L11.

I only consider Bootstrap 4 for now.

The `@showModal` argument controls if `.show` CSS class is applied. The `@inDom` argument controls if `.d-block` CSS class and `display: block` CSS is applied. So the question is `.show` CSS class and `display: block` should be applied at the same time or not.

Bootstrap applies both the `.show` CSS class and sets `display: block` in the same `_showElement` method: https://github.com/twbs/bootstrap/blob/20cfbdff79e19e77c24596cdca59c6694283e242/js/src/modal.js#L231-L254 So there doesn't seem to be timing difference for showing the modal.

But for hiding the modal the `.show` CSS class is removed _before_ the animation starts and `display: none` is set _after_ the animation is over. https://github.com/twbs/bootstrap/blob/20cfbdff79e19e77c24596cdca59c6694283e242/js/src/modal.js#L178 and https://github.com/twbs/bootstrap/blob/20cfbdff79e19e77c24596cdca59c6694283e242/js/src/modal.js#L316

Unless I missed something `@showModal` should be true for `opening` and `open` state, while `@inDom` should be true for `opening`, `open` and `closing` states.

Applying both the `.d-block` CSS class and `display: block` if `@inDom` is `true` seems to be redundant. See [live demo in Bootstrap documentation](https://getbootstrap.com/docs/4.6/components/modal/#live-demo) seem to have only `display: block` CSS property. But I guess for FastBoot support we need to apply the CSS class as well. At least as long as ember-style-modifier does not have support FastBoot yet.

jelhan: I don't get why `@inDom` argument is passed at all to `<BsModal::Dialog>`. Also in current implementation the component is teared down if `this.inDom` is `false`: https://github.com/kaliber5/ember-bootstrap/blob/6aeead2d17604afbdccc4295a4dffba621771239/addon/components/bs-modal.hbs#L3 It looks as if we are doing this:

```hbs
{{#if this.inDom}}
  <Foo @bar={{this.inDom}} />
{{/if}}
```

This seems to be the same as

```
{{#if this.inDom}}
  <Foo @bar={{true}} />
{{/if}}
```

because the code path is only triggered for `this.inDom === true`. :confused:

jelhan: Animation timing should now be the same as before. At least for `.modal` itself. Haven't added test coverage for backdrop timing yet.

addon/components/bs-modal.js
jelhan: Great suggestion. Done in 708adc7e97cffbc38d52fec70a4239fd2c5bbce7.

addon/components/bs-modal.hbs
simonihmig: > Bootstrap applies both the .show CSS class and sets display: block in the same _showElement method:

I think the difference here between plain BS and our components is that by having rich interactive components instead of static markup created by the user and some JS on top, we are able (and actually doing that) to render the modal's DOM only when needed, hence the `{{#if this.inDom}}`. But in this case we must make sure that first the DOM is created, and *after* that we add the `show` class, otherwise the CSS transitions (triggered by adding `show`) would not apply. That's why we also need these two different but similar states: `inDom` and `showModal`. Does that make sense?

> Applying both the .d-block CSS class and display: block if @inDom is true seems to be redundant. 

> I don't get why @inDom argument is passed at all to <BsModal::Dialog>

Seems you are right about these points. I can only explain it that these things were messed up during the course of refactorings, and indeed the state management of this component is a bit confusing.

addon/components/bs-modal.hbs
```diff
@@ -1,15 +1,24 @@
...
 {{did-update this.handleVisibilityChanges @open}}
 
-{{#if this.isModalShown}}
+{{!
+  ember-on-helper do not support FastBoot yet. This guard can be removed as
+  soon as ember-on-helper guards for FastBoot itself. It is tracked in this
+  GitHub issue: https://github.com/buschtoens/ember-on-helper/issues/3
+}}
+{{#unless this.isFastBoot}}
+  {{on-window "resize" this.adjustDialog}}
```
simonihmig, line 10: Very reasonable refactoring :+1:

addon/components/bs-modal.js
```diff
@@ -186,7 +186,16 @@ export default class Modal extends Component {
...
+  @tracked isInDom = this.state !== 'closed';
+
+  /**
+   * Controls if the modal is visible.
+   *
+   * It must be kept in sync with the non-tracked property `state`.
+   * To avoid getting out of sync it should be set only the setter
+   * of `state` property.
+   */
+  @tracked isShowModal = this.state === 'open' || this.state === 'opening';
```
simonihmig, line 198: Can't this be a simple (auto-tracked) getter instead? The only time we set this seems to be in `set state`, but there we also set it to the same or condition, so basically it seems like it is a completely derived state, or am I missing something?

tests/integration/components/bs-modal-test.js
```diff
@@ -246,4 +246,51 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+    await settled();
+    assert.dom('.modal').doesNotExist();
+  });
+
+  testBS4('it animates opening and closing the modal', async function (assert) {
+    this.set('open', false);
+    await render(hbs`<BsModal @open={{this.open}}>Hello world!</BsModal>`);
+
+    this.set('open', true);
+    await waitFor('.modal.show');
```
simonihmig, line 278: The closing part of the test looks good, but not sure about the opening: we basically have to assert that the `.modal` exists first, with `display: 'block'`, but *without* `.show`. And only later `.show` should be added (to trigger CSS transitions, see my other comment). This test would not detect a case where we render the modal with `.show` applied immediately (which would break CSS transitions), right?

I guess this is difficult to test though, as `.show` is added with only a minimal delay (`afterRender`), but `waitFor` is polling the DOM with a much bigger interval, so hard to catch that intermediate state where the modal is in DOM but doesn't have `.show`!? 🤔
(The closing part is easier, as `.show` is removed with a large delay (after the CSS transition))

simonihmig (commented):
Thanks again for working on this @jelhan, and sorry for the long delay here! It's just now that I have reserved some time again for maintenance on this project. 😔

This looks good, and I would like to merge this sooner than later, just one possible change to look into here (getter vs. tracked)...

tests/integration/components/bs-modal-test.js
jelhan: Not sure if I can fully follow. Let me try to rephrase with my own words:

To trigger the transition the step must have two phases:

1. An element exist and is attached to the DOM, which does _not_ have the `.show` class.
2. The `.show class is added to that element.

This is required due to the following code of bootstrap: https://github.com/twbs/bootstrap/blob/6ffb0b48e455430f8a5359ed689ad64c1143fac2/scss/_transitions.scss#L1-L7

I'm not sure if the current implementation takes that into account. I guess it would be best implemented using a modifier, which adds the `.show` class. A modifier is guaranteed to be executed _after_ the element is insert into the DOM. So timing-wise it would be a good fit and implementation shouldn't be too complex.

But testing sounds difficult indeed. We would need to execute an assert _after_ the element is insert into the DOM but _before_ the modifier runs. I'm not sure if that is possible at all.

simonihmig: I believe the current implementation (in this PR) is already correct in that regard, isn't it?
* state is set to `opening` (https://github.com/kaliber5/ember-bootstrap/pull/1402/files#diff-e071449aeb5436c622964ec59e95bbb45deb6552acca75046b8f29f479d263bbR465)
* on `afterRender` (when DOM is updated), state transitions to `open` (https://github.com/kaliber5/ember-bootstrap/pull/1402/files#diff-e071449aeb5436c622964ec59e95bbb45deb6552acca75046b8f29f479d263bbR488)

So this is just about testing. And given that you wanted to add tests asserting the correct animation behavior, it seems to me the opening part of it will also pass even with a broken behavior (i.e. inserting into DOM and applying `.show` happens in one step). 

If we can't reasonably test this for the above mentioned reasons, them I am ok with leaving the test as-is, just wanted to point this out.

jelhan: I played a little bit around with it as I'm not an expert with CSS animation. It seems as we not only need to wait until the element is attached to DOM. We also need to wait until browser has time to repaint. The animations have not run if I added the class immediately after attaching element to DOM. Only after delaying it for later with `setTimeout(fn, 0)` CSS animations are playing. Here is a JSFiddle showing it: https://jsfiddle.net/8ycm34df/

I _think_ we get similar timing if scheduling in Ember's `afterRender` runloop queue. But this makes testing even more challenging. Will try to ask in Ember Discuss as it seems to be an interesting one. :smile_cat:

jelhan: Maybe we can use [`transitionend` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event) to assert that a CSS transition has happened without relying on any implementation details at all. :thinking: Will try to investigate when I have a few minutes left.

PR #1406, opened 2021-01-20
<BsForm::Element> should respect @value argument

### LINKED ISSUE #1403: Form element @value argument is not rendered as initial value of form control
Upgrading from Ember Bootstrap 4.4.0 to 4.6.1, the `@value` argument on `<BsForm::Element>` is no longer set as the form control value when the control is initially rendered.

Looking at the code, in `4.4.0` the `_value` property of the component was decorated with `@defaultValue`.
https://github.com/kaliber5/ember-bootstrap/blob/19ded7030e6de572521595e026bdad7c955381ec/addon/components/bs-form/element.js#L296-L297

In `4.6.1` it is decorated with `@tracked`. However there is nothing that is setting the initial value of `_value`, so whatever you provide via `@value` is not rendered.
https://github.com/kaliber5/ember-bootstrap/blob/fdcd8ee53586b6611278c0607b4f68f9638ec683/addon/components/bs-form/element.js#L278

jelhan:
#1339 refactored `<BsForm::Element>` from `@ember/component` to `@glimmer/component`. For a `@ember/component` setting the `@value` argument executed the `value` setter of the component. But this is not the case anymore after changing to `@glimmer/component`. Therefore the component was not using the `@value` argument anymore at all.

The good thing is that keeping a local copy of the state it is not needed anymore at all. As argument and component property is decoupled by `@glimmer/component` we can handle all cases in the getter.

Thanks a lot to @lindyhopchris for providing the failing test.

Closes #1403 and #1404

### DIFF
--- a/addon/components/bs-form/element.js
+++ b/addon/components/bs-form/element.js
@@ -258,24 +258,17 @@ export default class FormElement extends FormGroup {
    * @public
    */
   get value() {
+    assert(
+      'You can not set both property and value on a form element',
+      isBlank(this.args.property) || isBlank(this.args.value)
+    );
+
     if (this.args.property && this.args.model) {
       return get(this.args.model, this.args.property);
     }
-    return this._value;
+
+    return this.args.value;
   }
-
-  set value(value) {
-    assert('You cannot set both property and value on a form element', isBlank(this.args.property));
-
-    this._value = value;
-  }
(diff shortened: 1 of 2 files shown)


simonihmig (approved):
Thanks for the quick fix!

Issue #1414, opened 2021-01-28
Bug with `showOwnValidation` updating more than once in same computation


lindyhopchris:
Seeing the following error since upgrading to `4.6` (currently on the latest version - i.e. `4.6.2`):

```
index.js:172 Uncaught (in promise) Error: Assertion Failed: You attempted to update `showOwnValidation` on `BsFormElementWithChangesetValidationsSupport`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

`showOwnValidation` was first used:

- While rendering:
  -top-level
    application

Stack trace for the update:
    at dirtyTagFor (validator.js:581)
    at setter (validator.js:867)
    at BsFormElementWithChangesetValidationsSupport.set [as showOwnValidation] (index.js:3959)
    at BsFormElementWithChangesetValidationsSupport.showValidationOnHandler (element.js:646)
    at BsFormElementWithChangesetValidationsSupport.showValidationOnHandler (element.js:68)
    at clear (runtime.js:144)
    at UpdatableBlockImpl.reset (runtime.js:1271)
```

After discussion on Discord with @jelhan, we've worked out that this is actually a bug in Ember Bootstrap rather than something I'm doing wrong in my app. Another user on Discord also reported seeing the same error.

I've managed to reproduce the error in the Ember Bootstrap repo using this test which reproduces the scenario in which I'm seeing a failing test in my application:

```js
test('it has a show own validation bug', async function (assert) {
  assert.expect(1);
  this.set('submit', ({ name }) => assert.equal(name, 'Zoey'));
  this.set('beforeSubmit', (model) => {
    if (this.user) {
      model.set('name', this.user.name);
    }
  });

  await render(hbs`
    <BsForm
      @model={{this}}
      @onBefore={{this.beforeSubmit}}
      @onSubmit={{this.submit}}
      as |Form|
    >
      {{#unless this.user}}
        <Form.element @property="name" />
      {{/unless}}

      <Form.submitButton @defaultText="Submit" />
    </BsForm>
  `);

  await fillIn('input', 'Tomster');

  this.set('user', { name: 'Zoey' });

  await click('button');
});
```

@jelhan has suggested that removing these two lines probably fixes the problem, so I'm going to have a go at a PR:
[...]

PR #1415, opened 2021-01-28
Fix for showOwnValidation throwing "same computation" Ember assertion

lindyhopchris:
Adds a test for #1414 that fails before the fix (also included in this PR) was implemented. The fix is to remove two lines from `bs-form/element.js` as suggested by @jelhan

### DIFF
--- a/tests/integration/components/bs-form-test.js
+++ b/tests/integration/components/bs-form-test.js
@@ -1147,4 +1147,38 @@ module('Integration | Component | bs-form', function (hooks) {
     );
     assert.dom('input').doesNotHaveAttribute('readonly');
   });
+
+  /**
+   * @see https://github.com/kaliber5/ember-bootstrap/issues/1414
+   */
+  test('it does not have a showOwnValidation bug', async function (assert) {
+    assert.expect(1);
+    this.set('submit', ({ name }) => assert.equal(name, 'Zoey'));
+    this.set('beforeSubmit', (model) => {
+      if (this.user) {
+        model.set('name', this.user.name);
+      }
+    });
+
+    await render(hbs`
+      <BsForm
+        @model={{this}}
+        @onBefore={{this.beforeSubmit}}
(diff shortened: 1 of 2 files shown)


jelhan (approved):
Thanks a lot for working on it. The test is a little bit different from the rest of the test suite. But I think we can refactor it later. @simonihmig Any concerns?

tests/integration/components/bs-form-test.js
```diff
@@ -1147,4 +1147,38 @@ module('Integration | Component | bs-form', function (hooks) {
...
+    });
+
+    await render(hbs`
+      <BsForm
+        @model={{this}}
+        @onBefore={{this.beforeSubmit}}
+        @onSubmit={{this.submit}}
+        as |Form|
+      >
+        {{#unless this.user}}
```
simonihmig, line 1170: Is removing the form element the essential thing here in this test, that triggers the exception?

simonihmig (approved):
Thank you @lindyhopchris!

>  The test is a little bit different from the rest of the test suite. But I think we can refactor it later.

Yeah, as you can see by my inline comment, I am not really sure what the root issue was. So without that understanding refactoring the test later is a bit difficult. 

Looking back at the discussion on Discord, seems it is somehow related to having the input still focused when submitting, do I read this right?

Anyway, given the change seems fine, we can merge this now.

simonihmig:
Released: https://github.com/kaliber5/ember-bootstrap/releases/tag/v4.6.3

tests/integration/components/bs-form-test.js
lindyhopchris: In my scenario it is the combination of using `fillIn` on the element before then removing it that triggered the error.

I had another instance (not in tests) where the same thing was happening on a login form - the error was triggered when a route transition was started as part of a successful submit. In that case, my best guess was that the form being ripped out of the DOM as part of the transition seemed to trigger it.

PR #1421, opened 2021-02-03
Do not use implicit this fallback in templates

jelhan:
Implicit `this` fallback triggers a deprecation in canary now. We still had many implicit `this` fallbacks in tests. I activated the template lint rule and fixed all of them manually.

I did not touched the documentation app. That one still has many implicit this fallbacks. I decided to skip that one as it does neither affect consumers application nor our tests.

This is not the only deprecation added in canary. Additionally I see this deprecation in CI log:

> Versions of modifier manager capabilities prior to 3.22 have been deprecated. You must update to the 3.22 capabilities.

As it's not related it is better addressed by a separate pull request in my opinion. It needs to be fixed upstream first: https://github.com/ember-modifier/ember-modifier/pull/63

### DIFF
--- a/.template-lintrc.js
+++ b/.template-lintrc.js
@@ -5,6 +5,15 @@ module.exports = {
   rules: {
     'no-html-comments': false,
     'attribute-indentation': false,
+    'no-implicit-this': true,
   },
   ignore: ['fastboot-tests/fixtures/**/*'],
+  overrides: [
+    {
+      files: ['docs/**/*'],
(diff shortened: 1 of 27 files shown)


jelhan:
@simonihmig I _think_ I haven't broken something and failing tests are only related to #1440. But please give it a sanity check.

simonihmig (approved):
Thanks a lot @jelhan! 

Seems this was quite some work!

Issue #1440, opened 2021-02-26
Failing builds in CI


simonihmig:
Currently CI is failing, due to two issues it seems in Ember:
* ~https://github.com/emberjs/ember.js/issues/19408~
* ~https://github.com/emberjs/ember.js/issues/19431~
* ~https://github.com/embroider-build/embroider/issues/701~
* Failing tests due to raised deprecations
  * ~deprecations in canary due to missing `this.` in some templates, makes the tests fail~
  * ~Expected no deprecations, found: "Versions of modifier manager capabilities prior to 3.22 have been deprecated. You must update to the 3.22 capabilities."~
  * Invoking the `<LinkTo>` component with positional arguments is deprecated. WIP: #1455 
  * `Usage of the Ember Global is deprecated.`

Posting here for visibility

rwjblue:
https://github.com/emberjs/ember.js/issues/19408 should have been fixed on canary already in https://github.com/emberjs/ember.js/pull/19395, please confirm?


https://github.com/emberjs/ember.js/issues/19431 is fixed by https://github.com/ember-cli/ember-cli-htmlbars/releases/tag/v5.6.0.

simonihmig:
> emberjs/ember.js#19408 should have been fixed on canary already in emberjs/ember.js#19395, please confirm?

@rwjblue yes, does not happen in canary anymore, but still see it in ember-release.

rwjblue:
https://github.com/emberjs/ember.js/issues/19408 should be fixed (released in 3.24.3 and 3.25.3).

simonihmig:
Closed by #1477

PR #1455, opened 2021-03-14
Refactor Nav/Navbar/Dropdown's LinkTo components to not extend from Ember's LinkComponent

simonihmig:
According to RFC671 and deprecations introduced in latest Ember Canary.

Should fix the last remaining deprecation, see #1440.


addon/components/bs-link-to.js
```diff
@@ -0,0 +1,113 @@
...
+
+ @class LinkComponent
+ @namespace Components
+ @extends Component
+ @private
+*/
+@tagName('')
+class LinkComponent extends Component {
+  // We still need to use the private service, so this component can react to changes of `currentState`, e.g. changing
+  // the model while the route in unchanged.
```
simonihmig, line 22: @jelhan do you happen to know if there is another way to support this, using only public API?

addon/components/bs-link-to.js
```diff
@@ -0,0 +1,113 @@
...
+    if (typeof currentWhen === 'boolean') {
+      return currentWhen;
+    }
+
+    let { _models: models, _routing: routing } = this;
+
+    if (typeof currentWhen === 'string') {
+      return currentWhen
+        .split(' ')
+        .some((route) => routing.isActiveForRoute(models, undefined, this._namespaceRoute(route), routerState));
```
simonihmig, line 64: Ah, shit. I haven't copied `this._namespaceRoute()` from the original LinkComponent, which seems to be needed for the `current-when`. This would require even more stupid copy-pasting.

I wonder if we need to support this: `current-when` is nowhere mentioned in our API docs, it's just the fact that we documented the component to extend from `Ember.LinkComponent` (shouldn't have done this probably!?). So can we drop this nasty stuff without making this a breaking change? 🤔

addon/components/bs-link-to.js
```diff
@@ -0,0 +1,113 @@
...
+  // eslint-disable-next-line ember/no-component-lifecycle-hooks
+  didReceiveAttrs() {
+    super.didReceiveAttrs(...arguments);
+
+    let { params } = this;
+    if (!params || params.length === 0) {
+      return;
+    }
+
+    params = params.slice();
```
simonihmig, line 79: We haven't  formally deprecated positional params for link params it seems. While this is now deprecated for Ember's original LinkComponent, we do not trigger this deprecation anymore. So we should deprecate this explicitly by ourselves I think. Should be a follow up PR.

simonihmig:
The deprecation in Canary is gone, but something else related to FastBoot is failing now... 🤔

jelhan (commented):
Thanks a lot for picking this one up!

Was wondering why you decided to have a base class, which is than extended. Wouldn't implementation be easier to follow and require less code duplication, if `<BsLinkTo>` is invoked in templates instead of extending from it?

If we can avoid much of the complexity by dropping support for position params, `currentWhen` and `activeClass` arguments and may accept some differences in edge cases I would be fully onboard. I'm a little bit afraid of the maintenance burden if trying to exactly replicate current functionality.

addon/components/bs-link-to.js
simonihmig: > Haven't we dropped support for curly-bracket component invocation in v4

We haven't declared so, no. We dropped arguments for HTML attributes, and more or less recommended to use angle brackets, but haven't explicitly stopped support for curly invocation. At the time there was no reason for this, and this is still not deprecated from Ember's side AFAIK. Except for this positional arguments issue there is no reason why curly invocation should bother us, as long as Ember supports it we can basically not care about it. 

We can choose to drop support in a next major, or just say that the use of our custom link component drops support for positional arguments and all the other things (currentWhen) that accidentally became part of our implementation due to extending from LinkComponent, whether or not we consider this a public API or not. (it's a "gray area" I guess)

IIRC you were more on the side of accepting major versions more often, right? So should we actually do this refactoring as part of a v5, with just this edge case-y thing as the (potentially, but rather rare) breaking change?

simonihmig:
> Was wondering why you decided to have a base class, which is than extended. Wouldn't implementation be easier to follow and require less code duplication, if <BsLinkTo> is invoked in templates instead of extending from it?

I just tried that, which made me remember why: doesn't work for positional arguments. However if we want to drop them anyway - as part of a breaking change or not, see my other comment - then this makes sense!

simonihmig:
> I just tried that, which made me remember why: doesn't work for positional arguments. However if we want to drop them anyway - as part of a breaking change or not, see my other comment - then this makes sense!

I refactored this more, this is now yielding the BsLinkTo component directly, without extending it or even having to use a template-only wrapper. This still allows positional arguments, so no need for a semi-breaking change yet...

Will still look if I can remove the old cruft from the BsLinkTo component, i.e. just use the public `router.isActive()`...

simonihmig:
> Will still look if I can remove the old cruft from the BsLinkTo component, i.e. just use the public router.isActive()...

I have done this now as much as possible...
* Our link component only uses public API of the router service, which is nice.
* I removed that `current-when` and other stuff implicitly inherited from Ember's LinkComponent, as indeed we never indicated that was supported
* However we did have tests for positional arguments! 😕  So removing those kinda feels like introducing a breaking change. I therefore left support, but will add a deprecation for those. 
* The link component is still an `Ember.Component` as it uses the `ComponentChild` mixin. We *could* get rid of the mixin itself, but still require the use of `Ember.Component` as we have to [find the "parent" component](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/mixins/component-child.js#L21) (e.g. a nav item), for this annoying entangling of `<a>` having knowledge or the routing params and `<li>` needing to set the `.active` class. (Btw this is not the case anymore in BS5 it seems! :tada:)
* This could be solved much easier if the navItem would yield the navLinkto component as it could then simply curry a `parent=this` argument. But unfortunately that's not our API, we yield the navLinkto from the nav itself 😔. I think we should change that (and deprecate the other way), right? Or accept it like it is today, and hope for people to switch to BS5 sooner than later? 🤔

simonihmig:
Finally CI is passing. @jelhan want to have another look?

addon/components/bs-navbar/link-to.hbs
```diff
@@ -0,0 +1,14 @@
+<BsLinkTo
+  @route={{this.route}}
+  @model={{this.model}}
+  @models={{this.models}}
+  @query={{this.query}}
+  @replace={{@replace}}
+  @disabled={{@disabled}}
+  @current-when={{@current-when}}
+  @activeClass={{@activeClass}}
```
jelhan, line 9: I don't think `<BsLinkTo>` supports `@replace`, `@current-when` and `@activeClass`. I guess a left-over from an earlier iteration.

jelhan (approved):
Looks very good. Thanks a lot for working on this.

addon/components/bs-navbar/link-to.hbs
simonihmig: Good catch, will fix!

Issue #912, opened 2019-10-24
Support custom form controls checkbox, radio and switch


jelhan:
Bootstrap 4 supports advanced checkboxes and radio controls. First one could also be styled as a switch. Ember Bootstrap should support them as well. If I didn't missed something it's only requires different CSS classes.

Bootstrap docs:
- Checkboxes and radios: https://getbootstrap.com/docs/4.3/components/forms/#checkboxes-and-radios-1
- Switches: https://getbootstrap.com/docs/4.3/components/forms/#switches

adambedford:
I'd love to see this!

jelhan:
@adambedford Let me know if you have time to work on this one. Happy to help you with any question.

lindyhopchris:
@jelhan hope you're ok! was wondering what needs to happen to get this progressed? I really need to use the switches, so I'd be prepared to do a PR. I just need some guidance as to the expectation for how this should be implemented, so that I don't waste any of my limited time on the PR?

Using switches as an example, is the intent to add a new control type? Or should "checkbox" still be used but some other argument control the custom control type?

jelhan:
Thanks a lot for taking this one up.

> Using switches as an example, is the intent to add a new control type? Or should "checkbox" still be used but some other argument control the custom control type?

I think we have two options for public API:

```
<BsForm as |form|>
  <form.element @controlType="switch" />
</BsForm>
````

and

````
<BsForm as |form|>
  <form.element @controlType="checkbox" as |element|>
    <element.control @style="switch" />
  </form.element>
</BsForm>
```

I haven't evaluated if both styles are supported by current architecture. It depends on the HTML markup expected by Bootstrap. But I hope both are.

To evaluate if you could achieve (and to have a quick work-a-round) you could try to construct HTML markup expected by Bootstrap inside of block yielded by `<form.element>`. If this isn't possible, we need to talk. Otherwise it could be implemented as a flavor of `checkbox` control.

If possible I would tend towards controlling by an argument of `checkbox` control. Switch and checkbox seems to be a visual difference only. It's not another input concept. But I'm not very opinionated about it.

lindyhopchris:
So the problem is the wrapping `<div>`

A standard checkbox wraps the input in a `<div class="form-check">`.

The custom checkbox wraps the input in `<div class="custom-control custom-checkbox">` and the switch uses `<div class="custom-control custom-switch">`

[...]

Issue #1387, opened 2021-01-07
Docs aren't published since v4.4.0


jelhan:
The docs hosted at https://www.ember-bootstrap.com/ haven't been published anymore since v4.4.0:

```js
// developers tools when running https://www.ember-bootstrap.com/
JSON.parse(
  decodeURIComponent(
    document.querySelector('meta[name$="config/environment"]').content
  )
).APP.version // "v4.4.0+19ded703"
```

The build is timing out in CI:

- https://github.com/kaliber5/ember-bootstrap/runs/1493307110 (v4.5.0)
- https://github.com/kaliber5/ember-bootstrap/runs/1627376802 (v4.6.0)
- https://github.com/kaliber5/ember-bootstrap/runs/1664220607 (v4.6.0 restarted just know)

I tried to run `yarn run docs:build` locally but also got stucked with the same output.

simonihmig:
Thanks for reporting, will try to look into this...

jelhan:
Should be fixed by #1514. But I would propose to keep this one open until it has been deployed successfully.

simonihmig:
Just released `v4.7.0`, and docs were successfully deployed. So closing this.

Unfortunately there is a regression, also visible in the docs, will try to fix this asap: #1518.

PR #1495, opened 2021-05-12
Support switch control in forms (BS4 only)

lindyhopchris:
Adds a `switch` control type, for the Bootstrap 4 switch custom control type:
https://getbootstrap.com/docs/4.6/components/forms/#switches

See #912


lindyhopchris:
FYI I couldn't preview the changes to the docs site as it's not working when I run `ember s` locally in the docs folder - getting an error.

tests/integration/components/bs-form/element/control/switch-test.js
```diff
@@ -0,0 +1,16 @@
...
+import { testBS4 } from '../../../../../helpers/bootstrap';
+import hbs from 'htmlbars-inline-precompile';
+import setupNoDeprecations from '../../../../../helpers/setup-no-deprecations';
+
+module('Integration | Component | bs form/element/control/switch', function (hooks) {
+  setupRenderingTest(hooks);
+  setupNoDeprecations(hooks);
+
+  testBS4('it renders', async function (assert) {
+    await render(hbs`<BsForm::Element::Control::Switch />`);
```
jelhan, line 13: As above I would prefer to only use public API in tests:

```suggestion
    await render(hbs`
      <BsForm as |form|>
        <form.element @controlType="switch" />
      </BsForm>
    `);
```

tests/integration/components/bs-form/element/label-test.js
```diff
@@ -42,6 +42,14 @@ module('Integration | Component | bs form/element/label', function (hooks) {
     assert.dom('label').hasClass('form-check-label', 'component has form-check-label class');
   });
 
+  testBS4('component has custom-control-label class when using control type switch', async function (assert) {
+    await render(hbs`<BsForm::Element::Label @controlType="switch" />`);
```
jelhan, line 46: As above I would prefer to only use public API in tests:

```suggestion
    await render(hbs`
      <BsForm as |form|>
        <form.element @controlType="switch" />
      </BsForm>
    `);
```

tests/integration/components/bs-form/element-test.js
```diff
@@ -148,6 +148,22 @@ module('Integration | Component | bs-form/element', function (hooks) {
     });
   });
 
+  testBS4('controlType "switch" is supported', async function (assert) {
+    await render(hbs`<BsForm::Element @controlType="switch" />`);
```
jelhan, line 152: I know that this style is used at many existing tests. But would prefer to invoke it using public API only. Do you mind changing?

```suggestion
    await render(hbs`
      <BsForm as |form|>
        <form.element @controlType="switch" />
      </BsForm>
    `);
```

jelhan (commented):
Thanks a lot for working on this. Haven't had the time yet to do a detailed review. But what I have seen so far looks very good to me. Will try to finish review later today.

> FYI I couldn't preview the changes to the docs site as it's not working when I run `ember s` locally in the docs folder - getting an error.

The docs site is broken since quiet some time. :disappointed: See #1387 for details.

addon/components/bs-form/element/control/switch.hbs
```diff
@@ -0,0 +1,11 @@
...
+  type="checkbox"
+  id={{@id}}
+  disabled={{this.disabled}}
+  readonly={{this.readonly}}
+  aria-describedby={{this.ariaDescribedBy}}
+  checked={{this.value}}
+  class="custom-control-input {{this.formValidationClass}}"
+  ...attributes
+  {{on "click" this.handleClick}}
+>
```
jelhan, line 11: Wondering if it makes sense to refactor to:

```hbs
<BsForm::Element::Control::Checkbox
  @value={{@value}}
  @id={{@id}}
  @type="checkbox"
  @label={{@label}}
  @disabled={{@disabled}}
  @readonly={{@readonly}}
  @required={{@required}}
  @options={{@options}}
  @optionLabelPath={{@optionLabelPath}}
  @ariaDescribedBy={{@ariaDescribedBy}}
  @onChange={{@onChange}}
  @validationType={{@validationType}}
  @size={{@size}}
  class="custom-control-input"
  ...attributes
/>
```

We would avoid some code duplication. But needing to explicitly pass down all arguments is a big trade-off. We could only passthrough the arguments actually used by `<BsForm::Element::Control::Checkbox />`. But that would be also error prune...

So not a request to change. More a question on long-term architectural patterns for @simonihmig. :smile_cat:

jelhan (approved):
Looks good to me. Thanks a lot for working on this one!

addon/components/bs-form/element/control/switch.hbs
lindyhopchris: The problem is that the checkbox control adds a `form-check-input` class, which we do not want on the switch checkbox. Which is why I went with duplicating the control.

jelhan: Missed that one. Thanks for explaining.

lindyhopchris:
@jelhan thanks for reviewing. I'll make those changes to the test when I get a chance (hopefully later today)

addon/components/bs-form/element/control/switch.js
```diff
@@ -0,0 +1,20 @@
...
+import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
+
+/**
+
+ @class FormElementControlSwitch
+ @namespace Components
+ @extends Components.FormElementControl
+ @private
+ */
+@tagName('')
```
simonihmig, line 13: Hm, reminds me that we still have quite some work left over to move things to Glimmer components. It's ok to do this here as anything else would be out of scope for this PR, but still a bit unfortunate that we basically add additional tech debt here. Again, not your fault!

addon/components/bs-form/element/layout/horizontal/checkbox.hbs
```diff
@@ -8,11 +8,11 @@
     {{component @errorsComponent}}
     {{component @helpTextComponent}}
   {{else}}
-    <div class="form-check">
+    <div class="{{if this.isSwitch "custom-control custom-switch" "form-check"}}">
```
simonihmig, line 11: I would suggest to express the same without introducing a CP in the backing class, by using the `bs-eq` helper here. Would ease the eventual transition to a template-only component later. Same for the other components.

In general we started to move more of the very basic derived state like this into the template, to reduce the need to add trivial CPs/getters to the class, again a.o. with the goal to use template-only components wherever possible.

tests/integration/components/bs-form/element-test.js
simonihmig: I agree. Again not your fault, as examples like this do exist in the existing code base. Which I would consider bad practice nowadays, and should get refactored over time (e.g. in #1455)

simonihmig (commented):
Thanks a lot for working on this!

lindyhopchris:
@simonihmig thanks for your comments! All make sense, so I'll be able to push an update. Today is looking busy so hopefully I'll have time tomorrow.

lindyhopchris:
@jelhan @simonihmig

I've pushed the changes based on the review comments, so this should now be ready to merge. Do you need me to fix the merge conflicts? I tried back merging `master` into my `switch` branch but then all the tests started to fail.

simonihmig:
> Do you need me to fix the merge conflicts? 

Yes, unfortunately I cannot merge it with this conflict. I did a few things here yesterday, so sorry for causing this additional work! 😔

> I tried back merging master into my switch branch but then all the tests started to fail.

Hm, what exactly was the error? Did you update your dependencies and restarted `ember serve`?

lindyhopchris:
@simonihmig have pushed the merge fix. Turned out the failing tests were due to a bad merge on my part, re-did the merge and worked out how to fix it, so all should be good now.

simonihmig (approved):
Thanks a lot again!

PR #1496, opened 2021-05-13
Support Bootstrap v5 in addon index.js

simonihmig:
This is just the minimal change in index.js to support v5. More related PRs will follow.


simonihmig:
Failing canary tests due to #1498, so going to merge this.

PR #1497, opened 2021-05-13
Support Bootstrap5 in default blueprint

simonihmig:
Will update the setup guide later when the docs app is fixed...


jelhan (approved):
Looks good to me. Thanks a lot for pushing Bootstrap 5 support.

Do you know why the files haven't been prettified before?

simonihmig:
> Do you know why the files haven't been prettified before?

The folders had a local `.eslintrc.js`, which seems to have disabled the prettier linting effectively. So that's why the "wrongly" formatted files never caused linting errors. I removed those config files in this PR.

Issue #1494, opened 2021-05-11
Support Bootstrap 5


simonihmig:
Trying to summarize what the minimum changes are to support Bootstrap 5.

- [x] Support v5 in the addon's index.js. PR: #1496 
- [x] Add support for `--bootstrap-version=5` in the default blueprint (with v4 still being the default for now) PR: #1497
- [x] Update [setup docs](https://www.ember-bootstrap.com/getting-started/setup) #1617 
- [x] Setup CI for BS5. PR: #1499 
- [x] Deprecate BS3. Only loosely related to BS5 support, but I think we don't want to support all three versions for an extended time. But to introduce BS5 support without a major bump, we might have to do it for some time. As there are not many renamed classes, this seems doable. PR: #1536 
- [x] Update markup for Accordion, see https://getbootstrap.com/docs/5.0/components/accordion/ PR: #1561 
- [x] Update markup for Forms, see https://getbootstrap.com/docs/5.0/migration/#forms. PR: #1512
- [x] Update markup for Buttons, see https://getbootstrap.com/docs/5.0/migration/#buttons PR: #1537 
- [x] Update toggle buttons for ButtonGroups, see https://getbootstrap.com/docs/5.0/migration/#buttons PR: #1540 
- [x] Update markup for Navbars, see https://getbootstrap.com/docs/5.0/migration/#navbars PR: #1597
- [x] Update markup for Popovers, see https://getbootstrap.com/docs/5.0/migration/#popovers. PR: #1502
- [x] Update markup for Tooltips, see https://getbootstrap.com/docs/5.0/migration/#tooltips. PR: #1502
- [x] BS4 uses popper.js v2. Do we have to follow here? `ember-popper` hasn't been maintained much, https://github.com/alexlafroscia/ember-popper-modifier seems like a more modern and up to date solution, and uses v2.
- [x] BS5 drops support for IE11 :tada:. While this doesn't really affect us technically, it at least makes dropping support in our addon even more desirable. ~Probably~ in *our* v5, ~maybe including requiring Ember 4.0 as the minimum version then?~ #1596 
- [x] Change use of `.sr-only` to `.visually-hidden`
- [x] Fix modal close button, see [this comment](https://github.com/kaliber5/ember-bootstrap/issues/1494#issuecomment-917175019) #1614 

Any kind of contributions (PRs, documentation, testing) highly welcome! Just comment here if you are interested in helping out!

/cc @jelhan

vlascik:
[...]

Issue #1498, opened 2021-05-13
Failing CI for Ember canary


simonihmig:
Some change in GlimmerVM that was added to Ember canary just a day or two ago is causing [this test](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-form-test.js#L410-L426) (added by #877) to fail. 

This is the PR in Glimmer: https://github.com/glimmerjs/glimmer-vm/pull/1298.

> For backwards compatibility, we
also allow users to set values on the proxy, so long as they are not
setting any of the values defined on the proxy initially, since those
are read-only references.

Our form does set `name` on `{{hash name=...}}`, which is causing this assertion. Not really understanding the reasoning behind this though!? Feels like a breaking change to me... 🤔

/cc @jelhan

PR #1499, opened 2021-05-13
Add experimental/incomplete Bootstrap 5 support to existing components

simonihmig:
This change *pretends* that BS5 components require the same markup as BS4. This should be true for many components, but not all. Follow up PRs will be created to address the latter cases, see #1494. 

But at least this is better than what we had before, as the macros checked for only 3 or 4 and would leave the markup for BS5 broken basically everywhere. And this allows us to iterate step by step on fixing the remaining issues, with BS5 already enjoying full CI coverage.

@jelhan This was mostly a find-and-replace job, so no line by line review required. But would like to hear if you approve this approach here!


jelhan:
I like the approach. Wondering if we should name it `isNotBS3`. This would highlight that we could remove when dropping support for Bootstrap 3. And might be easier to read than `isBS45`. But not having strong opinion about.

simonihmig:
> Wondering if we should name it isNotBS3

I like, updated it!

CI seems to be green now finally (except for #1498)

PR #1500, opened 2021-05-13
Re-add broken fade arg to BsAlert component

gfmartinez:
Looks like [this PR](https://github.com/kaliber5/ember-bootstrap/pull/1224/files#diff-70e754d719d2e6b45827cb320ec7224ae80cae42862141739894b6be65c1f24bL99) removed the `fade` property from the bs-alert component. In the current version, whatever you set `fade` to, it will always be false and the 'fade' class is never added.

The alt approach to fix this is to just change `this.fade` to `this.args.fade` in the template, but I went this way because of the documented default of true that is still in the js file.

### DIFF
--- a/addon/components/bs-alert.js
+++ b/addon/components/bs-alert.js
@@ -104,6 +104,8 @@ export default class Alert extends Component {
    * @default true
    * @public
    */
+  @arg
+  fade = true;
 
   get showAlert() {
     return this._visible && this.args.fade !== false;


simonihmig (commented):
Thanks for your contribution!

Seems we had a hole in our test coverage. Would you mind adding an assertion for `.fade` in [this test](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-alert-test.js#L19-L25), and another similar test for the `@fade={{false}}` case?

simonihmig:
@gfmartinez I merged this as-is now, as I want to cut a release soon, and wanted to have this fix in, even without a test. Should you still find the time to add a test in a follow-up PR, this would be highly appreciated! Thanks again!

gfmartinez:
@simonihmig Thanks! Work just got crazy this week. Will aim to followup asap.

PR #1502, opened 2021-05-13
Update Tooltips and Popovers for Bootstrap 5

simonihmig:


simonihmig:
This is pretty weird. The arrow positioning tests fail here for BS5, although they do pass locally (both `ember serve` and `ember test`).

Even more weird, is that these tests do pass in CI for Ember release scenario, but fail for all others. I can't even see how the Ember version can cause this. 

And basically impossible to debug, when not reproducible locally. At least a screenshot of the test container would be helpful... 😩

simonihmig:
Hm, ember-release passes because it runs with Firefox. The CI setup is not quite right here, we don't run ember-release in Chrome...

simonihmig:
Finally CI is passing, by making thee target element have a fixed width. Not sure what was making the tests fail only in CI, but as these were some of the very few tests that assert the actual styling and layout, it could be for example some different font rendering under Linux that would cause the target element (a button) to have (slightly) different dimensions affecting the positioning of the actual popover/tooltip.

PR #1507, opened 2021-05-17
Update to Ember 3.26 blueprint

simonihmig:
Needed a quite a fix test/linting fixes due to updates of ember-qunit and ember-template-lint


addon/components/bs-form.hbs
```diff
@@ -1,6 +1,5 @@
 {{!-- template-lint-disable no-invalid-interactive --}}
 <form
-  role="form"
```
jelhan, line 3: Wondering why we had it in the first place. Don't find it in Bootstrap v3 or v4 docs.

tests/integration/components/bs-tooltip-test.js
```diff
@@ -496,8 +496,8 @@ module('Integration | Component | bs-tooltip', function (hooks) {
 
   test('can be shown and disposed in same loop', async function (assert) {
     this.set('show', true);
-    await render(hbs`{{#if this.show}}<div id="target">{{bs-tooltip title="Dummy" class="wide"}}</div>{{/if}}`);
-    triggerEvent('#target', 'mouseenter');
+    await render(hbs`{{#if this.show}}<div id="target"><BsTooltip @title="Dummy" class="wide"/></div>{{/if}}`);
+    await triggerEvent('#target', 'mouseenter');
```
jelhan, line 500: I'm not sure if adding `await` is correct here. Test description discusses _same loop_. Awaiting the application to get into settle state again would not _show and disponse in same loop_.

jelhan (changes requested):
Thanks a lot for working on this one!

tests/integration/components/bs-tooltip-test.js
simonihmig: Yeah, I had the same concern actually. But something about the implementation of the test helpers must have changed. This test was failing because it tried to trigger the event *after* the component was already destroyed and removed from DOM! So awaiting for it before destroying the component was the only way to enforce that order. 

I did not have much time to spend on this. Any idea?

jelhan: Very confusing. The `await settled()` a few lines later should prevent the test from finishing. Is this maybe a bug in upstream library?

jelhan: Maybe as a quick fix we could skip the test? Seems better to me than changing it so that it may not capture the bug case anymore.

simonihmig: > The await settled()

Yes, but the issue is probably not the test finishing, but through setting `show` to false the component is removed, and that happens before the (async) `triggerEvent` has done it's work!? 🤔

> Maybe as a quick fix we could skip the test? Seems better to me than changing it so that it may not capture the bug case anymore.

Makes sense. Will do that!

PR #1513, opened 2021-05-18
forward arguments in <BsNavbar::LinkTo> to <BsLinkTo>

jelhan:
Due to this bug `<BsNavbar>` was basically unusable. It was introduced in #1455, which has not been released yet. I have only noticed accidentally when trying to fix the docs app.

The bug uncovers a lack in our test coverage. I think we could cover at least some cases by tests if asserting for `href` attribute in an integration test or by using an acceptance test. Will try to investigate that one later.

I think we should add tests before merging. But feel free to merge without tests if it blocks a release otherwise.

### DIFF
--- a/addon/components/bs-navbar/link-to.hbs
+++ b/addon/components/bs-navbar/link-to.hbs
@@ -1,11 +1,11 @@
 <BsLinkTo
-  @route={{this.route}}
-  @model={{this.model}}
-  @models={{this.models}}
-  @query={{this.query}}
+  @route={{@route}}
+  @model={{@model}}
+  @models={{@models}}
+  @query={{@query}}
   @disabled={{@disabled}}
-  {{on "click" this.onClick}}
   ...attributes
+  {{on 'click' this.onClick}}
 >
   {{yield}}
 </BsLinkTo>
--- a/addon/components/bs-navbar/link-to.hbs
+++ b/addon/components/bs-navbar/link-to.hbs
@@ -4,8 +4,8 @@
   @models={{@models}}
   @query={{@query}}
   @disabled={{@disabled}}
-  ...attributes
   {{on 'click' this.onClick}}
+  ...attributes
 >
   {{yield}}
 </BsLinkTo>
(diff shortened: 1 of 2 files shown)


jelhan:
Tests are failing for Ember 3.16 and 3.20 with the following error:

```
not ok 396 Chrome 90.0 - [85 ms] - Integration | Component | bs-navbar: Clicking expanded navbar link collapses navbar
    ---
        actual: >
            [object Object]
        stack: >
            TypeError: Cannot read property 'isActiveIntent' of undefined
                at RouterService.isActive (http://localhost:7357/assets/vendor.js:20245:27)
                at LinkComponent.get (http://localhost:7357/assets/vendor.js:107322:54)
                at http://localhost:7357/assets/vendor.js:39226:29
                at runInAutotrackingTransaction (http://localhost:7357/assets/vendor.js:56788:9)
                at track (http://localhost:7357/assets/vendor.js:57465:9)
                at LinkComponent.desc.get (http://localhost:7357/assets/vendor.js:39225:40)
                at getPossibleMandatoryProxyValue (http://localhost:7357/assets/vendor.js:15465:19)
                at _getProp (http://localhost:7357/assets/vendor.js:15530:17)
                at _get3 (http://localhost:7357/assets/vendor.js:15516:55)
                at wrapperClass.callback (http://localhost:7357/assets/vendor.js:41553:31)
        message: >
            Cannot read property 'isActiveIntent' of undefined
        negative: >
            false
        browser log: |
    ...
```

Seems to be an issue with router service and integration tests. It maybe related to [this bug fix](https://github.com/emberjs/ember.js/pull/19080), which landed in Ember 3.24.

simonihmig:
> Seems to be an issue with router service and integration tests. It maybe related to this bug fix, which landed in Ember 3.24.

I wanted to point out that I had to [do this](https://github.com/kaliber5/ember-bootstrap/pull/1455/files#diff-eedec3a5a2f9cea74fde1c1c4aefe3d038867574d61868d612409079763aa56dR50) in my PR, only to find out you already figured it out! :+1:

simonihmig (approved):
> The bug uncovers a lack in our test coverage. I think we could cover at least some cases by tests if asserting for href attribute in an integration test or by using an acceptance test. Will try to investigate that one later.

Yeah. For navs we have this test: https://github.com/kaliber5/ember-bootstrap/blob/master/tests/acceptance/bs-nav-link-test.js
(also I just saw that it only tests `.active` handling, not `href`. Maybe we should add thishere also?)

Something like this for navbars maybe?

simonihmig:
Merged now without a test. A follow-up PR adding the test would be great! :wink:

PR #1514, opened 2021-05-18
Fix docs app by upgrading dependencies

jelhan:
This includes:

- Upgrade ember-source and ember-cli
- Upgrade blueprints to 3.26.1
- Upgrade ember-bootstrap-power-select and ember-power-select
- Upgrade ember-bootstrap-cp-validations

And many more.

To not require too much refactoring at once I changed some ESLint and template lint from `error` to `warn`. We should address them later.

It also prettifies the JavaScript files as Prettier is enabled by default in recent blueprints. That causes a lot of churn on the merge request. I hope review is not too hard anyways.

I think we should add basic test coverage for docs app to not end up in the same situation again. A simple check that the app could be build would have caught that problem early. But I think we can do so in a follow-up PR to not overload this one.


simonihmig (approved):
That's awesome. I had this on my to do list, but really glad you did it first! 😆
Thanks so much!

> It also prettifies the JavaScript files as Prettier is enabled by default in recent blueprints. That causes a lot of churn on the merge request. I hope review is not too hard anyways.

No problem. Tbh I didn't review this line by line, as there don't seem to be any controversial changes, and as I trust you any way! :wink:

> I think we should add basic test coverage for docs app to not end up in the same situation again. A simple check that the app could be build would have caught that problem early. But I think we can do so in a follow-up PR to not overload this one.

Makes sense! We could/should also configure dependabot to update the docs app also! (hopefully not causing separate PRs for the same dependencies! 🤔  In a *real* monorepo this works quite fine)

simonihmig:
Before the next release, I'll play with this locally. But merging this already now.

PR #1341, opened 2020-11-22
WIP: Fix optimized Embroider build

simonihmig:
One of the recent versions introduced more strict checks for dynamic component patterns. We no have to wrap with `ensure-safe-component` to make sure dynamic components are not passed as strings.

Blocked on:
* https://github.com/embroider-build/embroider/pull/624
* https://github.com/tildeio/ember-element-helper/pull/34

PR #1515, opened 2021-05-18
Allow template only imports to resolve correctly

thoov:
While adding `@embroider/test-setup` to ember-cp-validation ran into an issue: 

```
Build Error (PackagerRunner) in components/bs-modal-simple.js

Module not found: Error: Can't resolve '../../../node_modules/ember-bootstrap/components/bs-modal-simple' in '$TMPDIR/embroider/02a8a6/tests/dummy/components/bs-modal-simple.js'
```

This PR connects the components inside of app so that the addon namespace template only components can resolve correctly.

CC: @rwjblue

### DIFF
--- a/addon/components/bs-modal-simple.js
+++ b/addon/components/bs-modal-simple.js
@@ -277,5 +277,5 @@
  * @default null
  * @public
  */
- import templateOnly from '@ember/component/template-only';
- export default templateOnly();
+import templateOnly from '@ember/component/template-only';
+export default templateOnly();
--- a/addon/components/bs-modal-simple.js
+++ b/addon/components/bs-modal-simple.js
@@ -277,3 +277,5 @@
  * @default null
  * @public
  */
+ import templateOnly from '@ember/component/template-only';
+ export default templateOnly();
--- a/addon/components/bs-modal/body.js
+++ b/addon/components/bs-modal/body.js
@@ -7,3 +7,5 @@
  @extends Glimmer.Component
  @public
  */
+ import templateOnly from '@ember/component/template-only';
+ export default templateOnly();
--- a/addon/components/bs-modal/body.js
+++ b/addon/components/bs-modal/body.js
@@ -7,5 +7,5 @@
  @extends Glimmer.Component
  @public
  */
- import templateOnly from '@ember/component/template-only';
- export default templateOnly();
+import templateOnly from '@ember/component/template-only';
+export default templateOnly();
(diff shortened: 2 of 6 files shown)


simonihmig:
Thanks @thoov for looking into this!

Linting is failing, but just due to prettier complaining. Please run prettier on your changes, or use the `lint:fix` script...

So, to the real issue: while I trust you that this is is a valid fix, I still would like to understand this better. I didn't know of `@ember/component/template-only` before, seems to have been introduced by the colocation RFC481. The RFC states:

> Again, this function, though public, is not intended to be called by users directly. It is primarily used in the output emitted by build tools.

So why is this need here in user-land code? 

The issue seems to be related to Embroider, right? Which this addon supported once but this regressed through some changes in Embroider, fixes are just waiting for a release (#1341). Working on those fixes, cannot remember seeing the error you quoted, and the test suite was passing locally. 

So I wonder under exactly which circumstances are these changes here needed, for any arbitrary addon that exposes TO components?

rwjblue:
The issue is that they are basically "template only" but there **is** a JS file and that JS file just has no export (I think the file only exists for documentation purposes?). Embroider could and probably should fix this, but using the templateOnly function seemed like a reasonable temporary work around.

simonihmig:
> The issue is that they are basically "template only" but there is a JS file and that JS file just has no export 

Hm, ok, I see. Still puzzles me, as I renamed all those "empty" JS files to have a dash prefix, for the reason that I got build errors otherwise. So there was `-bs-modal-simple.js` and `bs-modal-simple.hbs`. And Embroider still choked on this? 🤔

Which is again strange, because I worked yesterday on fixing the remaining Embroider issues in #1341, and it was all passing locally (CI is still failing for some other reason).

Also I just looked into `ember-cp-validations`, which uses a [very old version](https://github.com/offirgolan/ember-cp-validations/blob/master/package.json#L35) of e-b, apparently for its dummy/demo app . @thoov have you tried first upgrading to the latest version? The 3.x series still used some broccoli tree magic for its Bootstrap 3 vs. 4 support with some unusual internal tree layout, while 4.x is using `@embroider/macros`, so they looked and worked vastly different. TO/Glimmer components and colocation were only introduced during the 4.x cycle.

To be clear: I think the changes here are fine, even just for "internal sanity", as the dash-prefixed files were annoying. So I'll merge them gladly. I am just unsure if that was really the root cause of the problems...

> I think the file only exists for documentation purposes?

Correct

simonihmig (approved):

PR #1516, opened 2021-05-19
Fix horizontal form layout for checkbox elements

### LINKED ISSUE #1511: Horizontal form layout broken for checkboxes
This is from https://www.ember-bootstrap.com/components/forms/ (layout selection is broken 😕 -  but using Ember inspector you can manually set the layout to horizontal):

image

simonihmig:
Fixes #1511

The culprit was extending from the wrong class, which probably happened during those larger refactorings. Added test coverage for all control types.


jelhan (approved):
Thanks a lot!

PR #1528, opened 2021-05-25
refactor to async/await instead of callbacks in BsModal

jelhan:
`<BsModal>` was heavily using callbacks in logic to show and hide a modal. The resulting code was hard to read. This pull requests refactors it to `async`/`await`.

Pulled this change out of #1402, which got way too big.

### DIFF
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -12,6 +12,10 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 import arg from '../utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
 
+function nextRunloop() {
+  return new Promise((resolve) => next(resolve));
+}
+
 /**
   Component for creating [Bootstrap modals](http://getbootstrap.com/javascript/#modals) with custom markup.
 
@@ -416,7 +420,7 @@ export default class Modal extends Component {
    * @method show
    * @private
    */
-  show() {
+  async show() {
     if (this._isOpen) {
       return;
     }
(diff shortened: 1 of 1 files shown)


addon/components/bs-modal.js
```diff
@@ -506,45 +508,33 @@ export default class Modal extends Component {
    * @param callback
    * @private
    */
-  handleBackdrop(callback) {
-    let doAnimate = this.usesTransition;
+  async handleBackdrop() {
```
jelhan, line 511: Would love to refactor `handleBackdrop` into two separate methods: `showBackdrop` and `hideBackdrop`. But this would also require renaming the existing `showBackdrop` property. And I wanted to avoid touching too much in one pull request. I may do that one later.

addon/components/bs-modal.js
```diff
@@ -425,36 +429,34 @@ export default class Modal extends Component {
...
+    await this.handleBackdrop();
+
+    if (this.isDestroyed) {
+      return;
+    }
+
+    this.checkScrollbar();
+    this.setScrollbar();
+
+    schedule('afterRender', async () => {
```
simonihmig, line 443: While we are at it, could we refactor this also to an async helper, similar to `nextRunloop()`, that we can await and then put the other code into the method body itself, rather than the callback?

I like these runloop-related helpers really, makes code much better readable, than the callback-based original primitives. Maybe extract into utils, or potentially an own addon? 😆

addon/components/bs-modal.js
```diff
@@ -473,10 +475,10 @@ export default class Modal extends Component {
     this.showModal = false;
 
     if (this.usesTransition) {
-      transitionEnd(this.modalElement, this.transitionDuration).then(() => this.hideModal());
-    } else {
-      this.hideModal();
+      await transitionEnd(this.modalElement, this.transitionDuration);
     }
+
+    this.hideModal();
```
simonihmig, line 481: Should we `await` this?

simonihmig (commented):
Thanks for the work!

The existing code originated from ancient times, where I mostly tried to copy the original bootstrap.js, and ember-ify it. This is way better! :+1:

simonihmig (approved):

jelhan:
You were a little bit quicker than I'm. Actually planned to move the `nextRunloop` and `afterRender` utils into a separate file. But got distracted. :see_no_evil: Will try to recall when I need them next time. :smile_cat:

PR #1466, opened 2021-04-01
Add test coverage for modal animation

jelhan:
This adds test coverage for modal animation. It encodes the results of my investigation on the topic documented in https://github.com/kaliber5/ember-bootstrap/pull/1402/files#r605945018. I thought it might be helpful to get this in before #1402. And maybe even if I don't find the time to finish that one.

PR #1529, opened 2021-05-26
Shadow DOM compatibility

lolmaus:
This PR avoids looking up elements via `document` and id, because such method is unable to pass through the Shadow DOM border.

Instead, it uses https://github.com/lifeart/ember-ref-bucket by @lifeart to reference elements directly.

### DIFF
--- a/addon/components/bs-contextual-help/element.js
+++ b/addon/components/bs-contextual-help/element.js
@@ -5,6 +5,7 @@ import { scheduleOnce } from '@ember/runloop';
 import arg from 'ember-bootstrap/utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
 import { getOwnConfig, macroCondition } from '@embroider/macros';
+import { trackedRef } from 'ember-ref-bucket';
 
 /**
  Internal (abstract) component for contextual help markup. Should not be used directly.
@@ -96,6 +97,8 @@ export default class ContextualHelpElement extends Component {
 
   offset = [0, 0];
 
+  @trackedRef('popperElement') popperElement;
+
   /**
    * popper.js modifier config
    *
@@ -104,15 +107,19 @@ export default class ContextualHelpElement extends Component {
    * @private
    */
   get popperModifiers() {
-    let id = this.args.id;
+    const context = this;
+
+    // We need popeerElement, so we wait for this getter to recompute once it's available
+    if (!this.popperElement) return {};
+
     return {
       arrow: {
-        element: `.${this.arrowClass}`,
+        element: this.popperElement.querySelector(`.${this.arrowClass}`),
       },
       offset: {
         offset: this.offset.join(','),
         fn(data) {
-          let tip = document.getElementById(id);
+          let tip = context.popperElement;
           assert('Contextual help element needs existing popper element', tip);
 
           // manually read margins because getBoundingClientRect includes difference
--- a/addon/components/bs-contextual-help/element.js
+++ b/addon/components/bs-contextual-help/element.js
@@ -111,10 +111,8 @@ export default class ContextualHelpElement extends Component {
 
     // We need popeerElement, so we wait for this getter to recompute once it's available
     if (!this.popperElement) {
-      console.log('empty!');
       return {};
     }
-    console.log('not empty!');
 
     return {
       arrow: {
--- a/addon/components/bs-contextual-help/element.js
+++ b/addon/components/bs-contextual-help/element.js
@@ -110,7 +110,11 @@ export default class ContextualHelpElement extends Component {
     const context = this;
 
     // We need popeerElement, so we wait for this getter to recompute once it's available
-    if (!this.popperElement) return {};
+    if (!this.popperElement) {
+      console.log('empty!');
+      return {};
+    }
+    console.log('not empty!');
 
     return {
       arrow: {
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -5,12 +5,12 @@ import { next, schedule } from '@ember/runloop';
 import { inject as service } from '@ember/service';
 import transitionEnd from 'ember-bootstrap/utils/transition-end';
 import { getDestinationElement } from 'ember-bootstrap/utils/dom';
-import { guidFor } from '@ember/object/internals';
 import usesTransition from 'ember-bootstrap/utils/decorators/uses-transition';
 import isFastBoot from 'ember-bootstrap/utils/is-fastboot';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 import arg from '../utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
+import { ref } from 'ember-ref-bucket';
 
 /**
   Component for creating [Bootstrap modals](http://getbootstrap.com/javascript/#modals) with custom markup.
@@ -210,30 +210,6 @@ export default class Modal extends Component {
    */
 
   /**
-   * The id of the `.modal` element.
-   *
-   * @property modalId
-   * @type string
-   * @readonly
-   * @private
-   */
-  get modalId() {
-    return `${guidFor(this)}-modal`;
-  }
-
-  /**
-   * The id of the backdrop element.
-   *
-   * @property backdropId
-   * @type string
-   * @readonly
-   * @private
-   */
-  get backdropId() {
-    return `${guidFor(this)}-backdrop`;
-  }
-
-  /**
    * Property for size styling, set to null (default), 'lg' or 'sm'
    *
    * Also see the [Bootstrap docs](http://getbootstrap.com/javascript/#modals-sizes)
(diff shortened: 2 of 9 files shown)


addon/components/bs-contextual-help/element.js
```diff
@@ -96,6 +97,8 @@ export default class ContextualHelpElement extends Component {
 
   offset = [0, 0];
 
+  @trackedRef('popperElement') popperElement;
```
lifeart, line 100: Looks like here there is no need for trackedRef, it should be used if node used inside some reactive getter

lolmaus:
Tests are failing in Canary, but those seem to be related to new deprecations.

Should I attempt fixing them in this PR?

jelhan:
> Tests are failing in Canary, but those seem to be related to new deprecations.
> 
> Should I attempt fixing them in this PR?

Failing tests in beta and canary are a known issue. You can just ignore them unless they are failing in a different way than on current master. See here for details: https://github.com/kaliber5/ember-bootstrap/pull/1466#issuecomment-847961620

addon/components/bs-contextual-help/element.js
jelhan: It is used in the `popperModifiers` getter. There is even a inline comment that it should recompute when `popperElement` is set later:

> ```
> // We need popeerElement, so we wait for this getter to recompute once it's available
> ```

I noticed that `addon/components/bs-popover/element.hbs` uses `{{create-ref "popperElement"}}`, while `addon/components/bs-tooltip/element.hbs` uses `{{create-tracked-ref "popperElement"}}`. Is it different by intention?

addon/utils/dom.js
```diff
@@ -76,7 +76,10 @@ function closest(el, selector) {
 
 export function getDestinationElement(context) {
   let dom = getDOM(context);
-  let destinationElement = findElementById(dom, 'ember-bootstrap-wormhole');
+  const id = 'ember-bootstrap-wormhole';
+  console.log('dom', dom);
+  let destinationElement = findElementById(dom, id) || findElemementByIdInShadowDom(context, id);
+  // AFJLAJLSFJ AFJ LJ
```
jelhan, line 82: `console.log()` and the comment looks like left-over from debugging.

addon/components/bs-contextual-help/element.js
lifeart: yes, `{{create-tracked-ref "popperElement"}}` mean we have to track Node mutations (using mutation observer and such), `{{create-ref "popperElement"}}` should be used in current case

addon/utils/dom.js
```diff
@@ -104,3 +107,8 @@ export function getDestinationElement(context) {
 
   return destinationElement;
 }
+
+export function findElemementByIdInShadowDom(context, id) {
+  const owner = getOwner(context);
+  return owner.rootElement.querySelector && owner.rootElement.querySelector(`[id="${id}"]`);
```
jelhan, line 113: Is `querySelector('#my-id')` not working within a Shadow DOM?

addon/components/bs-modal.js
```diff
@@ -396,9 +368,7 @@ export default class Modal extends Component {
 
   @action
   doSubmit() {
-    // replace modalId by :scope selector if supported by all target browsers
-    let modalId = this.modalId;
-    let forms = this.modalElement.querySelectorAll(`#${modalId} .modal-body form`);
+    let forms = this.modalElement.querySelectorAll('.modal-body form');
```
jelhan, line 371: To be honest I'm not sure why modal ID (or `:scope`) was required here in the first place. But why do you thing it is not needed anymore?

jelhan (commented):
Thanks a lot for working on Shadow DOM compatibility. Looks good to me overall. But I haven't much experience with `ember-ref-bucket`. Would need to look more in detail into it (especially related to autotracking). But maybe @simonihmig can pick it up before. :smile_cat:

addon/components/bs-contextual-help/element.js
simonihmig: > Is it different by intention?

I don't think so. Seems wrong?

lolmaus: Hmm... Initially I wrote everything with non-tracked versions and it did not solve my case. Making that single entry tracked resolved my case. 🤷‍♂️

simonihmig: The use of tracked-ref vs. ref seems to be exchanged for overlayElement and popperElement, when you compare popovers and tooltips!

addon/utils/dom.js
lolmaus: Yes.

addon/components/bs-modal.js
lolmaus: Well, the immediate reasons were that:

* `#some-id` selectors did not work inside Shadow DOM for me. Not sure why, I did not investigate.
* I removed the `modalId` entirely.

I made this specific change light hearted because I noticed that `modalId` pointed at `this.modalElement` anyway, so there was no reason to scope the selector by `this.modalElement` when searching inside `this.modalElement`. 🤷‍♂️

AFAIK, the only legit reason to reference the modal itself would be to use the immediate descendant combinator, e. g. `:scope > div`, because simply `> div` would be considered an invalid selector.

Also note that passing`":scope"` as `modalId` into the original code:

```
this.modalElement.querySelectorAll(`#${modalId} .modal-body form`);
```

...would end up as invalid selector `#:scope .modal-body form`.

addon/utils/dom.js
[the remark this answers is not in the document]
lolmaus: I just marked a line during a Meet to work on later. 😓

addon/utils/dom.js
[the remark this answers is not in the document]
lolmaus: Initially yes, but then we stubmled randomly upon `owner.rootElement` being a string.

So I decided to keep this tiny guard. Does not hurt.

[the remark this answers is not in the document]
simonihmig: Ah, ok. I misunderstood the code, should have look at it for a second longer 🤦‍♂️
So I was basically "complaining" about that not being guarded, as I was expecting some `typeof owner.rootElement` like check, but I see it is effectively guarded. You could refactor this to use optional chaining though...

```suggestion
  return owner.rootElement.querySelector?.(`[id="${id}"]`);
```

addon/components/bs-contextual-help/element.js
lolmaus: Uhm... I've just tested it out, and it turns out that having the `@trackedRef` decorator on `popperElement` is enough for the `popperModifiers` getter (which consumes `this.popperElement`) to start recomputing.

In the template I have replaced `{{create-tracked-ref}}` with `{{create-ref}}` and it still works. Whereas replacing `@trackedRef` with `@ref` breaks it.

addon/utils/dom.js
lolmaus: Removed.

addon/utils/dom.js
```diff
@@ -76,7 +76,10 @@ function closest(el, selector) {
 
 export function getDestinationElement(context) {
   let dom = getDOM(context);
-  let destinationElement = findElementById(dom, 'ember-bootstrap-wormhole');
+  const id = 'ember-bootstrap-wormhole';
+  console.log('dom', dom);
```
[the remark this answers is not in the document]
lolmaus: Removed.

addon/components/bs-contextual-help/element.js
simonihmig: Seems good now. I didn't really know about `{{create-tracked-ref}}` before, but it seems to cover a different but not needed use case here, when properties of the DOM element itself change, like adding a MutationObserver and the like!? See https://github.com/lifeart/ember-ref-bucket#definition-of-create-tracked-ref-modifiers

But again, we don't need this here, just need the ref-property in the component to be tracked.

simonihmig (approved):
LGTM now! :+1:

addon/utils/dom.js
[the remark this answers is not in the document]
lolmaus: Is optional chaining transpiled correctly? I've only used it with TypeScript so far.

simonihmig:
FWIW it seems the `{{hash}}` helper deprecation now found it's way also into `ember-release`, so that is now also causing failed CI jobs (unrelated to this PR). I prepared a fix upstream at https://github.com/josemarluedke/ember-focus-trap/pull/35. @jelhan fyi

addon/utils/dom.js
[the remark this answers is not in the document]
simonihmig: Yes, recent enough versions of ember-cli-babel support it ootb. We use that here frequently!

lolmaus:
> the `{{hash}}` helper deprecation


`{{hash}}` is deprecated?..

simonihmig:
No, not at all. But setting properties on `{{hash}}` outside of the template. Like you pass it to a component `@model={{hash foo=true}}`, and the component does `this.args.model.bar=true;`. This raises now a deprecation. See:
* https://github.com/emberjs/ember.js/pull/19548
* https://github.com/glimmerjs/glimmer-vm/pull/1298#issuecomment-840530878

jelhan:
> > the `{{hash}}` helper deprecation
> 
> `{{hash}}` is deprecated?..

Not the `{{hash}}` helper itself but mutating the value returned by it:

> You set the 'initialFocus' property on a {{hash}} object. Setting properties on objects generated by {{hash}} is deprecated. Please update to use an object created with a tracked property or getter, or with a custom helper.

It doesn't seem to be listed yet on https://deprecations.emberjs.com/v3.x.

PR #1531, opened 2021-05-26
refactor handleBackdrop method of <BsModal> into separate show and hide methods

jelhan:
The `handleBackdrop` method was responsible both for showing and hiding the backdrop. This was not only hard to read. It was also error prune as `handleBackdrop` may have done the opposite of what was expected. Actually that was exactly the case in one of the animation tests recently added. For some reason `this.open` was `true` when `handleBackdrop` was called by `hide` method. The test was passing but not asserting the closing animation. :exploding_head:

This is another refactor, which I pulled out of #1402.


tests/integration/components/bs-modal-test.js
```diff
@@ -247,59 +248,69 @@ module('Integration | Component | bs-modal', function (hooks) {
...
-
-    this.set('open', true);
-    await waitFor('.modal.show');
-    assert.dom('.modal').hasStyle({ display: 'block' });
-
-    await waitUntil(() => {
-      return find('.modal').getAnimations().length > 0;
+  module('it animates opening and closing the modal', function () {
+    hooks.before(function () {
+      skipTransition(false);
```
simonihmig, line 253: Oh, right! 🤦‍♂️

simonihmig (approved):
Looks great! Thanks for iterating on this, code is improving by the minute! 😁

simonihmig:
There is just a merge conflict left...

PR #1532, opened 2021-05-27
remove trailing bracket in <BsModal::Header>

jelhan:
Got introduced in #1341, which was released in v4.8.0 a few hours ago.


simonihmig:
Oh, wtf, good catch, thanks!

PR #1533, opened 2021-05-27
harmonize code paths used in FastBoot and non-FastBoot of <BsModal>

jelhan:
Running same code paths in FastBoot as in "regular" environment reduces the risk of bugs. To do so I moved the guards for FastBoot as close as possible to the places in code, which access `document` or `window`. Tests are passing and I manually verified that `/fastboot/modal` URL of dummy app shows the modal even with JavaScript being disabled. Not sure if additional manual tests are required.

Another small refactoring pulled out of #1402.

### DIFF
--- a/addon/components/bs-modal.js
+++ b/addon/components/bs-modal.js
@@ -417,8 +417,10 @@ export default class Modal extends Component {
       return;
     }
 
-    this.checkScrollbar();
-    this.setScrollbar();
+    if (!isFastBoot(this)) {
+      this.checkScrollbar();
+      this.setScrollbar();
+    }
 
     await afterRender();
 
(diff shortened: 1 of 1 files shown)


jelhan:
Main driver of the refactoring was to run `show` method in `handleVisibilityChanges` and not skip it early. Mostly because that relies on modal being opened in a different way than through `show` method. Basically the coupling of internal and external modal state, which I tried to address in #1402.

Not sure if I have any opinion if `setScrollbar` should guards against FastBoot or if the caller should do so. Happy to refactor to something like this if you prefer:

```js
async show() {
  // ...

  if (!isFastBoot(this)) {
    this.checkScrollbar();
    this.setScrollbar();
  }
  
  // ...
}
```

simonihmig (approved):

PR #1535, opened 2021-05-28
Fix BsForm element causing "same computation" Glimmer assertion in rare edge case

simonihmig:


jelhan:
Wondering if this is also fixing #1414 and https://github.com/kaliber5/ember-bootstrap-changeset-validations/issues/37. In that cases the reproduction included submitting the form using enter key. But it is easy to imagine that it has the same affect.

jelhan (approved):

simonihmig:
Could very well be the case! 

https://github.com/kaliber5/ember-bootstrap/pull/1415 Seems to have fixed a problem by not *reading* the tracked property, while this will not update it when the value hasn't actually changed.

PR #1537, opened 2021-05-28
Remove `.btn-block` support for Bootstrap 5

simonihmig:
See https://getbootstrap.com/docs/5.0/migration/#buttons

### DIFF
--- a/addon/components/bs-button.hbs
+++ b/addon/components/bs-button.hbs
@@ -1,7 +1,7 @@
 <button
   disabled={{this.__disabled}}
   type={{this.buttonType}}
-  class="btn {{if @active "active"}} {{if this.block "btn-block"}} {{bs-size-class "btn" @size}} {{bs-type-class "btn" @type default=(if (macroCondition (macroGetOwnConfig "isBS3")) "default" "secondary") outline=@outline}}"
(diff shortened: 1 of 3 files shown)


jelhan (changes requested):
I think we need to update API docs for `@block` argument as well to mention that it is not supported for BS5. Maybe an assertion throwing if the argument is used in BS5 would be also helpful.

simonihmig:
I updated the JSDocs comment. Not sure about the assertion, like yes, would be helpful, but we have similar cases throughout the addon where we don't do this, so do we really start to do this here and not elsewhere?

jelhan:
> I updated the JSDocs comment. Not sure about the assertion, like yes, would be helpful, but we have similar cases throughout the addon where we don't do this, so do we really start to do this here and not elsewhere?

I think a rule to add assertions to the arguments, not required by Bootstrap 5 would be helpful for migration. Do you think it would be a lot of work?

I would not care much about APIs required only for Bootstrap 3 or not supported for Bootstrap 3. The APIs either get mandatory or removed when dropping Bootstrap 3 support.

jelhan (approved):
I gave assertions for arguments only supported in some bootstrap versions another thought. I think you are right that we shouldn't invest time into. It will be better addressed when we have typed templates.

PR #1538, opened 2021-05-28
Fix BsButtonGroup radio to not trigger onChange when clicking active button

simonihmig:
The test was actually broken,so a false positive. Refactored tests to angel brackets.

The case for a checkbox type group was not covered by tests, so not sure if I should cover that only, as this could be seen as a breaking change possible?


jelhan:
Do you know when this bug was introduced? Wondering if fixing could be seen as a breaking change.

simonihmig:
No, not sure.

PR #1560, opened 2021-06-07
Ignore `heading-order` a11y check

simonihmig:
This fails CI, apparently a newer version of axe introduced stricter rules, without a major bump. For now disabling the offending rule for 2 components, not sure if we can actually fix this, given that Bootstrap dictates the expected markup.

### DIFF
--- a/tests/integration/components/bs-accordion-test.js
+++ b/tests/integration/components/bs-accordion-test.js
@@ -269,6 +269,7 @@ module('Integration | Component | bs-accordion', function (hooks) {
           enabled: false,
         },
         'color-contrast': { enabled: false },
+        'heading-order': { enabled: false },
       },
     });
     assert.ok(true, 'A11y audit passed');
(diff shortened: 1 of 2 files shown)


jelhan:
> not sure if we can actually fix this, given that Bootstrap dictates the expected markup.

Would be great if you could point me to failing tests.

simonihmig:
Well, if you look at the CI runs, all of them are failing for modal and accordion a11y checks like this: https://github.com/kaliber5/ember-bootstrap/runs/2762319596?check_suite_focus=true#step:5:2149

jelhan:
> Well, if you look at the CI runs, all of them are failing for modal and accordion a11y checks like this: https://github.com/kaliber5/ember-bootstrap/runs/2762319596?check_suite_focus=true#step:5:2149

Thanks. Haven't noticed them yet.

I was wondering if heading levels are mandatory by Bootstrap or just used as an example. And even after reading latest docs it seems unclear to me. 😒

Issue #999, opened 2020-02-14
Fix ARIA for accordion


simonihmig:
* accordions have roles like `tab`, `tablist` and `tabpanel` stemming from Bootstrap 3 examples (https://getbootstrap.com/docs/3.4/javascript/#collapse-example-accordion). These should be removed.
* instead `aria-expanded` should be used, see https://getbootstrap.com/docs/4.4/components/collapse/#accessibility and https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties

PR #1561, opened 2021-06-07
Add Bootstrap v5 support for accordions

simonihmig:
Had to refactor the tests (selectors) a bit, as they were not *quite* right, and failed for BS5 (e.g. assumed you were clicking on the heading, instead of the button contained within it).

Didn't touch the a11y markup (yet), needs to be fixed separately, see #999.

### DIFF
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -31,7 +31,7 @@
     {{/if}}
     {{#if (macroCondition (macroGetOwnConfig "isBS3"))}}
       <h4 class="panel-title">
-        <a href="#" class="{{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}" disabled={{@disabled}}>
+        <a href="#" class="{{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}">
           {{yield}}
         </a>
       </h4>
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -1,23 +1,40 @@
 {{!-- template-lint-disable no-nested-interactive --}}
 {{!-- @todo fix this, see https://github.com/kaliber5/ember-bootstrap/issues/999 --}}
-<div
-  class="{{if (macroCondition (macroGetOwnConfig "isNotBS3")) "card-header"}} {{if (macroCondition (macroGetOwnConfig "isBS3")) "panel-heading"}}"
-  role="tab"
-  ...attributes
-  {{on "click" this.handleClick}}
->
-  {{#if (macroCondition (macroGetOwnConfig "isNotBS3"))}}
-    <h5 class="mb-0">
-      <button class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}" type="button" disabled={{@disabled}}>
-        {{yield}}
-      </button>
-    </h5>
-  {{/if}}
-  {{#if (macroCondition (macroGetOwnConfig "isBS3"))}}
-    <h4 class="panel-title">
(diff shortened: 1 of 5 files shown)


addon/components/bs-accordion/item/title.hbs
```diff
@@ -1,23 +1,40 @@
...
+{{else}}
+  <div
+    class="{{if (macroCondition (macroGetOwnConfig "isBS3")) "panel-heading"}} {{if (macroCondition (macroGetOwnConfig "isBS4")) "card-header"}}"
+    role="tab"
+    ...attributes
+    {{on "click" this.handleClick}}
+  >
+    {{#if (macroCondition (macroGetOwnConfig "isBS4"))}}
+      <h5 class="mb-0">
+        <button class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed"}}" type="button" disabled={{@disabled}}>
```
simonihmig, line 27: Had to change BS3/4 markup a little, putting `.collapsed` on the button/a instead of the parent div. This is actually expected (see e.g. https://getbootstrap.com/docs/4.6/components/collapse/#accordion-example), so could be seen here as a bug fix. Though it might theoretically cause some regressions for users... 🤔

It does not affect the CSS styling in any way for BS3/4 AFAICT, but it does for BS5, so had to be changed.

addon/components/bs-accordion/item/title.hbs
```diff
@@ -1,23 +1,40 @@
...
+          {{yield}}
+        </button>
+      </h5>
+    {{/if}}
+    {{#if (macroCondition (macroGetOwnConfig "isBS3"))}}
+      <h4 class="panel-title">
+        <a href="#" class="{{if @disabled "disabled"}} {{if @collapsed "collapsed"}}" disabled={{@disabled}}>
```
jelhan, line 34: This is also fixing a bug independently from Bootstrap 5 support, isn't it?

jelhan (commented):
I'm very sorry that review took that long. To be honest I fully missed this one.

addon/components/bs-accordion/item/title.hbs
simonihmig: > So I guess we don't need to address that bug here.

You are right. Extracted the changes to #1590.

Will rebase this PR once the fix has landed.

simonihmig: @jelhan done now! The diff is now much smaller, and confined to only BS5 related changes!

addon/components/bs-accordion/item/title.hbs
```diff
@@ -1,23 +1,40 @@
...
+    {{#if (macroCondition (macroGetOwnConfig "isBS4"))}}
+      <h5 class="mb-0">
+        <button class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}" type="button" disabled={{@disabled}}>
+          {{yield}}
+        </button>
+      </h5>
+    {{/if}}
+    {{#if (macroCondition (macroGetOwnConfig "isBS3"))}}
+      <h4 class="panel-title">
+        <a href="#" class="{{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}" disabled={{@disabled}}>
```
jelhan, line 34: Did you added the `disabled` attribute intentionally? I don't think it is suported for link elements (`<a>`). Also it does not seem to be related to Bootstrap 5 support.

jelhan (changes requested):
Much easier to review after the bug fixes has been extracted into separate merge requests. Thanks a lot!

Just a small question left. Beside that everything looks good.

addon/components/bs-accordion/item/title.hbs
simonihmig: Hm, was probably left over from the previous monolithic PR. Removed it now, good catch!

jelhan (approved):

PR #1572, opened 2021-06-29
Show validation errors added by onSubmit

misli:
The `onSubmit` action may set additional validation errors
(typically errors received from a backend server).
This change makes these errors visible.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -454,7 +454,10 @@ export default class Form extends Component {
                 return;
               }
 
-              this.set('isRejected', true);
+              this.setProperties({
+                showAllValidations: true,
+                isRejected: true,
+              });
 
               throw error;
             })


jelhan (changes requested):
Thanks a lot.

Are you sure that it isn't overruled by `finally`, which sets `showAllValidations` to `undefined`? Would be great if you could add tests. Not only to verify that one but also to ensure that we don't introduce a regression later.

I'm wondering if this is the correct behavior in all cases. Or if it should be controlled via an option like `hideValidationsOnSubmit`. A `@onSubmit` handler could fail for many cases. Violating validation rules only enforced by backend is only one of them.

Similar I'm wondering if it should be considered a breaking change. In that case we would need to control it by option. And the default must match current behavior. At least until release next major. @simonihmig any thoughts on that one?

simonihmig:
> Would be great if you could add tests.

Yes, please!

> Similar I'm wondering if it should be considered a breaking change

I agree with your concerns. When not adding validation errors in `onSubmit`, setting `showAllValidations` will introduce a side effect that likely is unwanted, as all fields will be marked as successful (green), even when `hideValidationsOnSubmit` has been set, or some (not required) field hasn't been filled at all.

So an option like `showValidationsOnRejection` seems preferable probably!? 🤔

PR #1580, opened 2021-07-14
Add @manualCloseOnly property to dropdown to permit blocking click outside closing

gfmartinez:
We have several use cases that would benefit from being able to disable closing of the dropdown menu when there is a click outside.

For example, when we have a small form inside of the dropdown. 
If a user enters some information and accidentally clicks outside the menu, or tabs past a submit button to the other elements on the page, the dropdown is closed and information entered is lost since the child component is destroyed.

What would you think about something like this to allow an override?

Happy to reconsider naming of the override, elaborate on test, whatever works.

Thanks again for the great library!

### DIFF
--- a/addon/components/bs-dropdown.js
+++ b/addon/components/bs-dropdown.js
@@ -197,6 +197,18 @@ export default class Dropdown extends Component {
   isOpen = false;
 
   /**
+   * By default clicking outside a menu, tab or esc keypress will close it. Set this property to true for the menu to stay open.
+   * Use of this override will require calling the closeDropdown method in the menu to close it.
+   *
+   * @property manualCloseOnly
+   * @default false
+   * @type boolean
+   * @public
+   */
+  @defaultValue
+  manualCloseOnly = false;
+
+  /**
    * By default clicking on an open dropdown menu will close it. Set this property to false for the menu to stay open.
    *
    * @property closeOnMenuClick
@@ -311,6 +323,8 @@ export default class Dropdown extends Component {
    */
   @action
   closeHandler(e) {
+    if (this.manualCloseOnly) return;
+
     let { target } = e;
     let { toggleElement, menuElement } = this;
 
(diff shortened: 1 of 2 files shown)


jelhan:
I like the proposal. But I'm not convinced by having one `@manualCloseOnly` argument.

If I get it right from the API description added in this pull request, you like to disable to close the dropdown in four cases:

1. If user clicks outside of the dropdown (backdrop).
2. If user presses escape key.
3. If user moves focus to an element outside of the backdrop using tab key.
4. If user clicks on menu item. This is already supported by `@closeOnMenuClick`.

Doing all this with one argument does not seem to fit well with the existing APIs:

- `<BsDropdown>` already has a `@closeOnMenuClick` argument, which controls if dropdown is closed when user clicks on one if its items.
- `<BsModal>` has a `@backdropClose` argument, which controls if modal is closed by a click on dropdown.
- `<BsModal>` has a `@onHide` event, which allows to prevent closing the modal by returning `false`.

`<BsDropdown>` already supports a `@onHide` event. Maybe it should also support preventing close of the modal by returning `false`? This would solve your use case as well, wouldn't it? At the same time it would allow more flexibility. To stay with your example of a form within a dropdown, it would allow a consumer to prevent closing only if user already entered data.

gfmartinez:
@jelhan Thanks so much for following up. You are correct in the use cases.
I love your suggestion to support prevention of close via `@onHide` callback.
That would be very flexible and fit better within the structure.
I'd be happy to submit a followup if that works.

simonihmig:
> I love your suggestion to support prevention of close via @onHide callback.
> That would be very flexible and fit better within the structure.
> I'd be happy to submit a followup if that works.

Sounds good to me! Looking forward to a change of this PR (or a new one?)! Meanwhile will put this into draft mode, please change it back when again ready to review!

Thanks for working on this @gfmartinez!

gfmartinez:
Closed in favor of https://github.com/kaliber5/ember-bootstrap/pull/1607

PR #1581, opened 2021-07-15
BsButton should not silenty catch errors thrown by @onClick handler

### LINKED ISSUE #1568: <BsButton>'s @onClick handler silently catches errors for async function
If an `@onClick` event handler returns a `Promise`, which rejects, `<BsButton>` silently catches them. They will never get to console in developer tools. Or be logged by error reporting tools like Sentry. This makes debugging very challenging.

Please note that this only affects _async_ event handlers. If the function passed to `@onClick` does not return a `Promise` errors are _not_ catched and reported in console as expected.

Please find a reproduction here: https://github.com/jelhan/reproduce-bs-button-on-click-event-handler-silently-catching-errors/commit/54b1534f60a25b69d237f7e050d4e4b070f9f5bc

I'm already working on a fix. But facing some issues with test setup. As that may delay the bug fix, I wanted to report to make others aware.

jelhan:
Fixes #1568 and add tests coverage for `<BsForm>`'s `@onSubmit` handler to not have a similar issue.

### DIFF
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -17,12 +17,12 @@ module('Integration | Component | bs-button', function (hooks) {
 
   hooks.beforeEach(function (assert) {
     /*
-     * The custom assertions `rejectsErrorNotGoinghThroughEmberOnError` allows to test
+     * The custom assertions `rejectsErrorNotGoingThroughEmberOnError` allows to test
      * functionality, which rejects intentionally. It serves a similar use case as
      * `assert.throws` and `setupOnerror` utility provided by `@ember/test-helpers`.
      * Sadly for some use cases none of them is working.
      *
-     * `rejectsErrorNotGoinghThroughEmberOnError` should only be used for cases in which
+     * `rejectsErrorNotGoingThroughEmberOnError` should only be used for cases in which
      * neither  `assert.throws` nor `setupOnerror` could be used. Otherwise these two
      * methods to test rejecting promises should be preferred.
      */
--- a/tests/integration/components/bs-button-test.js
+++ b/tests/integration/components/bs-button-test.js
@@ -1,5 +1,5 @@
 import { run } from '@ember/runloop';
-import { defer, reject, resolve } from 'rsvp';
+import { defer, resolve } from 'rsvp';
 import { module, skip } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { find, render, click, settled, waitUntil } from '@ember/test-helpers';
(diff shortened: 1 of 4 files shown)


tests/integration/components/bs-button-test.js
```diff
@@ -9,11 +9,44 @@ import hbs from 'htmlbars-inline-precompile';
...
 import sinon from 'sinon';
+/* global QUnit */
 
 module('Integration | Component | bs-button', function (hooks) {
   setupRenderingTest(hooks);
   setupNoDeprecations(hooks);
 
+  hooks.beforeEach(function (assert) {
+    /*
+     * The custom assertions `rejectsErrorNotGoinghThroughEmberOnError` allows to test
```
simonihmig, line 20:
```suggestion
     * The custom assertions `rejectsErrorNotGoingThroughEmberOnError` allows to test
```

tests/integration/components/bs-button-test.js
```diff
@@ -9,11 +9,44 @@ import hbs from 'htmlbars-inline-precompile';
...
+     * functionality, which rejects intentionally. It serves a similar use case as
+     * `assert.throws` and `setupOnerror` utility provided by `@ember/test-helpers`.
+     * Sadly for some use cases none of them is working.
+     *
+     * `rejectsErrorNotGoinghThroughEmberOnError` should only be used for cases in which
```
simonihmig, line 25:
```suggestion
     * `rejectsErrorNotGoingThroughEmberOnError` should only be used for cases in which
```

tests/integration/components/bs-button-test.js
```diff
@@ -9,11 +9,44 @@ import hbs from 'htmlbars-inline-precompile';
...
 module('Integration | Component | bs-button', function (hooks) {
   setupRenderingTest(hooks);
   setupNoDeprecations(hooks);
 
+  hooks.beforeEach(function (assert) {
+    /*
+     * The custom assertions `rejectsErrorNotGoinghThroughEmberOnError` allows to test
+     * functionality, which rejects intentionally. It serves a similar use case as
+     * `assert.throws` and `setupOnerror` utility provided by `@ember/test-helpers`.
+     * Sadly for some use cases none of them is working.
```
simonihmig, line 23: Can you explain under which specific circumstances this does not work?

tests/integration/components/bs-button-test.js
```diff
@@ -9,11 +9,44 @@ import hbs from 'htmlbars-inline-precompile';
...
+     *
+     * `rejectsErrorNotGoinghThroughEmberOnError` should only be used for cases in which
+     * neither  `assert.throws` nor `setupOnerror` could be used. Otherwise these two
+     * methods to test rejecting promises should be preferred.
+     */
+    assert.rejectsErrorNotGoingThroughEmberOnError = async function (callback, expectedError) {
```
simonihmig, line 29: Should we put this into our custom test helpers module, so we can reuse it in the future?

tests/integration/components/bs-button-test.js
```diff
@@ -403,6 +448,20 @@ module('Integration | Component | bs-button', function (hooks) {
...
+  test('it does not catch errors throws by @onClick event handlers', async function (assert) {
+    const expectedError = new Error('error thrown for testing');
+
+    this.set('clickHandler', async () => {
+      throw expectedError;
+    });
+
+    await render(hbs`<BsButton @onClick={{this.clickHandler}} />`);
+
+    await assert.rejectsErrorNotGoingThroughEmberOnError(async () => {
```
simonihmig, line 460: For `@onSubmit` you tested also the sync case. Should we do this here also?

simonihmig (commented):
The changes look good to me, per se.

But I wonder if we have to regard this as a breaking change? We could pretend this is just a bug fix, but the tests asserting a different behavior seem like a strong signal towards the former?

We could do a major bump, including dropping node 10 support, and maybe some other breaking changes (do we have existing deprecations?) 🤔

tests/integration/components/bs-button-test.js
jelhan: `assert.throws` can only be used for synchronous functions. Test helpers like `click`, `visit`, `fillIn` provided by `@ember/test-helpers` are async. So that can not be used at all. But to be honest, I think I wanted to mention `assert.rejects` here. :see_no_evil: I addressed this in 4f036bc826b6cfc2c1ce38d00c2e720399df7f3c.

Test helpers like `click` and `fillIn` may cause application code to run, which throws an unhandled error or causes an unhandled rejection. But they will resolve even in that case. So `assert.rejects(async () => await click('button'))` will not work because `click('button')` will resolve even if it causes an unhandled rejection.

`setupOnerror` only works for errors, which are handled by `Ember.onerror`. It is only a nice helper about mocking `Ember.onerror` in tests.

Some errors don't go through `Ember.onerror`. This is discussed here: https://github.com/emberjs/ember-qunit/issues/592 To be honest I still don't understand the why. For me it sounds like a bug. :shrug: And I don't know for sure in which cases an error (or rejection) is not handled by `Ember.onerror`. Especially I don't understand why we aren't facing the same issues in the tests for `@onSubmit` handler of `<BsForm>`. :confused: 

When investigating (and before noticing that issue in ember-qunit) I created this playground to reproduce: https://github.com/jelhan/reproduce-component-event-handler-throws-testing-issues/compare/b95001721b9b027c444926db168975ed846fdac3...master

jelhan (commented):
> But I wonder if we have to regard this as a breaking change? We could pretend this is just a bug fix, but the tests asserting a different behavior seem like a strong signal towards the former?

Silently catching an error thrown by consumers is clearly a bug for me. But I see that some consumers could rely on that behavior. Even some of our tests were relying on the buggy behavior.

A similar change for `@onSubmit` was done as part of `v3.0.0` release candidate. If I recall correctly having that one part of a major release was only a coincidence. So it may not serve as a reference.

We could do the same here if planning a major version dump anyways. But I think it should not serve as a reference for future cases. I fear we have more of bugs of these kind. And requiring a new major for each of them might limit our ability to fix them quickly.

Also it would be great if we don't delay this specific fix by bundling it in a major version. Have wasted quite some time on debugging of an issue, which would have been obvious if the error would have been logged...

jelhan:
Tests are failing for canary only. It is not related to this change. They are failing for simple upgrades (e.g. #1587) as well.

tests/integration/components/bs-button-test.js
simonihmig: Ok, thanks for the explanation!

simonihmig (approved):
Sorry it took me so long to get back at this!

I'll accept your view regarding bug vs. breaking changes, hope that it does not cause any trouble for users...

Issue #1261, opened 2020-09-25
disabled, is-required and is-validating classes rendered on .form-group element do not exist in Bootstrap


jelhan:
I noticed that `.form-group` element rendered by `<Form::element>` has some classes, which aren't existing in Bootstrap: https://github.com/kaliber5/ember-bootstrap/blob/ce9f350afd8001aa005fd5b5ace4850ee3480e9b/addon/components/bs-form/element.hbs#L3

`.disabled` and `.is-required` classes were both added in https://github.com/kaliber5/ember-bootstrap/pull/71 a long time ago. `.is-validating` was added around the same time by https://github.com/kaliber5/ember-bootstrap/pull/70.

`.is-required` and `.is-validating` classes do not seem to exist in Bootstrap 3 and 4 at all. `.disabled` class exists and is used in the docs a lot. But I wasn't able to find any usage on the `.form-group` element neither in Bootstrap 3 nor 4 docs.

I think we should remove all three. But as that would be a breaking change we can not do it before next major release.

Theoretically we can deprecate them by adding an optional feature, which conditionally removes them. We could print a deprecation warning if it's not explicitly enabled. But I'm not sure if it makes sense to spent the time on it.

PR #1590, opened 2021-08-06
Fix `.collapsed` markup for accordions

simonihmig:
Extracted the bug fix from #1561

### DIFF
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -1,21 +1,21 @@
 {{!-- template-lint-disable no-nested-interactive --}}
 {{!-- @todo fix this, see https://github.com/kaliber5/ember-bootstrap/issues/999 --}}
 <div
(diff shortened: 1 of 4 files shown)


simonihmig:
@jelhan do you want to / have the time to review this? I think this should not be very controversial, just a lot of noise in the tests diff. So if you don't have the time, then I would like to move this forward quickly nevertheless, also as it blocks #1561...

jelhan (approved):
Looks good to me.

I'm just a little bit confused that I haven't found `.expanded` class in BS3 nor BS4 docs. Are you sure that class is needed? Or is that maybe another big?

simonihmig:
> I'm just a little bit confused that I haven't found .expanded class in BS3 nor BS4 docs. Are you sure that class is needed? Or is that maybe another big?

Hm, I wasn't aware tbh. Probably something along the lines of https://github.com/kaliber5/ember-bootstrap/issues/1261, where some classes were added for convenience (e.g. styling an expanded item specifically, without putting the styles into all items, and overriding them for `.collapsed`). But I think we can remove that in v5 if we want. All supported browsers will then also support `:not()`...

PR #1595, opened 2021-08-06
Drop node 10 support

simonihmig:


jelhan:
What about documentation app? I guess we need to upgrade node there as well? https://github.com/kaliber5/ember-bootstrap/blob/3f2853f1899d79ff573609cba7143a814ac92921/docs/package.json#L73-L75

PR #1596, opened 2021-08-06
Drop IE11 support

simonihmig:
As Ember 4.0 will also drop it, there is no point in keeping support.

@jelhan I think we can also basically revert the changes of #775 here, or is there still any value in keeping the [this fallback behavior](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-form/element/control/input-test.js#L28)?


tests/dummy/config/targets.js
```diff
@@ -2,13 +2,6 @@
 
 const browsers = ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
 
-const isCI = Boolean(process.env.CI);
-const isProduction = process.env.EMBER_ENV === 'production';
-
-if (isCI || isProduction) {
-  browsers.push('ie 11');
-}
```
jelhan, line 10: Should we also remove IE11 from build targets for documentation app? https://github.com/kaliber5/ember-bootstrap/blob/master/docs/config/targets.js

jelhan (commented):
> @jelhan I think we can also basically revert the changes of #775 here, or is there still any value in keeping the [this fallback behavior](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-form/element/control/input-test.js#L28)?

I think we can revert. If I recall correctly only IE11 throws if using unsupported `type`.

tests/dummy/config/targets.js
simonihmig: Done

jelhan (approved):

PR #1607, opened 2021-09-01
Permit blocking of dropdown close if onHide returns false

gfmartinez:
Originally: https://github.com/kaliber5/ember-bootstrap/pull/1580

We have several use cases that would benefit from being able to disable closing of the dropdown menu when there is a click outside.

For example, when we have a small form inside of the dropdown.
If a user enters some information and accidentally clicks outside the menu, or tabs past a submit button to the other elements on the page, the dropdown is closed and information entered is lost since the child component is destroyed.

Note: The syntax here is different than in the bs-modal because the bs-dropdown component is not using `this.args`. I'd assume this would refactor when the whole bs-dropdown component is updated in the future.

### DIFF
--- a/addon/components/bs-dropdown.js
+++ b/addon/components/bs-dropdown.js
@@ -274,6 +274,7 @@ export default class Dropdown extends Component {
 
   /**
    * Action is called when dropdown is about to be hidden
+   * Returning `false` will block closing the dropdown
    *
    * @event onHide
    * @param {*} value
@@ -298,8 +299,9 @@ export default class Dropdown extends Component {
 
   @action
   closeDropdown() {
+    if (this.onHide() === false) return;
+
     this.set('isOpen', false);
-    this.onHide();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jelhan (approved):
Looks good to me. Thanks a lot for the contribution!

simonihmig (approved):
Thank you! :+1:

simonihmig:
@gfmartinez this is good to be merged, thanks again for your contribution!

*However* please note that the current `master` branch is already used for the future 5.0 version, that is containing a bunch of breaking changes (a.o. removing BS3 support). Is it ok if we land your changes here with that version? Alternatively we could think of publishing another 4.x release with this change, so people/you can use this without having to upgrade (yet) to that new major. What do you think?

gfmartinez:
@simonihmig Thanks for asking! I'm fine with whatever is easiest for you: you guys know what the roadmap is.
If this type of thing means we have to upgrade when the next version is released, that's just one more reason to do it!

PR #1609, opened 2021-09-03
Drop ember-popper in favor of ember-popper-modifier and popper.js v2

simonihmig:
ember-popper wasn't really maintained much, and ember-popper-modifier seems like the more robust and modern solution. Also it uses popper.js v2, which is also the version that original Bootstrap 5 uses.

Note that due to the fact that Bootstrap 4 uses popper.js v1 (in its original JS components, that we don't use of course), some of the markup (margins of arrows) are working well with popper.js v2 (which ignores these margins), so this required some not so beautiful workarounds (CSS overrides).

Hopefully this should now also maker Ember 4 tests pass, which failed due to some outdated dependency of ember-popper...

### DIFF
--- a/addon/components/bs-contextual-help/element.js
+++ b/addon/components/bs-contextual-help/element.js
@@ -1,7 +1,5 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
-import { assert } from '@ember/debug';
-import { scheduleOnce } from '@ember/runloop';
 import arg from 'ember-bootstrap/utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
 import { getOwnConfig, macroCondition } from '@embroider/macros';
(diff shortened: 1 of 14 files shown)


vst:
@simonihmig Do you need us to test this PR? Or do you have anything to commit before we do so?

Our large codebase is now on `ember-bootstrap@5.0.0-alpha.1`, and things seem to be working fine. We noticed that BS5 `dropdown-menu-end` is not working as expected, though (from BS5 docs):

>  **Heads up!** Dropdowns are positioned thanks to Popper except when they are contained in a navbar.

simonihmig:
> Do you need us to test this PR? Or do you have anything to commit before we do so?

@vst thanks for chiming in! Testing would be awesome for sure. I think I can merge this and publish a new beta version by the end of the week!

> We noticed that BS5 dropdown-menu-end is not working as expected, though (from BS5 docs):

Could you please file an issue, with some more details?

@jelhan I am going to merge this now, as this makes CI green again finally. If you still want to review later - after your vacation IIRC or whenever you have time - the n please do so, I can still address feedback in a later iteration!

vst:
> > We noticed that BS5 dropdown-menu-end is not working as expected, though (from BS5 docs):
> 
> Could you please file an issue, with some more details?

@simonihmig Just testing `5.0.0-beta.0`. It seems that this has been fixed automatically :+1:

PR #1624, opened 2021-10-04
Upgrade to ember-auto-import v2

simonihmig:
This requires the app to also have eai v2. But should fix failing Canary builds (which now also requires eai v2, so requires our addon's dummy app to have eai v2 and webpack, see https://github.com/emberjs/ember.js/pull/19761). And enables us to eventually ship a native v2 addon without further breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "broccoli-funnel": "^3.0.8",
     "broccoli-merge-trees": "^4.2.0",
     "chalk": "^4.1.1",
-    "ember-auto-import": "^1.11.2",
+    "ember-auto-import": "^2.2.0",
     "ember-cli-babel": "^7.26.4",
     "ember-cli-build-config-editor": "0.5.1",
     "ember-cli-htmlbars": "^5.7.1",
@@ -143,7 +143,8 @@
     "qunit-dom": "^1.6.0",
     "release-it": "^14.9.0",
     "release-it-lerna-changelog": "^3.1.0",
-    "striptags": "^3.0.0"
+    "striptags": "^3.0.0",
+    "webpack": "^5.56.1"
   },
   "engines": {
     "node": "12.* || >= 14.*"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1150,6 +1150,19 @@
     semver "^7.3.2"
     typescript-memoize "^1.0.0-alpha.3"
 
+"@embroider/shared-internals@^0.40.0":
+  version "0.40.0"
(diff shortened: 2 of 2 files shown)


jelhan:
I'm wondering if it is too early to do that change. Ember CLI blueprints for new applications are still using `ember-auto-import@^1.11.3` for new applications: https://github.com/ember-cli/ember-new-output/blob/ce21320b131b4c54dee9b031353a8877194b6bd5/package.json#L32 It has been updated on `master` branch: https://github.com/ember-cli/ember-cli/blob/ebea201b9dfe451aa6ee8678c88e883a88e03628/blueprints/app/files/package.json#L35 I guess the plan is to ship it with Ember CLI v4. But shouldn't we at least wait until that one is official released? There is still the risk that it won't be shipped even with v4, isn't it?

simonihmig:
> I'm wondering if it is too early to do that change.

Too early for what? What's your specific concern?

Ember v4 (which is also an addon) already [requires it](https://github.com/emberjs/ember.js/blob/master/package.json#L62). And the only breaking change is that any app depending on an addon that requires eai v2 also has to update, but that's all. What's the risk?

jelhan:
> Ember v4 (which is also an addon) already [requires it](https://github.com/emberjs/ember.js/blob/master/package.json#L62).

Does it require it? I assumed/hoped that an Ember v4 application can use an addon, which depends on `ember-auto-import@v1`. Otherwise the upgrade pain would be very hard. But maybe I was too optimistic. 🤞

simonihmig:
The logic is as follows, see also https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md:
* if *any* addon requires eai2, then the app also has to have it
* an eai2 app can still depend on as many eai1 addons as you want

So it's not that all addons have to upgrade. But apps will likely "feel the pressure" to do so, once some addons do upgrade. One such "addon" is Ember 4. But upgrading should be easy and cheap, and unlocks future improvements. So I don't see a good reason to hold this back!?

jelhan:
> The logic is as follows, see also https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md:
> 
>     * if _any_ addon requires eai2, then the app also has to have it
> 
>     * an eai2 app can still depend on as many eai1 addons as you want
> 
> 
> So it's not that all addons have to upgrade. But apps will likely "feel the pressure" to do so, once some addons do upgrade. One such "addon" is Ember 4. But upgrading should be easy and cheap, and unlocks future improvements. So I don't see a good reason to hold this back!?

I see. Thanks a lot for the explanation.

I still fear the pain of being one of the first to do that change. Do we have any benefits from doing that change now beside avoiding another major only for that reason later?

PR #1600, opened 2021-08-17
Remove Bootstrap 3 support from components

simonihmig:
Some other related PRs will need to follow, e.g. for the default blueprint.

PR #1625, opened 2021-10-04
Reflect dropped BS 3 and added BS 5 support in readme

jelhan:
Updates documentation to reflect #1600 and #1494.


simonihmig (approved):
Good catch!

PR #1680, opened 2021-11-06
Update to `@ember/render-modifiers` v2

simonihmig:
The preceding renovate PR #1656 failed. The changes in https://github.com/emberjs/ember-render-modifiers/pull/42/files#diff-858a7ab2792f6dd402c28b973894abf3472e1119c6ad45b4e356626cf84c44c5 regarding autotracking seem to have caused the known GlimmerVM "same computation" assertion in `<BsCollape>` (and thus in others using it like `<BsAccordion>`) as well as `<BsAlert>`.

Ideally the `{{did-update}}` should probably be refactored to something else (resources maybe?), so this is just the short-term fix.

### DIFF
--- a/addon/components/bs-alert.js
+++ b/addon/components/bs-alert.js
@@ -75,6 +75,12 @@ export default class Alert extends Component {
   hidden = !this.visible;
 
   /**
+   * This is an unfortunate duplication of the previous property, but this is untracked to avoid causing the dreaded "same computation" assertion in GlimmerVM when reading a tracked property before setting it.
+   * @private
+   */
+  _hidden = !this.visible;
+
+  /**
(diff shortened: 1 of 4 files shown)


addon/components/bs-alert.js
```diff
@@ -185,7 +191,7 @@ export default class Alert extends Component {
    * @private
    */
   hide() {
-    if (this.hidden) {
```
simonihmig, line 188: Reading the tracked property before setting it a few lines below seemed to cause the assertion here.

jelhan (approved):
Looks good to me. But I think in mid-term we should refactor away from `{{did-update}}` helper and modifier.

simonihmig:
> But I think in mid-term we should refactor away from {{did-update}} helper and modifier.

Absolutely.

PR #1689, opened 2021-11-07
Make Bootstrap 5 the new default when adding the addon

simonihmig:
This is not a breaking change, as for existing apps the Bootstrap version is pinned in `ember-cli-build.js`.


jelhan (approved):
Looks good to me.

PR #1699, opened 2021-11-15
modal footer should not render a <form> element

### LINKED ISSUE #1427: `BsModalComponent` renders footer as a `form` tag
`BsModalComponent` renders footer as a `form` tag. This can be observed even in the simple modal demo here: https://www.ember-bootstrap.com/components/modal

1. Click Open Modal
1. Observe `<form class="modal-footer">` in the rendered markup

jelhan:
Closes #1427


simonihmig (approved):
LGTM!

I still wonder why we had that, probably some very old patterns of using a submit button in the footer. But as all tests are passing, we can happily get rid of this legacy cruft!

PR #1700, opened 2021-11-15
Upgrade Ember Auto Import to v2 and avoid prerender of unexisting route to fix docs app

jelhan:
- Ember Bootstrap requires Ember Auto Import v2 since #1624. Docs app was still using v1. This causes the build of docs app to fail.
- Route `/getting-started/bootstrap-4` has been removed in #1617. But the route was still included in the list of routes, which should be prerendered. This causes the production build of docs app to fail.


simonihmig (approved):
:+1: thanks!

PR #1706, opened 2021-11-18
Remove use of curly component invocation for link components

simonihmig:
We don't want to support curly invocation of `LinkTo`, so removing tests. Also `(query-params)` helpers seems to be dropped from Ember 4, so this is helping to get CI green again.

I think the change is fine, so I'll merge without review. The question is if we want to tag this PR as breaking? After all the yielded link-to components use Ember's original one, which is not supposed to be used with curlies anyway. Do we actually consider any use of curly invocation still supported (as long as Ember supports it? @jelhan thoughts?


jelhan:
> Do we actually consider any use of curly invocation still supported (as long as Ember supports it?

I think we should officially drop support for any curly component invocation in v5. A short note in the changelog should be enough.

PR #1718, opened 2022-01-07
Call onShow(n)/onHide/onHidden when controlling tooltips and popovers programmatically

### LINKED ISSUE #1438: `onShow` action is not called when controlling bs-popover visibility programmatically
`onShow`, `onShown` are not executed when controlling bs-popover visibility through the `visible` property.
Also `ContextualHelp._show()` is not executed. This means that workarounds for touch devices, hover state and transition are not applied.

I would expect the actions to be called also when controlling the bs-popover component programmatically. The same as when triggered by an event. (It did work before glimmer, anyway)

I think the cause is that `@visible` is passed on directly to popper as `@showHelp`.

Workaround
----
```hbs
<BsPopover @visible={{@visible}}>
  <!-- ... -->
  {{#if @visible}}
    {{did-insert this.onShow}}
  {{/if}}
</BsPopover>
```

bootstrap:info
----
Output from `ember bootstrap:info`:

```
Npm packages:
ember-bootstrap: ^4.3.0 -> 4.6.3
ember-cli: ~3.25.0 -> 3.25.0
bootstrap: ^4.5.3 -> 4.6.0
bootstrap-sass: n/a
ember-cli-sass: ^10.0.1 -> 10.0.1
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
bootstrapVersion: 4
importBootstrapFont: false
importBootstrapCSS: false
```

simonihmig:
Fixes #1438 

The implementation is not really beautiful, as we still rely on those render-modifiers/helpers. But a complete refactor was out of scope for this quick fix.

### DIFF
--- a/tests/integration/components/bs-popover-test.js
+++ b/tests/integration/components/bs-popover-test.js
@@ -299,7 +299,7 @@ module('Integration | Component | bs-popover', function (hooks) {
     );
     await click('#target');
     assert.ok(showAction.calledOnce, 'show action has been called');
-    assert.notOk(shownAction.calledOnce, 'show action has not been called');
+    assert.ok(shownAction.notCalled, 'show action has not been called');
     assert.dom('.popover').doesNotExist('popover is not visible');
   });
 
@@ -344,7 +344,7 @@ module('Integration | Component | bs-popover', function (hooks) {
     await click('#target');
     await click('#target');
     assert.ok(hideAction.calledOnce, 'hide action has been called');
-    assert.notOk(hiddenAction.calledOnce, 'hidden action has not been called');
+    assert.ok(hiddenAction.notCalled, 'hidden action has not been called');
     assert.dom('.popover').exists({ count: 1 }, 'popover is visible');
   });
 
--- a/tests/integration/components/bs-popover-test.js
+++ b/tests/integration/components/bs-popover-test.js
@@ -260,6 +260,94 @@ module('Integration | Component | bs-popover', function (hooks) {
     assert.dom('#wrapper .popover').exists({ count: 1 }, 'Popover exists in place.');
   });
 
+  test('it calls onShow/onShown actions when showing popover by event [fade=false]', async function (assert) {
+    let showAction = sinon.spy();
+    this.set('show', showAction);
+    let shownAction = sinon.spy();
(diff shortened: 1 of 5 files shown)


tests/integration/components/bs-popover-test.js
```diff
@@ -260,6 +260,94 @@ module('Integration | Component | bs-popover', function (hooks) {
...
+    showAction.returns(false);
+    this.set('show', showAction);
+    let shownAction = sinon.spy();
+    this.set('shown', shownAction);
+    await render(
+      hbs`<div id="target"><BsPopover @title="Dummy" @fade={{false}} @onShow={{this.show}} @onShown={{this.shown}} /></div>`
+    );
+    await click('#target');
+    assert.ok(showAction.calledOnce, 'show action has been called');
+    assert.notOk(shownAction.calledOnce, 'show action has not been called');
```
jelhan, line 302: I think it would be better to assert that it hasn't been called. Otherwise assertion would also pass if it has been called more than once.

```suggestion
    assert.ok(shownAction.notCalled, 'show action has not been called');
```

tests/integration/components/bs-popover-test.js
```diff
@@ -260,6 +260,94 @@ module('Integration | Component | bs-popover', function (hooks) {
...
+    this.set('hide', hideAction);
+    let hiddenAction = sinon.spy();
+    this.set('hidden', hiddenAction);
+    await render(
+      hbs`<div id="target"><BsPopover @title="Dummy" @fade={{false}} @onHide={{this.hide}} @onHidden={{this.hidden}} /></div>`
+    );
+    await click('#target');
+    await click('#target');
+    assert.ok(hideAction.calledOnce, 'hide action has been called');
+    assert.notOk(hiddenAction.calledOnce, 'hidden action has not been called');
```
jelhan, line 347: As above I would recommend to assert that it hasn't been called at all.
```suggestion
    assert.ok(hiddenAction.notCalled, 'hidden action has not been called');
```

jelhan (approved):
Thanks a lot for picking this up. Looks good to me. I especially like the increased test coverage. 👏

I'm only surprised that it does not remove another mechanism to react to changes of `@visible`.

PR #1736, opened 2022-01-28
Upgrade Ember to 4.1, drop support for Ember < 3.24

simonihmig:


.eslintrc.js
```diff
@@ -63,6 +62,14 @@ module.exports = {
       plugins: ['node'],
       extends: ['plugin:node/recommended'],
     },
+    {
+      // test files
+      files: ['tests/**/*-test.{js,ts}'],
+      extends: [
+        // @todo enable this while fixing linting errors in tests...
```
simonihmig, line 69: There were too many linting errors to fix in this PR. Should revisited this though!

config/ember-try.js
```diff
@@ -59,20 +59,6 @@ module.exports = async function () {
           // FAIL_ON_DEPRECATION: true,
         },
       },
-      {
-        name: 'ember-default-with-jquery',
```
jelhan, line 63: Do we implicitly drop support for using with Ember jQuery? I guess there isn't a real risk to break jQuery support?

jelhan (approved):

config/ember-try.js
simonihmig: No, I don't think so. This has been removed from the default blueprint. And it makes sense, since jQuery-support is now basically not a feature of Ember anymore. You can still use the user-land addon `@ember/jquery`, but I don't think we need to explicitly test anymore. Again, there is almost zero risk of breaking things, and we also don't test with any other possible JS library people might happen to use with Ember, so why jQuery...

package.json
```diff
@@ -46,8 +46,7 @@
     "*.js": "eslint --cache --fix"
   },
   "resolutions": {
-    "ember-cli-htmlbars/semver": "~7.3.0",
-    "ip-regex": "^2.1.0"
+    "@embroider/util": "^1.0.0"
```
simonihmig, line 49: Had to add this, to fix he Embroider scenarios, which are now running with Ember 4.1. See https://github.com/tildeio/ember-element-helper/pull/73#pullrequestreview-866490304

@jelhan maybe we need to postpone the release a bit, until a new version of element-helper is released, to not cause the same issue for users?

jelhan: I guess postponing the stable release might be good in that case. I fear it would otherwise give users of Ember Bootstrap the wrong impression that the release is broken and relying on Embroider is causing too much issues.

Maybe we could do another pre-release in the meantime and announce that this pre-release will very likely be the final one?

simonihmig: @jelhan done: https://github.com/kaliber5/ember-bootstrap/releases/tag/v5.0.0-rc.0

PR #1741, opened 2022-01-30
Implement fullscreen modal with BS breakpoints

Ayoahha:


addon/components/bs-modal.js
```diff
@@ -193,6 +193,20 @@ export default class Modal extends Component {
...
 
+    /**
+   *  Allows adding fullscreen mode for modals. It will
+   *  apply the `modal-fullscreen' class when using 'true' and
+   *  `modal-fullscreen-[x]-down` class when using BS breakpoints
+   *   ([x] = 'sm', 'md', 'lg', 'xl', 'xxl').
+   *  
+   * @property fullscreen
+   * @type {string}
+   * @default 'true'
```
jelhan, line 204: It must not have a default value. Otherwise every modal would be fullscreen unless developer overwrites it explicitly. This would break backward compatibility.

addon/components/bs-modal.js
```diff
@@ -267,7 +281,6 @@ export default class Modal extends Component {
    * @public
    */
   @arg
-  transitionDuration = 300;
```
jelhan, line 270: I guess you had removed this line by accident?

addon/components/bs-modal.js
```diff
@@ -278,7 +291,6 @@ export default class Modal extends Component {
    * @public
    */
   @arg
-  backdropTransitionDuration = 150;
```
jelhan, line 281: As per above, I assume you had removed this line by accident?

addon/components/bs-modal/dialog.hbs
```diff
@@ -13,9 +13,15 @@
...
-    <div class="modal-content"
-     tabindex="-1"
+  <div class="modal-dialog {{this.sizeClass}} {{if @centered "modal-dialog-centered"}} {{if @scrollable "modal-dialog-scrollable"}} 
+  {{#if @fullscreen}}
+    {{#if (bs-eq @fullscreen true)}}
+      modal-fullscreen
+    {{else}}
+      modal-fullscreen-{{@fullscreen}}-down
+    {{/if}}
+  {{/if}}" role="document">
```
jelhan, line 23: If I recall correctly, we are currently trying to prevent linebreaks in `class` attributes even if that is causing very long lines.

```suggestion
  <div class="modal-dialog {{this.sizeClass}} {{if @centered "modal-dialog-centered"}} {{if @scrollable "modal-dialog-scrollable"}} {{#if @fullscreen}}{{#if (bs-eq @fullscreen true)}}modal-fullscreen{{else}}modal-fullscreen-{{@fullscreen}}-down{{/if}}{{/if}}" role="document">
```

addon/components/bs-modal/dialog.hbs
```diff
@@ -13,9 +13,15 @@
...
+    <div class="modal-content" tabindex="-1"
```
jelhan, line 24: Any specific reason for which you removed the linebreak before `tabindex`? Or was that just by mistake?

tests/integration/components/bs-modal-test.js
```diff
@@ -312,5 +312,31 @@ module('Integration | Component | bs-modal', function (hooks) {
       await settled();
       assert.dom('.modal').doesNotExist();
     });
+
+    test('it passes along fullscreen class attribute', async function (assert) {
+      this.set('fullscreen', 'full');
```
jelhan, line 317: If I understand API documentation correctly, `full` is not a supported option. It should be `true` for this test case, isn't it?

```suggestion
      this.set('fullscreen', true);
```

tests/integration/components/bs-modal-test.js
```diff
@@ -312,5 +312,31 @@ module('Integration | Component | bs-modal', function (hooks) {
       await settled();
       assert.dom('.modal').doesNotExist();
     });
+
+    test('it passes along fullscreen class attribute', async function (assert) {
```
jelhan, line 316: Small proposal to improve wording for clarity:

```suggestion
    test('it allows to render modal in fullscreen using @fullscreen argument', async function (assert) {
```

jelhan: Would be great if you could add a test that none of the classes is applied if `@fullscreen` argument is not passed.

jelhan: It seems that this test is currently grouped under the `it animates opening and closing the modal`. Would be great if you could move it above into the main module.

addon/components/bs-modal.js
```diff
@@ -193,6 +193,20 @@ export default class Modal extends Component {
...
+   *  `modal-fullscreen-[x]-down` class when using BS breakpoints
+   *   ([x] = 'sm', 'md', 'lg', 'xl', 'xxl').
+   *  
+   * @property fullscreen
+   * @type {string}
+   * @default 'true'
+   * @public
+   */
+  @arg
+  fullscreen = 'true';
```
jelhan, line 208: This doesn't seem to be used at all as you are not accessing `{{this.fullscreen}}` in the template but only `{{@fullscreen}}`. I think it could be removed safely.

```suggestion
```

addon/components/bs-modal.js
```diff
@@ -193,6 +193,20 @@ export default class Modal extends Component {
   @arg
   scrollable = false;
 
+    /**
+   *  Allows adding fullscreen mode for modals. It will
+   *  apply the `modal-fullscreen' class when using 'true' and
```
jelhan, line 198: `true` was using quotes instead of backticks and therefore won't be formatted as code.

```suggestion
   *  apply the `modal-fullscreen' class when using `true` and
```

addon/components/bs-modal.js
```diff
@@ -193,6 +193,20 @@ export default class Modal extends Component {
...
+   *  `modal-fullscreen-[x]-down` class when using BS breakpoints
+   *   ([x] = 'sm', 'md', 'lg', 'xl', 'xxl').
+   *  
+   * @property fullscreen
+   * @type {string}
```
jelhan, line 203: Reading the description I think the type is either `boolean` or `string`.

```suggestion
   * @type {(Boolean|String)}
```

jelhan (changes requested):
Thanks a lot for providing this pull request!

Looks good overall to me. I only have noticed some minor points. Would be great if you could fix them as well.

@simonihmig Would be great if you could have a look at the proposed public API. I haven't found any similar case in our existing source code. Are you fine with having a union type (boolean or string) here?

addon/components/bs-modal.js
Ayoahha: Yes exact!!!

addon/components/bs-modal.js
Ayoahha: Yes exact!!!

addon/components/bs-modal/dialog.hbs
Ayoahha: OK. Fixed

addon/components/bs-modal/dialog.hbs
Ayoahha: I think this is my linter or beautifyer addon !

tests/integration/components/bs-modal-test.js
Ayoahha: Fixed

addon/components/bs-modal.js
Ayoahha: Fixed

addon/components/bs-modal.js
Ayoahha: OK, but in this case, some of the other comments have also quotes !
So we need to update all of these ?

addon/components/bs-modal.js
Ayoahha: Fixed

tests/integration/components/bs-modal-test.js
Ayoahha: 1. Fixed
2. I do not understand what you are asking...

addon/components/bs-modal.js
jelhan: It is okay to only address the new API docs added by this pull request here. I would recommend to fix other issues in separate pull requests to keep focus.

I just noticed that at the end of `modal-fullscreen` it is also a single quote and not a backtick. Here is an updated suggestion.

```suggestion
   *  apply the `modal-fullscreen` class when using `true` and
```

tests/integration/components/bs-modal-test.js
jelhan: The test suite uses `modules` to group related tests. The main module is declared very at the top of the file: https://github.com/kaliber5/ember-bootstrap/blob/e24988a19d7b21607ee7f386b49ce3f9b0159b0c/tests/integration/components/bs-modal-test.js#L11 But it contains an additional module declared later focused on the animation: https://github.com/kaliber5/ember-bootstrap/blob/e24988a19d7b21607ee7f386b49ce3f9b0159b0c/tests/integration/components/bs-modal-test.js#L251 You have put your new test within that module. It would be better if you could move your test out of that sub-module. So basically moving it above line 251.

Ayoahha: Exact. Understood
**Fixed**

addon/components/bs-modal.js
Ayoahha: Fixed

addon/components/bs-modal/dialog.hbs
jelhan: I think you must use inline `{{if}}` in this case. https://guides.emberjs.com/release/components/conditional-content/#toc_inline-if

Ayoahha: Yes it was my proposal... But it seems tricky here in this case...

addon/components/bs-modal.js
```diff
@@ -193,6 +193,18 @@ export default class Modal extends Component {
...
 
+  /**
+   *  Allows adding fullscreen mode for modals. It will
+   *  apply the `modal-fullscreen' class when using `true` and
+   *  `modal-fullscreen-[x]-down` class when using BS breakpoints
+   *   ([x] = 'sm', 'md', 'lg', 'xl', 'xxl').
+   *
+   * @property fullscreen
+   * @type {(Boolean|String)}
+   * @default true
```
jelhan, line 204: I think the default value is either `undefined` or `false`. If it would be `true`, the modal would be shown if `<BsModal>` is invoked without setting `@fullscreen` argument.

I guess documenting it as `false` provides most clarity for readers. That the default value actually used is `undefined`, seems to be an implementation detail on our side an user does not need to care about.

```suggestion
   * @default false
```

tests/integration/components/bs-modal-test.js
```diff
@@ -312,5 +312,35 @@ module('Integration | Component | bs-modal', function (hooks) {
       await settled();
       assert.dom('.modal').doesNotExist();
     });
+
+    test('it allows to render modal in fullscreen using @fullscreen argument', async function (assert) {
+      this.set('fullscreen', null);
+      await render(hbs`<BsModal>`);
+      assert.dom('.modal-dialog').doesNotHaveClass(/modal-fullscreen/);
```
jelhan, line 319: I like how you solved that with a RegExp. :clap:

addon/components/bs-modal.js
```diff
@@ -193,6 +193,18 @@ export default class Modal extends Component {
   @arg
   scrollable = false;
 
+  /**
+   *  Allows adding fullscreen mode for modals. It will
+   *  apply the `modal-fullscreen' class when using `true` and
+   *  `modal-fullscreen-[x]-down` class when using BS breakpoints
+   *   ([x] = 'sm', 'md', 'lg', 'xl', 'xxl').
```
jelhan, line 200: Would be great if you could add the link to the Bootstrap documentation for clarity: https://getbootstrap.com/docs/5.1/components/modal/#fullscreen-modal

addon/components/bs-modal/dialog.hbs
```diff
@@ -13,9 +13,8 @@
   {{did-insert this.getOrSetTitleId}}
   {{did-insert this.setInitialFocus}}
 >
-  <div class="modal-dialog {{this.sizeClass}} {{if @centered "modal-dialog-centered"}} {{if @scrollable "modal-dialog-scrollable"}}" role="document">
-    <div class="modal-content"
-     tabindex="-1"
+<div class="modal-dialog {{this.sizeClass}} {{if @centered "modal-dialog-centered"}} {{if @scrollable "modal-dialog-scrollable"}} {{if @fullscreen (if (bs-eq @fullscreen true) "modal-fullscreen" "modal-fullscreen-{{@fullscreen}}-down")}}" role="document">
```
jelhan, line 16: I just noticed that fullscreen modals are only supported by Bootstrap 5. Ember Bootstrap supports both Bootstrap 4 and 5. We need to exclude this code if Ember Bootstrap is used with Bootstrap 4.

We are using build-time macros provided by Embroider to do so: `{{macroCondition}}` to have dead-code elimination at build-time. `{{macroGetOwnConfig}}` to get the value to decide, which code path should be eliminated. Together it looks like this:

```hbs
{{macroCondition (macroGetOwnConfig "isBS5") "only-included-in-the-build-if-app-is-using-bootstrap-5"}}
```

You can see an example here: https://github.com/kaliber5/ember-bootstrap/blob/4b261eba2d6352eafdade589e4ba2a5a9f3e750a/addon/components/bs-button.hbs#L4

Please also update the API docs to mention that it is only supported for Bootstrap 5.

tests/integration/components/bs-modal-test.js
```diff
@@ -312,5 +312,35 @@ module('Integration | Component | bs-modal', function (hooks) {
       await settled();
       assert.dom('.modal').doesNotExist();
     });
+
+    test('it allows to render modal in fullscreen using @fullscreen argument', async function (assert) {
+      this.set('fullscreen', null);
+      await render(hbs`<BsModal>`);
```
jelhan, line 318: You need to close the `<BsModal>` component. Otherwise it is a syntax error, which breaks the build.

```suggestion
      await render(hbs`<BsModal>Hello World!</BsModal>`);
```

tests/integration/components/bs-modal-test.js
```diff
@@ -312,5 +312,35 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+      assert.dom('.modal-dialog').doesNotHaveClass(/modal-fullscreen/);
+
+      this.set('fullscreen', true);
+      await render(hbs`<BsModal @fullscreen={{this.fullscreen}}>`);
```
jelhan, line 322: No need to re-render again. The updated `{{this.fullscreen}}` value should be picked up automatically.

```suggestion
```

If it is not working, you may need to `await settled()` before to ensure Ember had time to pick up the state change and update the DOM. To be honest I'm still confused when `this.set()` in tests is triggering a sync re-render and when not.

It's the same for the other cases of course.

jelhan (changes requested):
I have done another review. Noticed that I missed some things in the first one. I'm sorry for raising up that many topics again.

addon/components/bs-modal.js
Ayoahha: Fixed

addon/components/bs-modal.js
Ayoahha: Fixed

addon/components/bs-modal/dialog.hbs
Ayoahha: Well done !
Fixed in doc and in hbs

tests/integration/components/bs-modal-test.js
Ayoahha: Of course !! My mistake

addon/components/bs-modal/dialog.hbs
jelhan:
```suggestion
    <div class="modal-content"
      tabindex="-1"
```

tests/integration/components/bs-modal-test.js
```diff
@@ -127,6 +127,31 @@ module('Integration | Component | bs-modal', function (hooks) {
     assert.dom('.modal').hasAttribute('aria-labelledby', modalTitleId);
   });
 
+  test('it allows to render modal in fullscreen using @fullscreen argument', async function (assert) {
```
jelhan, line 130: As that feature is supported in Bootstrap 5 only, we should also only run the test for BS5. We have an utility function called `testBS5` to do so: https://github.com/kaliber5/ember-bootstrap/blob/4b261eba2d6352eafdade589e4ba2a5a9f3e750a/tests/helpers/bootstrap.js#L26-L28 You could find an usage example here: https://github.com/kaliber5/ember-bootstrap/blob/a254a1ff8de05f0cc6fa5d51bbb92d3ad827b5ce/tests/integration/components/bs-button-group-test.js#L54

tests/integration/components/bs-modal-test.js
```diff
@@ -127,6 +127,31 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+    this.set('fullscreen', null);
+    await render(hbs`<BsModal>Hello World!</BsModal>`);
+    assert.dom('.modal-dialog').doesNotHaveClass(/modal-fullscreen/);
+
+    this.set('fullscreen', true);
+    await render(hbs`<BsModal @fullscreen={{this.fullscreen}}>Hello World!</BsModal>`);
```
jelhan, line 136: This `await render()` should not be needed.

```suggestion
```

jelhan (changes requested):
I'm sorry for coming back with another set of changes, which I missed in my previous review.

tests/integration/components/bs-modal-test.js
Ayoahha: Fixed

addon/components/bs-modal/dialog.hbs
```diff
@@ -1,28 +1,40 @@
 <div
-  role="dialog"
-  tabindex="-1"
+  role='dialog'
```
jelhan, line 2: Need to revert the quote style changes in this file. I guess this happened by running Prettier on the file but not having Prettier configured for Handlebars in the repository.

jelhan:
@simonihmig Thanks a lot for merging. I assumed that I had approved long ago. Seems I missed that. Currently in vacation and therefore haven't had the chance to follow up.

PR #1750, opened 2022-02-10
Needs peerDependency on ember-source

ef4:
`ember-bootstrap` depends on `@ember/render-modifiers`:

https://github.com/kaliber5/ember-bootstrap/blob/fdd5c22788ea0080e6e7bb83c20539c8048de367/package.json#L51-L52

And `@ember/render-modifiers` has a peerDependency on `ember-source`:

https://github.com/emberjs/ember-render-modifiers/blob/7dcc032d4f968c17fba6d10345550845f99cea1b/package.json#L73

Which means that `ember-bootstrap` should satisfies this peer dependency. Given that we know `ember-source` should only be provided by the top-level app, this means `ember-bootstrap` should list `ember-source` in `peerDependencies` (as opposed to listing it in `dependencies`).

## General observations

1. In principle, every ember addon really has an unstated peer dependency on ember-source. Historically we just let them "magically" get access even without declaring it.

    But in an effort to gradually move the ecosystem toward fully declaring their dependencies in a way that is legible to package managers, new features like `@embroider/macros` dependencySatisfies are strict about it.

2. Most apps will not experience any breakage either way. `@ember/render-modifiers` will typically be able to see the app's copy of `ember-source` because both will end up as peers in a single top-level `node_modules` directory. 

    So while a strict interpretation of the problem would suggest that adding a new peerDependency is a breaking change, and `@ember/render-modifiers` should have done a semver major when they added this peerDep and now `ember-bootstrap` should do a semver major also, I don't necessarily think that's super important. As a practical matter, virtually every existing package that uses ember-bootstrap *does* have access to ember-source, so adding this peerDep now is just a formalization of the status quo.

3. So why should we bother at all? The lack of this peer dep is a problem for people with more complicated setups, including monorepos with multiple different ember apps on different versions of ember, or people trying to adopt more strict package managers.

    I noticed the issue because this undeclared peerDep *does* break a test case in embroider's monorepo. It manifests as the dependencySatisfies macro in `@ember/render-modifiers` not being able to locate ember-source, because its parent package (ember-bootstrap) doesn't declare any dependency on it, and our test infrastructure only provides declared dependencies, not accidental ones.

4. A nice side-benefit of this pattern is that it creates a formalized place to declare your minimum supported Ember version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,9 @@
     "striptags": "3.2.0",
     "webpack": "5.68.0"
   },
+  "peerDependencies": {
+    "ember-source": ">=3.24"
+  },
   "engines": {
     "node": "12.* || 14.* || >= 16"
   },


jelhan (approved):
Thanks a lot for including this detailed explanation!

simonihmig:
> Thanks a lot for including this detailed explanation!

Yes, indeed! Thanks @ef4! 

So I am fine with merging this change, especially as it doesn't do any harm (I agree with the semver considerations from a practical standpoint). And I understand why it makes sense technically.

However I think that all of that applies equally to any other addon. So I wonder how this should be tackled for the whole ecosystem?

Other addons might depend on `@ember/render-modifiers`.  And other addons might use `dependencySatisfies` themselves, or transitively in a similar way. Or some addon introduces `dependencySatisfies` in a patch/minor release, without the parent addon noticing this added peerDependency, like in this case. Especially since Ember addons traditionally have not cared at all about peerDependencies, especially for ember-source.

So doesn't this basically mean that *all* addons should declare ember-source as a peerDependency? I.e. it should be added to the default addon blueprint? How would that affect the ecosystem if all those thousands of addon don't catch up on this?

Besides that, is there a possible alternative here? Could `dependencySatisfies` treat certain packages like ember-source in a special way, due to these historical reasons?

jelhan:
> So doesn't this basically mean that all addons should declare ember-source as a peerDependency? I.e. it should be added to the default addon blueprint? How would that affect the ecosystem if all those thousands of addon don't catch up on this?

I strongly agree to this concerns. Especially as not even the default blueprints for a new addon declare that peer dependency yet.

Before this gets a de facto requirement it should go through RFC process in my opinion. Not because I disagree with the direction. Actually I agree that it's the correct path forward. But RFC will help us to ensure that transition path and teaching is well thought through as well.

Not sure if this merge request is the best place to have this discussion though.

ef4:
> Besides that, is there a possible alternative here? Could dependencySatisfies treat certain packages like ember-source in a special way, due to these historical reasons?

It depends on the specific reason somebody is using `dependencySatisfies('ember-source', ...)`.  For example, in this case: 

https://github.com/emberjs/ember-render-modifiers/blob/7dcc032d4f968c17fba6d10345550845f99cea1b/addon/modifiers/will-destroy.js#L45

the point is to pass different values to Ember depending on the version. In that situation, we could offer a new dedicated macro like `emberVersionSatisfies` instead and that would work well. 

But in other cases, people want to import something out of Ember: 

https://github.com/emberjs/ember-render-modifiers/blob/7dcc032d4f968c17fba6d10345550845f99cea1b/addon/modifiers/did-update.js#L5

In these cases it's important that we keep `dependencySatisfies` and the imports aligned to mean the same thing. If we changed dependencySatisfies to mean "the app's copy, not the one resolvable from here", then we'd also have to guarantee that all imports work the same way. At the moment, maybe they even already do work that way! But those layers of compatibility hacks are things we're trying to kill off as quickly as possible. So I don't really want to introduce a new way of relying on them.

> So doesn't this basically mean that all addons should declare ember-source as a peerDependency?

Probably yes. The only thing that makes me hesitate is the interaction with our other compatibility challenges, like the fact that even though a peerDep on ember-source is true to the current *implementation*, it's not true to the *interface*: people are importing from paths like `@ember/component`, not paths like `ember-source/whatever`.

If we tell addons to start declaring peer dependencies we lock in a bunch of new long-term state pointing at the current implementation. Should they also declare a peer dep on `@glimmer/component` if they use it, since that is a real separate package? How should developers know to treat that differently than they treat `@ember/component`, which is not a real package? What are the version range constraints between your ember-source peerDep and your glimmer/component peer dep? It gets complicated.

This discussion and some other bugs I'm looking at make me think I should be pushing toward an [import-map](https://wicg.github.io/import-maps/) based solution sooner rather than later. It's clear that we need a layer of indirection between NPM and actual module resolution, because NPM is a mess *and* because I want to make true ES modules work **in the browser**. Right now that layer of indirection is ad-hoc within embroider, focused on trying to translate classic patterns to NPM-compatible patterns. But if we instead focused on translating both of those to import maps, we could then run in all environments that support import maps, which will hopefully be many.

Getting concrete, it would mean that both imports and dependencySatisfies would follow an import map to decide what they're resolving, so we'd have more control over the resolution and could make both do the right thing.

That's not without downsides. TypeScript for example [doesn't have support yet](https://github.com/microsoft/TypeScript/issues/43326), so we'd be making a gamble on them adopting. That kind of compatibility is the big reason to try to keep dealing with doing things "the NPM way", because it's so broadly supported by tools.

simonihmig:
Thanks again for the detailed explanations. Going to merge this now. And yes, further discussions about that topic should probably happen elsewhere! :)

PR #1759, opened 2022-02-21
drop @class argument of <BsTooltip> and <BsTooltip::Element>

jelhan:
As for the other cases the `@class` argument is not documented as public API. In this case a consumer can use `class` attribute instead.

The `<BsTooltip>` and `<BsTooltip::Element>` components are not yielded. So no need to keep it in order to set class on yielded component instance.


tests/integration/components/bs-tooltip-test.js
```diff
@@ -13,7 +13,6 @@ import {
 import { assertPositioning, setupForPositioning } from '../../helpers/contextual-help';
 import setupStylesheetSupport from '../../helpers/setup-stylesheet-support';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
-import { gte } from 'ember-compatibility-helpers';
```
simonihmig, line 16: :+1:

simonihmig (approved):

PR #1760, opened 2022-02-21
drop @class argument of <BsPopover> and <BsPopover::Element>

jelhan:
This seems to be the very same case as #1759


simonihmig:
@jelhan thanks for all the. cleanup work! 🙏

You would have to remove a still existing test for `@class` here, to make CI happy!

simonihmig (approved):

PR #1775, opened 2022-03-18
Convert BsForm to Glimmer component

### LINKED ISSUE #1338: Tests for validation support in <BsForm> do not test validation plugin interface

simonihmig:
Also fixes tests to test real validation plugin interface, closes #1338

### DIFF
--- a/addon/components/bs-form.hbs
+++ b/addon/components/bs-form.hbs
@@ -13,8 +13,8 @@
         formLayout=this.formLayout
         horizontalLabelGridClass=this.horizontalLabelGridClass
         showAllValidations=this.showAllValidations
-        _disabled=this.disabled
-        _readonly=this.readonly
+        _disabled=@disabled
+        _readonly=@readonly
         onChange=this.elementChanged
         _onChange=this.resetSubmissionState
       )
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -1,14 +1,14 @@
-import { tagName } from '@ember-decorators/component';
-import { gt } from '@ember/object/computed';
-import Component from '@ember/component';
-import { action, computed, set } from '@ember/object';
+import Component from '@glimmer/component';
+import { action, set } from '@ember/object';
 import { assert } from '@ember/debug';
 import { isPresent } from '@ember/utils';
 import { schedule } from '@ember/runloop';
 import RSVP from 'rsvp';
-import defaultValue from 'ember-bootstrap/utils/default-decorator';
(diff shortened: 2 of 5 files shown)


simonihmig:
@jelhan we might need to update `ember-bootstrap-cp-validations` and `ember-bootstrap-changeset-validations` after landing this I guess? For them it might be a breaking change, while users should not be affected as they are not allowed to subclass components. But as we consider them "privileged" addons, we won't bump a major here, I think we agreed on that, right?

simonihmig:
OK, due to the use of a component class (vs. component definition as with the `{{component}}` helper) tests are failing for older Ember. But I think I know how to fix. Also need to look into Embroider, but not today anymore...

simonihmig:
I tested the changes locally with `ember-bootstrap-cp-validations`, and everything seems to work fine. `ember-bootstrap-changeset-validations` however fails. It is still using the legacy object model: https://github.com/kaliber5/ember-bootstrap-changeset-validations/blob/master/addon/components/bs-form.js#L6 🙁

So until that is fixed, I'll keep this open...

jelhan:
> I tested the changes locally with `ember-bootstrap-cp-validations`, and everything seems to work fine. `ember-bootstrap-changeset-validations` however fails. It is still using the legacy object model: https://github.com/kaliber5/ember-bootstrap-changeset-validations/blob/master/addon/components/bs-form.js#L6 slightly_frowning_face
> 
> So until that is fixed, I'll keep this open...

Good catch. I created an issue in that repository to track: https://github.com/kaliber5/ember-bootstrap-changeset-validations/issues/39 Sadly I don't have time to pick it up right now.

simonihmig:
I'll merge this now, as the fix on the `ember-bootstrap-changeset-validations` side has been merged as well. Will then cut a release for both addons soon...

PR #1781, opened 2022-04-27
Update ember-style-modifier to 0.8.0

kdagnan:
This version fixes deprecations for ember 3.28. IE:
- deprecate.js:136 DEPRECATION: ember-modifier (in StyleModifier at Error
    at StyleModifier.get [as args] 
- DEPRECATION: ember-modifier (in StyleModifier at Error
    at new ClassBasedModifier

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "ember-popper-modifier": "^2.0.0",
     "ember-ref-bucket": "^4.0.0",
     "ember-render-helpers": "^0.2.0",
-    "ember-style-modifier": "^0.7.0",
+    "ember-style-modifier": "^0.8.0",
     "findup-sync": "^5.0.0",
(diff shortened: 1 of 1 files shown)


simonihmig:
Thank you @kdagnan! This is failing CI, because you didn't update the `yarn.lock` file! You would need to run yarn locally to update it and commit it.

Usually we get these type of dependency updates automatically through our renovate bot. But in this case it was rate limited due to all other renovate PRs being blocked due to failing CI. It seems this is the case exactly because of not updating this modifier, which causes a deprecation that fails our CI (we have set this up intentionally, to make sure we don't cause any deprecations). 

I just asked renovate to forcefully create the `ember-style-modifier` update. If you manage to update your PR first, I'll gladly merge. Otherwise will merge the other.

PR #1788, opened 2022-05-08
Add additional tests for arguments of BsModal

### LINKED ISSUE #1742: Missing test coverage for arguments of <BsModal>
We seem to miss test coverage for some arguments supported by `<BsModal>`. I only had a very brief look but it seems to miss tests for at least the following arguments:

- `@size`
- `@centered`
- `@scrollable`

bvedad:
Tests added for the following arguments of BsModal:
- size
- position
- scrollable
- keyboard
- backdropClose

**TODO** later we can add tests for `transitionDuration` and `backdropTransitionDuration`.

Fixes #1742

### DIFF
--- a/tests/integration/components/bs-modal-test.js
+++ b/tests/integration/components/bs-modal-test.js
@@ -356,4 +356,15 @@ module('Integration | Component | bs-modal', function (hooks) {
     this.set('position', 'center');
     assert.dom('.modal-dialog').hasClass('modal-dialog-centered');
   });
+
+  test('shows scrollable modal with respective to @scrollable argument', async function (assert) {
+    this.set('scrollable', false);
+    await render(hbs`
+      <BsModal @scrollable={{this.scrollable}} />
+    `);
+    assert.dom('.modal').exists({ count: 1 }, 'Modal exists.');
+    assert.dom('.modal-dialog').hasNoClass('modal-dialog-scrollable');
+    this.set('scrollable', true);
+    assert.dom('.modal-dialog').hasClass('modal-dialog-scrollable');
+  });
 });
--- a/tests/integration/components/bs-modal-test.js
+++ b/tests/integration/components/bs-modal-test.js
@@ -1,7 +1,7 @@
 import Component from '@ember/component';
 import { module } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { click, find, render, settled, triggerEvent, waitFor, waitUntil } from '@ember/test-helpers';
+import { click, find, render, settled, triggerEvent, waitFor, waitUntil, triggerKeyEvent } from '@ember/test-helpers';
 import { test, testBS5, visibilityClass } from '../../helpers/bootstrap';
 import hbs from 'htmlbars-inline-precompile';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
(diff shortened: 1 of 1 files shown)


tests/integration/components/bs-modal-test.js
```diff
@@ -331,4 +331,76 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+    await triggerKeyEvent('.modal', 'keydown', 'Escape');
+    assert.dom('.modal').exists({ count: 1 }, 'Modal exists.');
+  });
+
+  test('Modal closes when click is done outside the modal if @backdropClose=true', async function (assert) {
+    await render(hbs`
+      <BsModal @backdropClose={{true}} />
+    `);
+    assert.dom('.modal').exists({ count: 1 }, 'Modal exists.');
+    await click('*');
```
jelhan, line 394: The selector `*` should trigger a click on all events. Or just on the first one, which is matched by the selector? I'm not sure. But I think `this.element` is more precise in any case.

```suggestion
    await click(this.element);
```

tests/integration/components/bs-modal-test.js
```diff
@@ -331,4 +331,76 @@ module('Integration | Component | bs-modal', function (hooks) {
...
+    assert.dom('.modal').doesNotExist('Modal does not exist.');
+  });
+
+  test('Modal does not close when click is done outside the modal if @backdropClose=false', async function (assert) {
+    await render(hbs`
+      <BsModal @backdropClose={{false}} />
+    `);
+    assert.dom('.modal').exists({ count: 1 }, 'Modal exists.');
+    await click('*');
```
jelhan, line 403: As above.

```suggestion
    await click(this.element);
```

jelhan (changes requested):
Thank a lot for taking care of this one. Just a minor adjustment.

jelhan:
> TODO later we can add tests for transitionDuration and backdropTransitionDuration.

Would be great if you could create an additional issue targeting that two. I _think_ I had created similar tests for a component in Ember Bootstrap. But not fully sure anymore.

jelhan:
> > TODO later we can add tests for transitionDuration and backdropTransitionDuration.
> 
> Would be great if you could create an additional issue targeting that two. I _think_ I had created similar tests for a component in Ember Bootstrap. But not fully sure anymore.

I think I recalled these tests: https://github.com/kaliber5/ember-bootstrap/blob/cf57ff67b775a8f12f3f46bf21f438c5c46af06d/tests/integration/components/bs-modal-test.js#L269-L333 Not covering the transition duration yet. But might be a starting point to write tests for it.

tests/integration/components/bs-modal-test.js
bvedad: `this.element` does not work, I have just tried it. `await click("*")` will trigger click event only on the first element which is matched by the selector.

We have already used this approach when closing the Bootstrap dropdown:
https://github.com/ember-bootstrap/ember-bootstrap/blob/master/tests/integration/components/bs-dropdown-test.js#L110

tests/integration/components/bs-modal-test.js
bvedad: https://github.com/ember-bootstrap/ember-bootstrap/pull/1788#discussion_r939671150

jelhan (approved):
Looks good to me. Thanks a lot!

PR #1790, opened 2022-05-11
Bug fix: carousel component missing from API docs

### LINKED ISSUE #1301: `<Carousel>` component is missing in API docs
The `<Carousel>` component is missing in latest API docs. Not exactly sure since when.

bvedad:
Fixes #1301


jelhan (approved):
Good catch. Thanks a lot for fixing!

Issue #1792, opened 2022-05-25
[RFC] Support form validation using native constraint validation API


simonihmig:
## Introduction

Currently we support form validation through additional integration addons, either using `ember-cp-validations` or `ember-changeset-validations`. These are entirely JavaScript-based validations, but come with their own drawbacks:
* the former is basically unmaintained, not compatible with Ember 4, and not in line with Octane idioms. Having that addon still in apps is basically a big issue of tech debt.
* the latter is coupled to the concept of changesets, which is not always what you want to use (e.g. you have a form component, with the form (model) backed by local state of the component itself, so you don't necessarily need a changeset to avoid mutating global state)

However there is also validation built into "the platform", in the form of HTML5 [constraint validation](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation). And support for that is actually build into Bootstrap (v5) itself, in two ways:
* using [browser defaults](https://getbootstrap.com/docs/5.2/forms/validation/#browser-defaults)
* with [custom styles](https://getbootstrap.com/docs/5.2/forms/validation/#custom-styles)

The former is easily supported already, as it does not require any additional JavaScript functionality. The intend of this "RFC" is to support the latter.

## Detailed Design

Bootstrap supports native validation with custom styles through special CSS pseudo selectors like `:invalid`, which will show the validation messages *when* an additional class `.was-validated` has been applied (to prevent showing them before the user has entered any data). 

While we *could* support that markup schema, it would mean we have to change the existing validation markup (what Bootstrap refers to as [server side markup](https://getbootstrap.com/docs/5.2/forms/validation/#server-side), like *not* applying the `.is-invalid` class when relying on `:invalid`, but only when using native validation. But given that we already have validation markup that works in both scenarios (native + JS-based validation), we can save the additional effort and complexity, and re-use what we already have.

What is needed now is "just" an additional validation integration addon to make `<BsForm>` aware of the native validation capabilities. It will follow the same semantics like the other addons:
* show all validations on submit
* show "local" validation on focusOut.
* generate validation messages automatically. In this case this means:
  * not to show a message for successful validation (unlike "Looks good!" in the [Bootstrap example](https://getbootstrap.com/docs/5.2/forms/validation/#custom-styles)). But this is the same as with the other validation integrations
  * using as the error message whatever the browser API ([`validationMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation#validity)) gives

[...]

PR #1793, opened 2022-05-25
Add form element to arguments passed to validate hook

simonihmig:
See https://github.com/kaliber5/ember-bootstrap/issues/1792

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -9,6 +9,7 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 import arg from '../utils/decorators/arg';
 import { tracked } from '@glimmer/tracking';
 import { cached } from 'tracked-toolbox';
+import { ref } from 'ember-ref-bucket';
 
 /**
   Render a form with the appropriate Bootstrap layout class (see `formLayout`).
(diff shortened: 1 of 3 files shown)


jelhan (commented):
Overall this looks good to me and I don't have concerns adding another argument on `validate`. It is not a public API. We have already changed it without major version as we have control over the validation addons using it.

I only have one question if we couldn't simplify the implementation.

simonihmig:
> I only have one question if we couldn't simplify the implementation.

@jelhan in which way?

addon/components/bs-form.js
```diff
@@ -400,7 +409,7 @@ export default class Form extends Component {
 
     return RSVP.resolve()
       .then(() => {
-        return this.hasValidator ? this.validate(model) : null;
+        return this.hasValidator ? this.validate(model, this._element) : null;
```
jelhan, line 412: Somehow my review comment got lost. :cry:

I'm wondering if we could pick up the `form` element from the `event` instead of tracking it through `{{ref}}`. The event listener is registered on the `form` element. Shouldn't the `form` element be available as `event.target` in that case?

simonihmig: Yes, indeed I had it this way, until I realized some existing test failed. It was one where we submit the form using the yielded submit action, and *not* a submit (browser) event. In which case we don't have `event.target`. That's why I also added the second test in this PR, which is replicating this very situation!

jelhan: Good catch. :clap:

PR #1808, opened 2022-08-18
Add Bootstrap List Group Component

onechiporenko:
Add support for List Group Component ([Bootstrap Docs](https://getbootstrap.com/docs/5.2/components/list-group/)).


addon/components/bs-list-group/item.hbs
```diff
@@ -0,0 +1,11 @@
+{{#let (element this.htmlTag) as |Tag|}}
+  <Tag
+    class="list-group-item {{if @type (bs-type-class "list-group-item" @type default="")}} {{if @action "list-group-item-action"}} {{if @disabled "disabled"}} {{if @active "active"}}"
+    aria-current="{{if @active true}}"
+    aria-disabled="{{if @disabled true}}"
```
simonihmig, line 5: :+1:

addon/components/bs-list-group/item.js
```diff
@@ -0,0 +1,80 @@
...
+   * `htmlTag` should be `a` or `button` in this case.
+   *
+   * Also see the [Bootstrap Docs](https://getbootstrap.com/docs/5.2/components/list-group/#links-and-buttons)
+   *
+   * @default false
+   * @type boolean
+   * @public
+   */
+  @arg
+  action = false;
```
simonihmig, line 55: I wonder if the naming might cause confusion here, like with `{{action}}` or a callback "action" as in DD**A**U?

Maybe `actionable` would be better?

addon/components/bs-list-group/item.js
```diff
@@ -0,0 +1,80 @@
...
+   * Setting its value to `button` will automatically add the attribute `type="button"`.
+   *
+   * This default attributes can be overridden.
+   *
+   * @default 'li'
+   * @type string
+   * @public
+   */
+  @arg
+  htmlTag = 'li';
```
simonihmig, line 27: Do we need to make the user specify this? Or can't we make this explicit, to be e.g. `button` when `@action={{true}}`, `li` otherwise? Seems for actionable items we force the user to set both, which seems redundant. It plain Bootstrap that's how it is with static markup, we can apply smartness into our component here, right?

Also, is there any good reason to use an `<a>` instead of a `<button>`? I feel if the link has no real `href`, it shouldn't be a link (e.g. a11y-wise)?

docs/app/templates/demo/list-group.hbs
```diff
@@ -0,0 +1,242 @@
...
+    <ListGroup.item>A third item</ListGroup.item>
+  </BsListGroup>
+  {{! END-SNIPPET }}
+</div>
+<div class="highlight">
+  <CodeSnippet @name="list-group-active-demo.hbs" />
+</div>
+
+<h2>Links and buttons</h2>
+<p>Use attribute <code>htmlTag</code> equal to <code>a</code> or <code>button</code> to create actionable list group items with hover, disabled, and active states by adding <code>action</code> attribute.</p>
```
simonihmig, line 63:
```suggestion
<p>Use attribute <code>htmlTag</code> equal to <code>a</code> or <code>button</code> to create actionable list group items with hover, disabled, and active states by adding the <code>action</code> argument.</p>
```

Let's call`@...` arguments, vs. HTML attributes.

docs/tests/unit/routes/demo/list-group-test.js
```diff
@@ -0,0 +1,11 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Route | demo/list-group', function(hooks) {
+  setupTest(hooks);
+
+  test('it exists', function(assert) {
+    let route = this.owner.lookup('route:demo/list-group');
+    assert.ok(route);
+  });
```
simonihmig, line 10: Probably auto-generated, but this is useless, so let's remove please.

tests/integration/components/bs-list-group/item-test.js
```diff
@@ -0,0 +1,134 @@
...
+
+  test('item can be disabled', async function (assert) {
+    await render(hbs`<BsListGroup as |ListGroup|>
+      <ListGroup.item class="first" @disabled={{true}}>First</ListGroup.item>
+      <ListGroup.item class="second">Second</ListGroup.item>
+    </BsListGroup>`);
+    assert
+      .dom('ul li.list-group-item.first')
+      .hasClass('disabled', 'First item has class disabled')
+      .hasAria('disabled', 'true', 'First item has aria-attr disabled');
```
simonihmig, line 38: `hasAria`? TIL, I didn't know about that one :)

simonihmig (commented):
Stunning work @onechiporenko, thanks a lot! 😍

And sorry for the very late review. 😔

Just a few debatable questions, otherwise this is almost flawless! :+1:

onechiporenko:
IDK what is wrong with tests for `bs-tooltip`.

However, it's much more interesting case with [3.24](https://github.com/ember-bootstrap/ember-bootstrap/actions/runs/3174729036/jobs/5171918253).
"Fail" can be isolated as:

```hbs
{{! app/components/some-btn.hbs}}
<button
  type="button"
  ...attributes>
  {{yield}}
</button>
```

```hbs
{{! application.hbs}}
<SomeBtn type="submit">Test</SomeBtn>
{{outlet}}
```

`type` will be overridden for 3.28+ (I didn't check for 3.25-3.27, because they are not lts), but not for 3.24 :/

jelhan:
> However, it's much more interesting case with [3.24](https://github.com/ember-bootstrap/ember-bootstrap/actions/runs/3174729036/jobs/5171918253). "Fail" can be isolated as:
> 
> ```
> {{! app/components/some-btn.hbs}}
> <button
>   type="button"
>   ...attributes>
>   {{yield}}
> </button>
> ```
> 
> ```
> {{! application.hbs}}
> <SomeBtn type="submit">Test</SomeBtn>
> {{outlet}}
> ```
> 
> `type` will be overridden for 3.28+ (I didn't check for 3.25-3.27, because they are not lts), but not for 3.24 :/

This is a known bug in older Ember versions if I recall correctly. :cry:

jelhan:
> IDK what is wrong with tests for `bs-tooltip`.

Should be fixed via #1814 by @simonihmig. You need to merge latest master in to fix CI.

Regarding tests, which are failing due to bug in Ember 3.24: I would recommend to skip them for that Ember version. You could use an approach like this: https://github.com/ember-bootstrap/ember-bootstrap/blob/5f79e8dca685aa2376e008d77bd786f61a0b0ae5/tests/integration/components/bs-button-test.js#L245-L247

PR #1809, opened 2022-08-20
Fix submitOnEnter when multiple input elements are present

elgordino:
This PR resolves an issue where 'submitOnEnter' would not fire when more than one input element was present in the form.

For example

```
<BsForm
    @model={{hash field1='x' field2='y'}}
    @formLayout='horizontal'
    @onSubmit={{perform this.submit}}
    @submitOnEnter={{true}} as |bsForm|>
    <bsForm.element @property="field1" @label="field1" />
    <bsForm.element @property="field2" @label="field2" />
</BsForm>
```

Would not call 'this.submit' when enter was pressed on one of the input elements.

The original issue arose in ember-bootstrap 4.0.1.

This PR resolves the issue by directing the `submitEvent` to the form, rather than the triggering input element. This effectively restores the behavior in ember-bootstrap 4.0.0 and earlier.

Interestingly the test case I have added failed with both a single <form.element> and with two <form.element>s. After the fix  the test case passes with both one and two elements, so I have left it in place with two elements as this is what was manifesting the problem.

Please let me know if any changes are desired?

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -483,7 +483,7 @@ export default class Form extends Component {
     if (code === 13 && this.args.submitOnEnter) {
       let submitEvent = document.createEvent('Event');
       submitEvent.initEvent('submit', true, true);
-      event.target.dispatchEvent(submitEvent);
+      event.target.closest('form').dispatchEvent(submitEvent);
     }
   }
 
--- a/tests/integration/components/bs-form-test.js
+++ b/tests/integration/components/bs-form-test.js
@@ -997,6 +997,21 @@ module('Integration | Component | bs-form', function (hooks) {
     assert.ok(submit.calledOnce, 'onSubmit action has been called');
   });
 
+  testRequiringFocus(
+    'Pressing enter on a form with multiple input elements and submitOnEnter submits the form',
+    async function (assert) {
+      let submit = sinon.spy();
+      this.actions.submit = submit;
+      await render(hbs`
+        <BsForm @onSubmit={{action "submit"}} @submitOnEnter={{true}} as |form|>
+          <form.element @property="input1" />
+          <form.element @property="input2" />
+        </BsForm>`);
+      await triggerKeyEvent('input', 'keypress', 13);
+      assert.ok(submit.calledOnce, 'onSubmit action has been called');
+    }
+  );
+
   test('prevents submission to be fired concurrently', async function (assert) {
     let deferredSubmitAction = defer();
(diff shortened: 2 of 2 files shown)


jelhan (approved):
Looks good to me. Thanks a lot for fixing it!

> Interestingly the test case I have added failed with both a single <form.element> and with two <form.element>s. After the fix the test case passes with both one and two elements, so I have left it in place with two elements as this is what was manifesting the problem.

I had a quick look at existing test case "Pressing enter on a form with submitOnEnter submits the form". It simulates a key press on `<form>` rather than `<input>`. This seems to be wrong.

I don't think a `<form>` element could have focus itself. So there might not be any realistic case in which user could trigger key event on that element.

I guess that test would have catched the error as well. If it would be correct.

@simonihmig do I miss something? Or should we adjust that test case as well?

addon/components/bs-form.js
```diff
@@ -483,7 +483,7 @@ export default class Form extends Component {
     if (code === 13 && this.args.submitOnEnter) {
       let submitEvent = document.createEvent('Event');
       submitEvent.initEvent('submit', true, true);
-      event.target.dispatchEvent(submitEvent);
+      event.target.closest('form').dispatchEvent(submitEvent);
```
simonihmig, line 486: I think we can just invoke `this.handleSubmit` here, can't we? This component catched the key press event bubbling up, so it is safe to assume that the component itself is the closest form, isn't it?

simonihmig (commented):
Sorry for the late response here, I didn't really pay attention much to my OSS "duties" the last weeks/months...

> do I miss something? Or should we adjust that test case as well?

Yes, I think we should do so. I guess historically the reason for this was that I was assuming the input event would bubble to the form, so the test just triggered it on the form.

I haven't really dug into it, but what change has actually caused this behavior to not work anymore. And AFAIU it is not related to two inputs, it just didn't work at all anymore, right?

jelhan:
@elgordino Do you have time to rebase and address the latest review comment from simonihmig?

jelhan:
@elgordino Please let me know if I should take over. Would love getting this in the next release.

jelhan:
Seems that I'm not allowed to push to the PR. Will create a new PR cherry-picking your fixes.

jelhan:
Superseded by #1911

elgordino:
Hi @jelhan sorry I totally missed your previous messages. Thanks for picking this up and merging it.

PR #1811, opened 2022-08-24
Yield form's model from <BsForm>

### LINKED ISSUE #1806: Yield form's model from <BsForm>
In some cases you may need to access the model used for `<BsForm>`. E.g. if you want to show a value of a form's property or getter in the form.

This could be currently done by using the same object as passed to `<BsForm>` as model:

```hbs
<BsForm @model={{this.someObject}} as |form|>
  <form.element @property="foo" />
  Current value of <code>foo</code> is {{this.someObject.foo}}
</BsForm>
```

Wondering if we should yield the form's model from `<BsForm>` as `form.model` to make this more readable:

```hbs
<BsForm @model={{this.someObject}} as |form|>
  <form.element @property="foo" />
  Current value of <code>foo</code> is {{form.model.foo}}
</BsForm>
```

This would also ease migration to decoupling form's model from default values. I proposed this in #1551 last year but haven't had the time to follow up. Providing access to form's model in `<BsForm>` block even if it is constructed and managed internally would be a first step towards that direction.

bvedad:
model yielded from form's <BsForm>
Closes #1806


jelhan (approved):
Thanks a lot for picking this one up. Looks good to me.

@simonihmig You haven't commented on that proposed new public API. Any concerns from your side?

simonihmig (commented):
No concerns, thanks a lot @bvedad!

But we should document this somewhere. We don't really have a perfect place, but maybe somewhere together with the submission state [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/f48854e6d72759cd61de6874eaa945e74b55202c/addon/components/bs-form.js#L79-L110)?

jelhan:
> But we should document this somewhere. We don't really have a perfect place, but maybe somewhere together with the submission state [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/f48854e6d72759cd61de6874eaa945e74b55202c/addon/components/bs-form.js#L79-L110)?

Agree. It doesn't fit anywhere very well. The form's model may deserves its own chapter in long-term. There are many conceptual points of it, which aren't explicitly mentioned anywhere. For the time being, I wouldn't care much where we mentioned that it is yielded as well.

jelhan:
@bvedad CI should be fixed by #1814. Would be great if you could rebase or merge latest master in.

Please let me know if you don't have time for adding the documentation. I can do it in a follow-up PR in that case.

jelhan:
@bvedad Do you have time to rebase this one and add docs? Otherwise I can take over. Would love to get it in the next release.

PR #1814, opened 2022-10-01
Fix CI for BS5

simonihmig:
Bootstrap 5.2 seems to have introduced some CSS changes to tooltips that make our tooltip positioning tests fail, as they are very coupled to CSS layout.

### DIFF
--- a/tests/integration/components/bs-tooltip-test.js
+++ b/tests/integration/components/bs-tooltip-test.js
@@ -9,6 +9,7 @@ import {
   tooltipArrowClass,
   tooltipPositionClass,
   visibilityClass,
+  versionDependant,
 } from '../../helpers/bootstrap';
 import { assertPositioning, setupForPositioning } from '../../helpers/contextual-help';
 import setupStylesheetSupport from '../../helpers/setup-stylesheet-support';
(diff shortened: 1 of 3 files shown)


tests/integration/components/bs-tooltip-test.js
```diff
@@ -9,6 +9,7 @@ import {
   tooltipArrowClass,
   tooltipPositionClass,
   visibilityClass,
+  versionDependant,
```
jelhan, line 12: Small typo causing it to fail:

```suggestion
  versionDependent,
```

tests/integration/components/bs-tooltip-test.js
```diff
@@ -390,7 +391,7 @@ module('Integration | Component | bs-tooltip', function (hooks) {
   test('should position tooltip arrow centered', async function (assert) {
     this.insertCSSRule('.margin-top { margin-top: 200px; }');
 
-    let expectedArrowPosition = 94;
+    let expectedArrowPosition = versionDependant(94, 100);
```
jelhan, line 394: Per above

```suggestion
    let expectedArrowPosition = versionDependent(94, 100);
```

tests/integration/components/bs-tooltip-test.js
```diff
@@ -420,7 +421,7 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     this.insertCSSRule('.margin-top { margin-top: 200px; }');
     this.insertCSSRule('#target { width: 100px; padding: 0; border: none; }');
 
-    let expectedArrowPosition = 144;
+    let expectedArrowPosition = versionDependant(144, 156);
```
jelhan, line 424: Per above

```suggestion
    let expectedArrowPosition = versionDependent(144, 156);
```

jelhan (changes requested):
Thanks a lot for addressing this one. I fear a small typo made it fail.

PR #1815, opened 2022-10-03
remove branch in tests for Ember version out of support

jelhan:
We are not supporting Ember < 3.4 anymore. Thought I quickly clan up while I was coming across it.


simonihmig (approved):
:+1:

PR #1826, opened 2022-11-08
support ember-style-modifier v1, v2 and v3

jelhan:


jelhan:
This is currently blocked because ember-style-modifier v1 dropped support for node 12. Ember Bootstrap still supports node 12. And uses it in CI. I haven't found a way yet to configure yarn to pick `0.8.0` from the version range. It always uses the latest.

Also not sure if it would be a breaking change. Must all versions within a supported version range support node 12? Or is it enough if one version supports node 12?

simonihmig:
Sorry for the delay... we haven't done that much in the 5.x series, but given we need to drop node 12 at some point anyway, we can do so now and release this as 6.0?

jelhan:
We will likely see another major release of ember-style-modifier to address `@ember/string` deprecation soonish. We may want to wait until that one is landed and include it in the version range as well.

package.json
```diff
@@ -46,7 +46,8 @@
     "*.js": "eslint --cache --fix"
   },
   "resolutions": {
-    "ember-cli-htmlbars/semver": "~7.3.0"
+    "ember-cli-htmlbars/semver": "~7.3.0",
+    "ember-modifier": "3.2.7"
```
jelhan, line 50: Tests were failing with `ember-modifier@^4`. Pinning to v3 for now. Will open a GitHub issue for it. Some issues with `ember-modifier@^4` were reported in Discord channel as well.

simonihmig (approved):

PR #1839, opened 2023-01-07
drop support for Ember 3.24

jelhan:
I feel we should drop support for Ember 3.24 in next major release. It is likely that polyfills etc support Ember 3.28 than supporting Ember 3.24 going forward. This will give us more flexibility. Applications can continue using Ember Bootstrap even if not being ready for Ember v4 upgrade. They just need to use latest minor release of the 3.x series.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,6 @@ jobs:
         allow-failure:
           - false
         scenario:
-          - ember-lts-3.24
           - ember-lts-3.28
           - ember-release
           - ember-beta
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -61,7 +61,6 @@ jobs:
         allow-failure:
           - false
         scenario:
-          - ember-lts-3.24
           - ember-lts-3.28
```
simonihmig, line 64: We could have also just run `ember-cli-update` here, which would both drop 3.24 and node 12. Anyway, separate PRs is even better, but then we should run `ember-cli-update` nevertheless, especially to update the test matrix here. We should have the 4.x LTS releases covered here. All of which can happen in a separate PR of course...

simonihmig (approved):

.github/workflows/ci.yml
jelhan: I prefer to have breaking changes in separate PRs as it makes it easier for consumers to review. I can run `ember-cli-update` later. :+1:

PR #1844, opened 2023-01-11
fix HTML markup for <BsAccordion> per Bootstrap docs

### LINKED ISSUE #1843: Failing CI on master
CI on master is failing, due to accordion triggering an axe a11y failure. Idk how this change (axe update?) sneaked into master without failing CI first...

jelhan:
This updates the markup for `<BsAccordion>` accordingly to Bootstrap documentation:

- for Bootstrap v5: https://getbootstrap.com/docs/5.3/components/accordion/
- for Bootstrap v4: https://getbootstrap.com/docs/4.6/components/collapse/#accordion-example

Neither of them are using `role="tab"`, which is causing the test to fail on latest `master` branch. See #1843 for context.

For Bootstrap v4 there were more mismatches between the markup rendered by Ember Bootstrap and the markup provided in Bootstrap docs. The fixes for those mismatches are breaking changes:

1. The title is rendered as a `h2`. Before it was a `h5`.
2. Only the title is clickable to expand / collapse the content of an accordion item. Before it was the entire header.
3. Attributes set on `<Accordion::Item::Title>` are applied to `h2`. Before they were applied to `.card-header`.

Closes #1843

### DIFF
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -19,9 +19,11 @@
 {{else}}
   <div
     class="card-header"
-    ...attributes
   >
-    <h2 class="mb-0">
+    <h2
+      class="mb-0"
+      ...attributes
+    >
       <button
         class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}"
         type="button"
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -3,13 +3,14 @@
 {{#if (macroCondition (macroGetOwnConfig "isBS5"))}}
   <h2
     class="accordion-header"
-    role="tab"
     ...attributes
   >
     <button
       class="accordion-button {{if @disabled "disabled"}} {{if @collapsed "collapsed"}}"
       type="button"
       disabled={{@disabled}}
+      aria-controls={{@controls}}
+      aria-expanded={{if @collapsed "false" "true"}}
       {{on "click" this.handleClick}}
     >
       {{yield}}
@@ -18,12 +19,17 @@
 {{else}}
   <div
     class="card-header"
-    role="tab"
     ...attributes
-    {{on "click" this.handleClick}}
   >
     <h5 class="mb-0">
-      <button class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}" type="button" disabled={{@disabled}}>
+      <button
+        class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}"
+        type="button"
+        disabled={{@disabled}}
+        aria-controls={{@controls}}
+        aria-expanded={{if @collapsed "false" "true"}}
+        {{on "click" this.handleClick}}
+      >
         {{yield}}
       </button>
     </h5>
--- a/addon/components/bs-accordion/item/title.hbs
+++ b/addon/components/bs-accordion/item/title.hbs
@@ -21,7 +21,7 @@
     class="card-header"
     ...attributes
   >
-    <h5 class="mb-0">
+    <h2 class="mb-0">
       <button
         class="btn btn-link {{if @disabled "disabled"}} {{if @collapsed "collapsed" "expanded"}}"
         type="button"
@@ -32,6 +32,6 @@
       >
         {{yield}}
       </button>
-    </h5>
+    </h2>
   </div>
 {{/if}}
--- a/tests/integration/components/bs-accordion-test.js
+++ b/tests/integration/components/bs-accordion-test.js
@@ -1,6 +1,6 @@
 import { module } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { click, render, settled } from '@ember/test-helpers';
+import { click, find, render, settled } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import {
   accordionClass,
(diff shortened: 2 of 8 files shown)


tests/integration/components/bs-accordion-test.js
```diff
@@ -46,8 +64,9 @@ module('Integration | Component | bs-accordion', function (hooks) {
     `);
 
     assert
-      .dom(`.${accordionItemClass()}:first-child .${accordionItemHeadClass()}`)
-      .hasNoClass('collapsed', `${accordionItemHeadClass()} has not collapsed class`);
+      .dom(`.${accordionItemClass()}:first-child ${accordionItemClickableSelector()}`)
+      .hasNoClass('collapsed', `${accordionItemHeadClass()} has not collapsed class`)
+      .hasAria('expanded', 'true', 'aria: is expanded');
```
jelhan, line 69: This test (and some others in this file) weren't actually testing what we expected. The `collapsed` class is set on the `button` not on `.accordion-header`. This wasn't noticed as we only tested for absence for a class. This is always risky as it may pass just because using a wrong selector.

addon/components/bs-accordion/item/title.hbs
```diff
@@ -18,12 +19,17 @@
 {{else}}
   <div
     class="card-header"
-    role="tab"
     ...attributes
-    {{on "click" this.handleClick}}
```
jelhan, line 23: Moving the event listener from `.card-header` to `.card-header h5 button` could be seen as a breaking change. Due to margin some parts of the header may not be interactive anymore. I would consider it a bug fix. But wanted to raise that concern anyways. I guess it's most likely causing issues with tests using wrong selectors.

This only affects Bootstrap v4.

jelhan:
Will have a look into the issues with Bootstrap v4. To be honest I was too lazy to test it locally. :see_no_evil:

addon/components/bs-accordion/item/title.hbs
simonihmig: Yeah, this was probably done this way to make the whole header clickable.

For BS4 this will change that:
image

But their example also has only the button clickable, so I guess it's ok then? Even more so since we are cutting a major release anyway, so we could even mark this as a breaking change!?

tests/integration/components/bs-accordion-test.js
```diff
@@ -26,14 +26,32 @@ module('Integration | Component | bs-accordion', function (hooks) {
...
+      .exists({ count: 2 }, 'collapsable area of accordion has correct class');
     assert
       .dom(`.${accordionClass()} .${accordionItemClass()}`)
       .exists({ count: 2 }, 'accordion item has correct class');
+    assert.dom('[data-test-item="1"] h2').hasText('TITLE1', 'renders title');
+    assert
+      .dom('[data-test-item="1"] .accordion-collapse')
+      .hasAria(
+        'describedby',
+        find('[data-test-item="1"] h2').id,
```
simonihmig, line 45: This is making the BS4 tests fail, as you are asserting for `h2`, while _current_ BS5 markup has `h5`. I am pretty sure there was a reason for `h2`, maybe it was like this in 4.0 and they changed their examples later?

Anyway given that we are doing a major release soon, I think we can unify this to `h2` everywhere, and mark this PR for extra safety as breaking, what do you think?

simonihmig (commented):
This is a great improvement, thanks so much!

jelhan:
I have updated this PR to

1. align the BS4 markup with Bootstrap documentation and
2. align splat attribute targets between BS4 and BS5.

Both is introducing breaking changes for BS4. I have documented them in the PR description. Please have a look.

jelhan:
`ember-release`, `ember-beta` and `ember-canary` are failing due to the [`@ember/string` deprecation](https://deprecations.emberjs.com/v4.x#toc_ember-string-add-package) added in Ember 4.10.

addon/components/bs-accordion/item/title.hbs
```diff
@@ -19,9 +19,11 @@
 {{else}}
```
simonihmig, line 19: Given markup between BS4 and BS5 is now much more aligned, doesn't it make sense to remove that if/else condition, and only apply conditionals where they apply (classes for the outer div and h2 AFAICT)?

simonihmig (approved):
Approved, with one minor suggestion.

Also could you please check if [this disabled linting rule](https://github.com/ember-bootstrap/ember-bootstrap/pull/1844/files/9553966e061a37ec100fd126ea4b435db217db23..b6b3293ea933ada8e6d327c36ef61a0f1d7ab96f#diff-ba2167bcdb7212a632787a71cb1a61e523cd8f1b5973a627f7e1193704b44ec2R1-R2) is still needed with the new a11y-improved markup?

Feel free to merge afterwards!

addon/components/bs-accordion/item/title.hbs
jelhan: I thought about that as well. But would like to avoid the unnecessary `<div>` for BS5. And not rendering the wrapping element conditionally is very complex in current Ember.

jelhan:
> Also could you please check if [this disabled linting rule](https://github.com/ember-bootstrap/ember-bootstrap/pull/1844/files/9553966e061a37ec100fd126ea4b435db217db23..b6b3293ea933ada8e6d327c36ef61a0f1d7ab96f#diff-ba2167bcdb7212a632787a71cb1a61e523cd8f1b5973a627f7e1193704b44ec2R1-R2) is still needed with the new a11y-improved markup?

The link seems to be broken. Which linting rule is this about?

jelhan:
Merged as is to unblock the other PRs. Please open an issue on GitHub to check the linting rule. I will do asap.

PR #1845, opened 2023-01-13
add @ember/string peer dependency to fix deprecation

jelhan:
Adds `@ember/string` package as a peer dependency. Importing from it without having it as an explicit dependency has been deprecated. For addons declaring it as a peer dependency is recommended. See this thread on Discord for more details: https://discord.com/channels/480462759797063690/1063399689992540171

This is a breaking change as applications must have a direct dependency on `@ember/string`. This was added to app blueprints only two days ago: https://github.com/ember-cli/ember-cli/pull/10125


jelhan:
Need to add it as a `devDependency` as well. Will do that later today.

simonihmig:
It seems we use that only [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/d98e054abd8ea172bbd47a81655ef300aeae7f87/addon/components/bs-collapse.js#L5), or did I miss something (only used GH search)?

In which case maybe it makes more sense to remove that usage. We should be able refactor [the use of `camelize()`](https://github.com/ember-bootstrap/ember-bootstrap/blob/d98e054abd8ea172bbd47a81655ef300aeae7f87/addon/components/bs-collapse.js#L226). Then we wouldn't need to bring in that dependency, and cause further churn for users. What do you think?

jelhan:
> It seems we use that only [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/d98e054abd8ea172bbd47a81655ef300aeae7f87/addon/components/bs-collapse.js#L5), or did I miss something (only used GH search)?
> 
> In which case maybe it makes more sense to remove that usage. We should be able refactor [the use of `camelize()`](https://github.com/ember-bootstrap/ember-bootstrap/blob/d98e054abd8ea172bbd47a81655ef300aeae7f87/addon/components/bs-collapse.js#L226). Then we wouldn't need to bring in that dependency, and cause further churn for users. What do you think?

Fully agree that dropping the dependency is a better way to go forward. I implemented it in #1849.

PR #1847, opened 2023-01-14
do not fail on @ember/string deprecation

jelhan:
Fixing the `@ember/string` deprecation does not seem to be as easy as hoped. Let's ignore it for now to have a green CI pipeline again and easy development. Will remove it again as part of #1845.


simonihmig (approved):
Thanks for looking into this!

PR #1849, opened 2023-01-15
drop dependency on @ember/string

jelhan:
This drops the dependency on `@ember/string`. This fixes the `@ember/string` deprecation for Ember Bootstrap itself. But it needs to be fixed in ember-style-modifier as well before we can throw on that deprecation again.

Thanks a lot @simonihmig for the hint in https://github.com/ember-bootstrap/ember-bootstrap/pull/1845#issuecomment-1383230945.

Closes #1845


simonihmig (approved):
👍

Issue #1701, opened 2021-11-15
Add basic regression tests for docs app


jelhan:
Breaking docs app is way too easy. See #1700 as a recent example. I think we should add some basic regressions tests. Most of the issues we have seen in the past would have be caught by a test, which simply tries to build the docs app. Maybe running `yarn docs:build` in CI is good enough as a starter.

Issue #1304, opened 2020-10-23
Consider switching to a real monorepo


jelhan:
This repository already contains two projects:

1. The `ember-bootstrap` package itself in the root.
2. The documentation app in `docs/` folder.

This is very confusing. Especially as both of them have their own dependencies. I just struggled myself with starting the docs app until I realized that I need to run `yarn install` within `docs/` folder. This is also preventing us from running dummy app and tests at the same time as documentation app.

I think it will be easier to maintain and less confusing to other developers if we switch to a classic monorepo file system layout:

```
- package.json (for yarn monorepo)
- yarn.lock
- packages/
  - ember-bootstrap/
  - docs/
```

Additionally this will allow us to pull official addons into this repository itself. This has some benefits like easier testing if an addon is affected by a change in Ember Bootstrap. But it also has some trade-offs as issues and pull requests are not separated anymore and tests may be run for packages, which aren't changed. I think we should discuss it separately. But wanted to highlight that it would *enable* us to do so.

cah-briangantzler:
One note for ember addons, if you adopt the mono repo approach, you will no longer be able to point an application to a branch or commit for testing. ie you can not do this in a package.json

    "ember-cli-mirage": "git+https://github.com/miragejs/ember-cli-mirage#de8a248",

Many times this is required to allow someone to test a pending change without actually releasing a version

drauschenbach:
I think you're misusing the term [monorepo](https://en.wikipedia.org/wiki/Monorepo), since you are advocating against multiple projects in the same repo. I almost always favor decomposition as you do as well.

jelhan:
> One note for ember addons, if you adopt the mono repo approach, you will no longer be able to point an application to a branch or commit for testing. ie you can not do this in a package.json
> 
> ```
> "ember-cli-mirage": "git+https://github.com/miragejs/ember-cli-mirage#de8a248",
> ```
> 
> Many times this is required to allow someone to test a pending change without actually releasing a version

Very valid point. Maybe we should not change filesystem layout but use yarn workspaces to share the dependencies and only have one install step? :thinking: But to be honest I'm not sure yarn workspaces support to have a package at the workspace root. Would need to test that out.

cah-briangantzler:
Since this seems to be the issue with the docs folder being a separate application, perhaps moving the docs into the tests/dummy application via ember-cli-addon-docs might be the solution

simonihmig:
[...]

PR #1850, opened 2023-01-15
run lint and tests for docs app in CI

jelhan:
There aren't any meaningful tests for the docs app yet. But nevertheless this has some benefits:

1. It ensures that dependencies of docs apps could actually be installed.
2. It runs linting for docs app.
3. It ensures that testing setup of docs app is working on a basic minimum.
4. It eases adding some acceptance tests for the docs app later.

First step towards #1701.


.github/workflows/ci.yml
```diff
@@ -97,6 +97,27 @@ jobs:
...
+    needs:
+      - test
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+      - name: Install Dependencies of Ember Bootstrap addon
+        uses: bahmutov/npm-install@v1
```
jelhan, line 112: This is needed because `ember-bootstrap` is not declared as a regular dependency on the docs app. Instead it uses a discovery mechanism provided by Ember CLI. At least if I understood correctly what those lines are doing: https://github.com/ember-bootstrap/ember-bootstrap/blob/56f723c9209b220751bc1ffdb58e7d7a44323582/docs/package.json#L80-L84

We may want to refactor away from that pattern and declare it as a regular dependency using `file:` or `link:` protocol.

simonihmig: Agree. Actually converting to a real monorepo probably? Kinda inevitable when eventually refactoring to a v2 addon...

simonihmig (approved):
Awesome!

.github/workflows/ci.yml
jelhan: Yeah. Let's reopen the discussion in #1304.

Issue #1855, opened 2023-01-19
Docs are not deployed


jelhan:
Docs build is broken once again. :sob: 

It seems that it is running into an endless loop. This is the first error, which seems to be relevant:

```
There was an error running your app in fastboot. More info about the error: 
 TypeError: Cannot read property 'getHelper' of undefined
    at ConstantsImpl.helper (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:49366:96)
    at resolveHelper (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46160:22)
    at encodeOp (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48590:18)
    at pushOp (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48504:7)
    at /tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46433:7
    at Compilers.compile (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46408:7)
    at expr (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46662:19)
    at CompilePositional (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46749:7)
    at /tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48404:17
    at Compilers.compile (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:46408:7)
    at compileStatements (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48508:17)
    at maybeCompile (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48483:18)
    at CompilableTemplateImpl.compile (/tmp/broccoli-2378Z4bSkT9Wfjss/out-446-broccoli_merge_trees/assets/vendor.js:48463:14)
[...]

PR #1860, opened 2023-01-22
build docs app as part of every CI run to catch issues early

jelhan:
We have seen docs build failing to often in the past _after_ a release. We haven't run the build at all before releasing. This caused us catching issues very late. Recent example for it is #1855. An incident with a longer time to resolve was #1387. I feel we should catch this issues early. Also to ease dependency upgrades.

I hoped that we reduce the risk of broken docs app by running linting and tests for it in CI. I added this in #1850. But there are many steps only done for a production build. So the risk of failing docs deployment is still high. As it has been proven by #1855.

This increases resource consumption of our CI pipeline. But I feel that trade-off is good for now.

I hope I haven't messed up with GitHub Action workflow. Not very experienced with GitHub Actions to be honest.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,10 +3,15 @@ name: Publish docs
 on:
   workflow_dispatch:
   release:
-    types: [published,edited]
+    types:
+      - published
+      - edited
+  workflow_run:
+    workflows:
+      - CI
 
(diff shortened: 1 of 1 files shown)


simonihmig (approved):
LGTM! Thank you!

PR #1861, opened 2023-01-25
Add glint types

vlascik:
First stab at glint types for e-bootstrap. Couple of notes:

1. Didn't bother with typing helpers, they are mostly not interesting for consumers of the addon anyway.
2. Types are definitely wrong at places. Docs app passes (except for helpers), my app passes, but that doesn't cover all of the addon and its components. Some components are fairly unclear and could use proper Octane migration first.
3. I mostly didn't bother with WithBoundArgs for yielded components.
4. It's quite hard to guess the intentions of the components sometimes, so some of the types are just wild guesses.
5. There are both .ts and .d.ts files here, .ts are mostly needed for template only components. That means that there probably should be a pre-publish step, like ember ts:precompile, but that produces files in /components folder, not sure if that's a good idea.

All in all, this PR so far should take it some 80-90% there, but there will be some more work needed. However after doing this much menial typing, I'm kind of spent, so someone else will probably need to invest couple of minutes (or hours, no idea) into checking and fixing these too.

Anyway, hope these'll be useful.


addon/components/bs-button.d.ts
```diff
@@ -0,0 +1,54 @@
+import Component from '@glimmer/component';
+
+export type BsButtonType = 'primary' | 'secondary' | 'success' | 'danger' | 'warning' | 'info' | 'light' | 'dark';
```
jelhan, line 3: Valid button types depend on the Bootstrap theme being used. They are derived values from the [color maps](https://getbootstrap.com/docs/5.3/customize/color/#color-sass-maps). In the case of `@type` argument of `<BsButton>` the allowed values are the entries in [`$theme-colors` SASS map](https://github.com/twbs/bootstrap/blob/aff8f0dc9ae0ca34f4239c9395433179d1438fe6/scss/_variables.scss#L312-L321).

Not sure how we best deal with it in Glint type. I see three options:

1. Typing those arguments as `string` rather than using an enum. This has the major trade-off that the types are by far less usable.
2. Typing those arguments as an enums matching Bootstrap default values but allowing consumers to override it somehow. Not sure how that would work technically to be honest.
3. Deriving the correct type from Bootstrap SASS theme being used. I have no idea if that's technical possible and if the required tooling exists already.

I feel wrong types are worse than no types. I would tend towards typing them as `string` unless we have an option to override or derive from Bootstrap theme. But I maybe opinionated as we customize `$theme-colors` SASS map at work.

I fear there are many more cases in which the type depends on the Bootstrap theme being used. So maybe we could use the `@type` argument of `<BsButton>` just as a showcase for that general topic.

addon/components/bs-form/element.d.ts
```diff
@@ -0,0 +1,41 @@
+import Component from '@glimmer/component';
+
+interface BsFormElementSignature<T> {
+  Element: HTMLDivElement;
+  Args: {
+    controlType?: 'text' | 'checkbox' | 'radio' | 'switch' | 'textarea' | string;
```
jelhan, line 6: Would be awesome if we could detect the registered control components. Not sure how that would work. But I guess it would be only needed as interim solution anyways. I assume we are going to deprecate that feature when [first-class component templates](https://rfcs.emberjs.com/id/0779-first-class-component-templates/) are available.

jelhan, line 6: I think / hope that passing a component is supported as well. But have never tested to be honest.

```suggestion
    controlType?: 'text' | 'checkbox' | 'radio' | 'switch' | 'textarea' | string | typeof Component<unknown>;
```

addon/components/bs-form/element.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    customError?: 'string';
```
jelhan, line 7:
```suggestion
    customError?: string;
```

addon/components/bs-form/element.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    customWarning?: string;
+    doNotShowValidationForEventTargets?: unknown;
```
jelhan, line 9: Per [API docs](https://www.ember-bootstrap.com/api/classes/Components.FormElement.html#property_doNotShowValidationForEventTargets):

```suggestion
    doNotShowValidationForEventTargets?: string[] | null;
```

addon/components/bs-form.d.ts
```diff
@@ -0,0 +1,40 @@
...
+    formLayout?: 'vertical' | 'horizontal' | 'inline';
+    hideValidationsOnSubmit?: boolean;
+    horizontalLabelGridClass?: string;
+    model?: T;
+    preventConcurrency?: boolean;
+    readonly?: boolean;
+    submitOnEnter?: boolean;
+    onBefore?: (model: T) => void;
+    onInvalid?: (model: T, error: unknown) => void;
+    onSubmit?: (model: T, result: unknown) => void;
```
jelhan, line 19: Consumer can return a promise to control submission state of the form.

```suggestion
    onSubmit?: (model: T, result: unknown) => Promise<void> | void;
```

addon/components/bs-button.d.ts
```diff
@@ -0,0 +1,54 @@
...
+    iconInactive?: string;
+    outline?: boolean;
+    pendingText?: string;
+    preventConcurrency?: boolean;
+    rejectedText?: string;
+    reset?: boolean;
+    size?: string;
+    type?: BsButtonType;
+    value?: unknown;
+    onClick?: (e?: Event, value?: unknown) => void;
```
jelhan, line 25: Can return a `Promise` to control pending state of the button.

```suggestion
    onClick?: (e?: Event, value?: unknown) => Promise<void> | void;
```

addon/components/bs-button.d.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+  Blocks: {
+    default: [
+      {
+        isFulfilled: boolean;
+        isPending: boolean;
+        isRejected: boolean;
+        isSettled: boolean;
+        actions: BsButtonActions;
```
jelhan, line 34: Where is this coming from? I don't see it is in yielded hash: https://github.com/ember-bootstrap/ember-bootstrap/blob/4ec513d3dd54e0d6d91a64dcb32cf3b57989bf93/addon/components/bs-button.hbs#L9-L16

addon/components/bs-button.d.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type BsButtonType = 'primary' | 'secondary' | 'success' | 'danger' | 'warning' | 'info' | 'light' | 'dark';
+
+interface BsButtonSignature {
+  Element: HTMLButtonElement;
+  Args: {
+    active?: boolean;
+    block?: boolean;
+    bubble?: boolean;
+    buttonType?: 'button' | 'submit';
```
jelhan, line 11:
```suggestion
    buttonType?: 'button' | 'submit' | 'reset';
```

jelhan (commented):
Thanks a lot for starting to work on this! This is _huge_!

I didn't had the time to review everything. But I though it might be helpful to submit my first findings anyways to get the discussion rolling. I fear there is at least one bigger question not answered yet: How to deal with types, which depend on the Bootstrap SASS theme being used? Please find more details in one of my review comments about that one.

addon/components/bs-button.d.ts
vlascik: I know, I think I generally typed the places where this is used as BsButtonType | string, just to have some default suggestions, but to allow any string anyway. Feel free to change it however you like though.

vlascik:
@jelhan thanks for the comments, but I think the best way, instead of discussing these types going back and forth for possibly a really long time, would be just you or other maintainers with a better knowledge of the addon taking this PR as a starting point and running with it.

I got it as far as I could, most of the grind should be done, but these definitely need refining past what I can do. I don't think I have much to add from this point on (don't know the addon that much), so doing it this way may be the fastest way to get it through the finish line.

jelhan:
> @jelhan thanks for the comments, but I think the best way, instead of discussing these types going back and forth for possibly a really long time, would be just you or other maintainers with a better knowledge of the addon taking this PR as a starting point and running with it.

Could do so. But I fear neither @simonihmig nor me having much time currently to work on it. So that may delay shipping it quite a bit.

addon/components/bs-form/element.d.ts
vlascik: this is propably not the place to pass a component type, but rather <input type={{controlType}} input type.

vlascik: as mentioned elsewhere, this should be just an arg for <input type={{controlType}} html attribute, so string should be enough.

addon/components/bs-button.d.ts
vlascik: So I tried this again, and it turns out that once there's `| string` in an union type, TS can't show default suggestions and widens the type to just `string`.
I will probably remove this type and put it as string. I don't think there's any way to infer this, definitely not from SASS.

As for wrong types being worse than no types, in general I would say any types are better than no types - there is no way to type an addon of this size correctly on a first try, but as long as it doesn't modify the runtime behavior, it doesn't matter, it's just a productivity boost, and some types will be better than no types People will refine them over time for sure.

addon/components/bs-button.d.ts
vlascik: if you mean `actions`, I was under impression this is how you should type them for Glint, but I might be wrong. will check later.
The others are in the yielded hash.

vlascik:
> Could do so. But I fear neither @simonihmig nor me having much time currently to work on it. So that may delay shipping it quite a bit.

I think I have another way to check these types in my app, so I will try to fix them a bit more, but ultimately, other people will have to check them too, my app doesn't use every component or feature of this addon by far.

I wouldn't personally be too concerned about these types being wrong at places though - they don't affect the runtime, so worst case I would put a warning at the types - and let people who get the typing squiggles fix the types over time on their own.

addon/components/bs-form/element.d.ts
jelhan: Ember Bootstrap supports custom control types. E.g. the [ember-bootstrap-power-select](https://github.com/ember-bootstrap/ember-bootstrap-power-select) addon provides an integration with [Ember Power Select](https://ember-power-select.com/) allowing it to be used as a control type.

```hbs
<BsForm as |form|>
  <form.element @controlType="power-select" />
</BsForm>
```

Current that relies on naming convention. If being passed a `power-select` control type, Ember Bootstrap tries to resolve a component `<BsForm::Element::Control::PowerSelect>`. If that does not exists it falls back to a default control type.

That lookup mechanism is not fitting well with Embroider and tree-shaking. With first class component templates passing the component directly would feel much more comfortable:

```js
import PowerSelectControl from 'ember-bootstrap-power-select';

<template>
  <BsForm as |form|>
    <form.element @controlType={{PowerSelectControl}} />
  </BsForm>
</template>
```

I think / hope that is already supported today. But have never double checked.

jelhan: See https://github.com/ember-bootstrap/ember-bootstrap/pull/1861#discussion_r1097477232

addon/components/bs-button.d.ts
jelhan: > As for wrong types being worse than no types, in general I would say any types are better than no types - there is no way to type an addon of this size correctly on a first try, but as long as it doesn't modify the runtime behavior, it doesn't matter, it's just a productivity boost, and some types will be better than no types. People will refine them over time for sure anyway.

Fully agree. My statement was not about having wrong types by mistake. Rather than if we should include types, which we know to be wrong in some cases.

jelhan: > So I tried this again, and it turns out that once there's `| string` in an union type, TS can't show default suggestions and widens the type to just `string`.
> I will probably remove this type and put it as string. I don't think there's any way to infer this, definitely not from SASS.

I feel in that case typing as `string` at least in first iteration seems to be the best. We could make the type more precise while still being correct in all cases in a later iteration.

vlascik: You're right, trying to provide default values doesn't bring any value for now, so I replaced the BsButtonType for string everywhere

Issue #1863, opened 2023-02-06
Dropdown toggle caret doesn't respect left/right direction


silhusk:
The caret in the dropdown toggle is shown pointing downward also when the `@direction` is `left` or `right`. The class being added results in `dropleft`, resp. `dropright`. Whereas, for BS5 they should be `dropstart`, resp. `dropend` (they respect RTL).

https://github.com/ember-bootstrap/ember-bootstrap/blob/4ec513d3dd54e0d6d91a64dcb32cf3b57989bf93/addon/components/bs-dropdown.js#L242

https://github.com/ember-bootstrap/ember-bootstrap/blob/4ec513d3dd54e0d6d91a64dcb32cf3b57989bf93/addon/components/bs-dropdown.js#L244

```console
$ npx ember bootstrap:info
Npm packages:
ember-bootstrap: ^6.0.0-1 -> 6.0.0-1
ember-cli: ^4.10.0 -> 4.10.0
bootstrap: ^5.1.1 -> 5.2.3
bootstrap-sass: n/a
ember-cli-sass: ^11.0.1 -> 11.0.1
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 5
importBootstrapFont: false
importBootstrapCSS: false
```

jelhan:
Is it working if using `@direction="start"` and `@direction="end"`? I also recently wondered if `left` and `right` values are still correct for those arguments if using BS5 when reviewing the Glint types (#1861). Maybe it is just an error in API docs?

silhusk:
By using `@direction="start"` it does add the correct class `dropstart` , but then popper is given a wrong argument and the menu opens down (the default).

https://github.com/ember-bootstrap/ember-bootstrap/blob/4ec513d3dd54e0d6d91a64dcb32cf3b57989bf93/addon/components/bs-dropdown/menu.js#L129

jelhan:
Thanks for investigating! I wasn't aware that it's also used with Popper. Do you know how that misalignment on namings is solved in Bootstrap itself?

silhusk:
Relevant snippets taken from https://github.com/twbs/bootstrap/blob/cbc4e3a4098506315c32346b194e9e1a4d7db091/js/src/dropdown.js

```js
// Line 50
const CLASS_NAME_DROPEND = 'dropend'
[...]

PR #1867, opened 2023-02-13
fix: correct dropdown direction class for BS5

silhusk:
Short-term fix for #1863, applying the correct Bootstrap 5 classes for dropdowns with direction left or right.

### DIFF
--- a/addon/components/bs-dropdown.js
+++ b/addon/components/bs-dropdown.js
@@ -3,6 +3,7 @@ import { action, computed } from '@ember/object';
 import Component from '@ember/component';
 import defaultValue from 'ember-bootstrap/utils/default-decorator';
 import { assert } from '@ember/debug';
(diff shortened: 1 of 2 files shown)


jelhan (approved):
Looks good to me. Only the linting issue needs to be fixed. @silhusk could you please take care of that one? `yarn lint:fix` should do the job.

jelhan:
@silhusk It seems that this breaks tests for FastBoot. Would be great if you could have a look.

jelhan:
> @silhusk It seems that this breaks tests for FastBoot. Would be great if you could have a look.

Please ignore. It is not related to your changes. I see the same error in other PRs as well.

jelhan:
@silhusk CI pipeline is fixed. Would be great if you could merge upstream master into your feature branch to double check.

PR #1875, opened 2023-02-28
WIP: drop RSVP dependency 

NullVoxPopuli:
This stems from an issue originally reported here: https://github.com/ef4/ember-auto-import/issues/564
Which was caused by this new feature to auto-import: https://github.com/ef4/ember-auto-import/pull/553
Which was intended to solve: https://github.com/ef4/ember-auto-import/issues/504
Which then had the default reverted here: https://github.com/ef4/ember-auto-import/pull/568
But an attempt to add more details around the hack is here: https://github.com/ef4/ember-auto-import/pull/566 (but this revealed just how important really solving the issue is, rather than continuing to work with the earlyBootSet)

Ultimately, RSVP isn't needed for anything anymore, so I figured I'd have a stab at just removing it for the time being

### DIFF
--- a/addon/components/bs-carousel.js
+++ b/addon/components/bs-carousel.js
@@ -7,7 +7,6 @@ import Component from '@ember/component';
 import ComponentParent from 'ember-bootstrap/mixins/component-parent';
 import { schedule, scheduleOnce } from '@ember/runloop';
 import { task, timeout } from 'ember-concurrency';
-import RSVP from 'rsvp';
 import defaultValue from 'ember-bootstrap/utils/default-decorator';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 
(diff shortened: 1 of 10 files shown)


NullVoxPopuli:
Looks like I have some tests to figure out

jelhan:
> Looks like I have some tests to figure out

I think you missed to address this usage of RSVP: https://github.com/ember-bootstrap/ember-bootstrap/blob/f4b6cefe1514ac59262bb94f783af179cbc8b3a7/tests/integration/components/bs-form-test.js#L4 It requires a little bit more refactoring work as it use `RSVP.defer`. We could introduce an utility function like this as a replacement for now:

```js
function defer() {
  let resolve;
  let reject;
  const promise = new Promise((resolvePromise, rejectPromise) => {
   resolve = resolvePromise;
   reject = rejectPromise;
  });
  
  return { promise, resolve, reject };
}
```

jelhan:
Considering it as an enhancement rather than internal change as it is one dependency less, which is causing potential issues for consumers and may increase bundle size.

jelhan:
@NullVoxPopuli Are you planning to follow-up on this one?

NullVoxPopuli:
Naw, feel free ho take it from here, i have enough on my plate

jelhan:
Implemented in #2055

PR #1878, opened 2023-03-01
add basic test coverage for docs app

### LINKED ISSUE #1701: Add basic regression tests for docs app

jelhan:
This adds very basic test coverage for docs app. It only checks that every route could be entered without throwing. I hope that's a good starting point to assume that changes (especially upgrading dependencies) will not break it. If we notice that something is broken, we could add tests for that specific parts than.

Closes #1701


jelhan:
@simonihmig Do you feel it would be safe enough merging all the pending dependency upgrades without any manual check after adding this very basic tests? Do you have a feeling how likely breaking something is?

simonihmig:
Thanks a lot for working on this, and sorry for the little attention I am currently able to give here!

> Do you feel it would be safe enough merging all the pending dependency upgrades without any manual check after adding this very basic tests

All the renovate stuff? Yeah, go ahead and merge I would say. Been doing that previously as well, can't afford manual checks IMHO, as long as CI is happy!

PR #1891, opened 2023-03-07
migrate husky configuration to v8

### LINKED ISSUE #1892: Commits on master fail with Command "husky-run" not found.
Husky configuration is currently broken. Husky was upgraded to v8 in #1794 but configuration was not adjusted. PR to do so is pending in #1891. Until this has been fixed contributors can commit with `--no-verify` flag as a workaround.

jelhan:
Husky was broken by #1794. Configuration needs to be migrated to be compatible with Husky v8.

Migration was done following the recommended steps here: https://typicode.github.io/husky/#/?id=husky-4-to-8-cli

1. Run [husky-4-to-8](https://github.com/typicode/husky-4-to-8#husky-4-to-8) migration script
2. Adjust created file to use local `lint-staged`  binary

Closes #1892

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn run lint-staged
--- a/package.json
+++ b/package.json
@@ -35,12 +35,8 @@
     "start": "ember serve",
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test",
-    "test:ember-compatibility": "ember try:each"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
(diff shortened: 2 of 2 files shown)


jelhan:
@simonihmig I don't have much experience using Husky. Would be great if you could take a look.

simonihmig (approved):
Unfortunately, I don't have tried husky 8 yet, so I cannot really contribute much here. But seems, and if it's working for you, then I think this should be fine.

FWIW, I remember us discussing whether to use husky at all here. If you prefer to get rid off it, as it's slowing us down, or not providing the value needed, then I wouldn't mind also to remove it. 🤷‍♂️

jelhan:
> FWIW, I remember us discussing whether to use husky at all here. If you prefer to get rid off it, as it's slowing us down, or not providing the value needed, then I wouldn't mind also to remove it. :man_shrugging:

Honestly I got used to it. :smile_cat:

PR #1908, opened 2023-03-25
remove ember-export-application-global dev dependency

jelhan:
`ember-export-application-global` is deprecated. It has been removed from addon and app blueprints. It is not compatible with Ember 5 and breaks canary tests therefore.

This only addresses one of the issues with Ember 5. Canary build is still failing due to issues with FastBoot.


jelhan:
Canary builds are expected to still fail. This PR only fixes one out of many incompatibilities with Ember 5.

PR #1919, opened 2023-04-14
Allow multiple custom grid classes in horizontal layout

dagroe:
This extends the logic for `horizontalLabelGridClass` to accept a string with multiple custom grid classes, e.g. `col-md-4 col-lg-2`.

I found this useful when adapting the form layout for multiple breakpoints.

### DIFF
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -189,7 +189,7 @@ export default class Form extends Component {
    */
 
   /**
-   * The Bootstrap grid classed for form labels. This is used by the `Components.FormElement` class as a default for the
+   * The Bootstrap grid classes for form labels. This is used by the `Components.FormElement` class as a default for the
    * whole form.
    *
    * @property horizontalLabelGridClass
--- a/addon/components/bs-form.js
+++ b/addon/components/bs-form.js
@@ -189,7 +189,7 @@ export default class Form extends Component {
    */
 
   /**
-   * The Bootstrap grid class for form labels. This is used by the `Components.FormElement` class as a default for the
+   * The Bootstrap grid classed for form labels. This is used by the `Components.FormElement` class as a default for the
    * whole form.
    *
    * @property horizontalLabelGridClass
--- a/addon/helpers/bs-form-horiz-input-class.js
+++ b/addon/helpers/bs-form-horiz-input-class.js
@@ -6,8 +6,14 @@ export default helper(function bsFormHorizInputClass([horizontalLabelGridClass] 
   if (isBlank(horizontalLabelGridClass)) {
     return undefined;
   }
-  let parts = horizontalLabelGridClass.split('-');
-  assert('horizontalInputGridClass must match format bootstrap grid column class', parts.length === 3);
-  parts[2] = 12 - parts[2];
-  return parts.join('-');
+
+  let originalClasses = horizontalLabelGridClass.split(' ');
+  let inputClasses = originalClasses.map((originalClass) => {
+    let parts = originalClass.split('-');
+    assert('horizontalInputGridClass must match format bootstrap grid column class', parts.length === 3);
+    parts[2] = 12 - parts[2];
+    return parts.join('-');
+  });
+
+  return inputClasses.join(' ');
 });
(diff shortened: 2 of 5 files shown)


addon/components/bs-form.js
```diff
@@ -189,7 +189,7 @@ export default class Form extends Component {
    */
 
   /**
-   * The Bootstrap grid class for form labels. This is used by the `Components.FormElement` class as a default for the
+   * The Bootstrap grid classed for form labels. This is used by the `Components.FormElement` class as a default for the
```
jelhan, line 192: Small typo

```suggestion
   * The Bootstrap grid classes for form labels. This is used by the `Components.FormElement` class as a default for the
```

addon/helpers/bs-form-horiz-input-class.js
```diff
@@ -6,8 +6,14 @@ export default helper(function bsFormHorizInputClass([horizontalLabelGridClass]
...
   }
-  let parts = horizontalLabelGridClass.split('-');
-  assert('horizontalInputGridClass must match format bootstrap grid column class', parts.length === 3);
-  parts[2] = 12 - parts[2];
-  return parts.join('-');
+
+  let originalClasses = horizontalLabelGridClass.split(' ');
+  let inputClasses = originalClasses.map((originalClass) => {
+    let parts = originalClass.split('-');
+    assert('horizontalInputGridClass must match format bootstrap grid column class', parts.length === 3);
```
jelhan, line 13: Disclaimer: Not related to this PR.

The assumption that all Bootstrap grid column classes having 3 parts is wrong. Bootstrap also supports columns without breakpoint: `col-3`. That is used for all breakpoints unless overruled by a breakpoint specific column class (e.g. `col-md-3`). Will verify if it actually breaks in that case and open a separate ticket to track.

addon/helpers/bs-form-horiz-input-class.js
```diff
@@ -6,8 +6,14 @@ export default helper(function bsFormHorizInputClass([horizontalLabelGridClass]
...
+    parts[2] = 12 - parts[2];
```
jelhan, line 14: Disclaimer: Not introduced by this PR.

This assumes that a 12 columns grid is used. That's Bootstrap's default value. But it could be changed by setting `$grid-columns` SASS variable as documented here: https://getbootstrap.com/docs/5.3/layout/grid/#sass-variables

Will verify that it actually breaks and open an issue to track it.

tests/integration/components/bs-form/element-test.js
```diff
@@ -251,6 +251,19 @@ module('Integration | Component | bs-form/element', function (hooks) {
...
+      await render(hbs`
+        <BsForm @formLayout="horizontal" as |form|>
+          <form.element @controlType="text" @label="some label" @options={{this.simpleOptions}} @horizontalLabelGridClass="col-md-3 col-lg-2" data-test-form-element />
+        </BsForm>
+      `);
+
+      assert.dom('[data-test-form-element] > label').hasClass('col-md-3');
+      assert.dom('[data-test-form-element] > div').hasClass('col-md-9');
+      assert.dom('[data-test-form-element] > label').hasClass('col-lg-2');
+      assert.dom('[data-test-form-element] > div').hasClass('col-lg-10');
```
jelhan, line 265: You could chain them to improve readability:

```suggestion
      assert.dom('[data-test-form-element] > label').hasClass('col-md-3').hasClass('col-lg-2');
      assert.dom('[data-test-form-element] > div').hasClass('col-md-9').hasClass('col-lg-10');
```

jelhan (changes requested):
Thanks a lot. Just one typo.to fix before merging. Additionally you may want to improve readability of the tests by chaining assertions. But that's not required for getting it merged.

The other two comments related to (potential) bugs are not related to this PR. They are only notes to myself to not forget looking into them. Not expected to be addressed in this PR.

tests/integration/components/bs-form/element-test.js
dagroe: 👍  Added chaining

dagroe:
Thanks for your quick feedback @jelhan. I added your suggestions.

jelhan (approved):
Thanks a lot for addressing review comments.

Failing tests does not seem to be related. But I would prefer resolving it before. Will have a look asap.

simonihmig (approved):
Thank you!

The failing tests seem to be related to Ember 5, which should get fixed by #1926.

PR #1927, opened 2023-05-20
Try to fix docs workflow

simonihmig:


simonihmig:
Was seeing a lot of jobs waiting for a runner, and getting canceled after 1d. But the change here seems not to fix that. Instead it seems there are way too many docs jobs getting spawned!?

image

@jelhan has anything changed here recently?

PR #1946, opened 2023-07-25
bs-dropdown dynamic positioning should not be used when in a navbar

craigteegarden:
fixes https://github.com/ember-bootstrap/ember-bootstrap/issues/1945

Bootstrap 5 (and 4) do not use popper to position dropdowns that are in navs. The bootstrap dropdown js detects if it is in a nav or specified to be static and instructs Popper to not apply any styles ([here's the dropdown disable code](https://github.com/twbs/bootstrap/blob/8fcfce1ebd988fd4e5ed2f51acaa5c49e6b301c9/js/src/dropdown.js#L312-L319)).

This change uses the existing `inNav` flag and passes it to the dropdown menu component and uses that to disable popper's styles in the same method as bootstrap.

### DIFF
--- a/tests/integration/components/bs-dropdown-test.js
+++ b/tests/integration/components/bs-dropdown-test.js
@@ -442,22 +442,30 @@ module('Integration | Component | bs-dropdown', function (hooks) {
   });
 
   test('dropdown in nav disables dynamic positioning', async function (assert) {
-    await render(hbs`
-    <BsDropdown @inNav={{true}} as |dd|>
-      <dd.toggle>Dropdown</dd.toggle>
-      <dd.menu as |menu|>
-        <menu.item>
-          <menu.link-to @route="index">Home</menu.link-to>
-        </menu.item>
-      </dd.menu>
-    </BsDropdown>`);
+    this.set('inNav', false);
+
+    await render(hbs`<BsDropdown @inNav={{this.inNav}} as |dd|>
+  <dd.toggle>Dropdown</dd.toggle>
+  <dd.menu as |menu|>
+    <menu.item>
+      <menu.link-to @route='index'>Home</menu.link-to>
+    </menu.item>
+  </dd.menu>
+</BsDropdown>`);
 
     await click('a.dropdown-toggle');
 
-    assert.equal(
+    assert.ok(
+      this.element.querySelector('.dropdown-menu').style.length > 0,
+      'Dynamic positioning applied when inNav is set to false',
+    );
+
+    this.set('inNav', true);
+
+    assert.strictEqual(
       this.element.querySelector('.dropdown-menu').style.length,
       0,
-      'Dynamic positioning not applied when inNav is set to true'
+      'Dynamic positioning not applied when inNav is set to true',
     );
   });
 
--- a/tests/integration/components/bs-dropdown-test.js
+++ b/tests/integration/components/bs-dropdown-test.js
@@ -441,6 +441,26 @@ module('Integration | Component | bs-dropdown', function (hooks) {
     );
   });
 
+  test('dropdown in nav disables dynamic positioning', async function (assert) {
+    await render(hbs`
+    <BsDropdown @inNav={{true}} as |dd|>
+      <dd.toggle>Dropdown</dd.toggle>
+      <dd.menu as |menu|>
+        <menu.item>
+          <menu.link-to @route="index">Home</menu.link-to>
+        </menu.item>
+      </dd.menu>
+    </BsDropdown>`);
+
+    await click('a.dropdown-toggle');
+
+    assert.equal(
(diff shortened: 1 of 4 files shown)


tests/integration/components/bs-dropdown-test.js
```diff
@@ -365,6 +365,26 @@ module('Integration | Component | bs-dropdown', function (hooks) {
...
+      <dd.menu as |menu|>
+        <menu.item>
+          <menu.link-to @route="index">Home</menu.link-to>
+        </menu.item>
+      </dd.menu>
+    </BsDropdown>`);
+
+    await click('a.dropdown-toggle');
+
+    assert.equal(
```
jelhan, line 381: Please use `assert.strictEqual`. Please see this ESLint rule for details: https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal.md

```suggestion
    assert.strictEqual(
```

tests/integration/components/bs-nav-test.js
```diff
@@ -59,6 +59,13 @@ module('Integration | Component | bs-nav', function (hooks) {
     assert.dom('.nav > li').exists({ count: 3 }, 'it has the nav item');
     assert.dom('.nav > li > a[href="/"]').exists({ count: 2 }, 'it has the nav link');
     assert.dom('.nav > li.dropdown').exists({ count: 1 }, 'it has a dropdown as a nav item');
+
+    await click('.nav a.dropdown-toggle');
+    assert.equal(
```
jelhan, line 64: As above.

```suggestion
    assert.strictEqual(
```

tests/integration/components/bs-dropdown-test.js
```diff
@@ -365,6 +365,26 @@ module('Integration | Component | bs-dropdown', function (hooks) {
...
+      this.element.querySelector('.dropdown-menu').style.length,
+      0,
+      'Dynamic positioning not applied when inNav is set to true'
+    );
```
jelhan, line 385: Do we have any test, which asserts the opposite? I mean that the element has custom styles if `@inNav` is not set (or set to `false`)? I fear otherwise this test can quickly turn to a false positive.

Does `@inNav` tracks changes? If so, I would prefer testing the assumption that it `style.length` is greater than `0` in the same test.

tests/integration/components/bs-nav-test.js
```diff
@@ -59,6 +59,13 @@ module('Integration | Component | bs-nav', function (hooks) {
...
+      this.element.querySelector('.nav > li.dropdown .dropdown-menu').style.length,
+      0,
+      'Dynamic positioning not applied when inNav is set to true'
+    );
```
jelhan, line 68: Can we assert instead that the dropdown is actually in the navbar? Can we compare the `.nav` box position with the `.dropdown-menu` box position and assert that it is inside of it? In that case we would test less the implementation details but more what is expected from end-user's perspective.

jelhan (changes requested):
Thanks a lot for working on this. I'm sorry that review took that long.

I left some comments. Please have a look.

Please resolve the merge conflicts. I assume they are caused by adopting Prettier for Glimmer templates. Maybe also by the fixes for `{{link-to}}` component with positional arguments. I have landed them in main branch recently.

tests/integration/components/bs-nav-test.js
craigteegarden: fixed

tests/integration/components/bs-dropdown-test.js
craigteegarden: I added before/after to this test. It initially starts with `inNav=false`, checks that the dynamic Popper styles are applied, flips `inNav=true` and checks that the dynamic styles are not applied. This should avoid a false positive in the future.

tests/integration/components/bs-nav-test.js
craigteegarden: Good question. Thinking about it more, I don't actually think we can check the position for this test. The in nav behavior happens on both the mobile and non-mobile styles. The non-mobile dropdown looks essentially identical to the dynamically positioned version and is still a "popup", meaning there isn't a difference that can be checked using box positions relative to the menu. I'd be happy to change this further, but I'm not sure how to do it.

craigteegarden:
Thank you for reviewing!

I rebased and fixed the lint issues.

I also (mostly) addressed the test comments, see individual comments for details.

tests/integration/components/bs-nav-test.js
jelhan: Thanks for looking into it. Seems like asserting positioning of the element would only work (and be meaningful) for mobile breakpoints. I'm not aware of a strategy how we could set viewport size for a specific test. Likely it wouldn't be worth it as we already have good test coverage asserting against other aspects. Let's keep it as it is. 👍

PR #1968, opened 2023-11-05
unpin dependencies of addon

jelhan:
Same issue as for the docs app (#1967). Mainly affecting the dev dependencies. But even some dependencies were pinned. Not sure why we haven't noticed that earlier.

Regenerated the lock file here as well.


jelhan:
Changing the configuration in #1970. But happy to revisit if there are better patterns. I feel its easier to maintain if we have ranges. But I may be wrong.

simonihmig:
It's recommended by renovate: https://docs.renovatebot.com/dependency-pinning/

jelhan:
> It's recommended by renovate: https://docs.renovatebot.com/dependency-pinning/

Thanks a lot for sharing. Very interesting. Wasn't aware of this document yet. They convinced me. 🙂

PR #2003, opened 2023-11-08
upgrade docs app to Ember 4.12

jelhan:


docs/app/models/component.js
```diff
@@ -1,27 +1,30 @@
-import { reads } from '@ember/object/computed';
-import EmberObject, { computed } from '@ember/object';
 import config from 'ember-bootstrap-docs/config/environment';
 
-export default EmberObject.extend({
-  id: null,
-  title: null,
-  description: null,
+export default class Component {
```
jelhan, line 3: Not sure why it was broken. But `demoUrl` was undefined with the old pattern in Ember 4.12. I didn't invested time debugging as we need to refactor to native classes anyways.

PR #2025, opened 2023-11-16
Refactor BsNavbar, BsNavbar::Content and BSNavbar::Toggle to Glimmer components

SanderKnauff:
* Converts `BsNavbar`, `BsNavbar::Toggle` and `BsNavbar::Content` to Glimmer components
* Replaces computed properties in `BsNavbar`, `BsNavbar::Toggle` and `BsNavbar::Content` with `get` properties
* Removes observer in `BsNavbar`

### DIFF
--- a/addon/components/bs-navbar.js
+++ b/addon/components/bs-navbar.js
@@ -1,9 +1,6 @@
-import { tagName } from '@ember-decorators/component';
-import { observes } from '@ember-decorators/object';
-import { action, computed } from '@ember/object';
-import Component from '@ember/component';
-import listenTo from 'ember-bootstrap/utils/cp/listen-to';
-import defaultValue from 'ember-bootstrap/utils/default-decorator';
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
 import { isBlank } from '@ember/utils';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
@@ -43,7 +40,7 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
   * [Components.NavbarToggle](Components.NavbarToggle.html)
   * [Components.NavbarNav](Components.NavbarNav.html)
 
-  Furthermore references to the following actions are yielded:
+  Furthermore, references to the following actions are yielded:
 
   * `collapse`: triggers the `onCollapse` action and collapses the navbar (mobile)
   * `expand`: triggers the `onExpand` action and expands the navbar (mobile)
@@ -52,7 +49,7 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
   ### Responsive Design
 
   For the mobile breakpoint the Bootstrap styles will hide the navbar content (`{{navbar.content}}`). Clicking on the
-  navbar toggle button (`{{navbar.toggle}}`) will expand the menu. By default all nav links (`<nav.linkTo @route="...">`) are already
+  navbar toggle button (`{{navbar.toggle}}`) will expand the menu. By default, all nav links (`<nav.linkTo @route="...">`) are already
   wired up to call the navbar's `collapse` action, so clicking any of them will collapse the navbar. To selectively
   prevent that, you can set its `collapseNavbar` property to false:
 
@@ -102,7 +99,6 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
   @extends Ember.Component
   @public
 */
-@tagName('')
 @deprecateSubclassing
 export default class Navbar extends Component {
   /**
@@ -113,15 +109,16 @@ export default class Navbar extends Component {
    * @default true
    * @public
    */
-  @defaultValue
-  collapsed = true;
+  get collapsed() {
+    return this._toggledCollapse ?? this.args.collapsed ?? true;
+  }
 
   /**
-   * @property _collapsed
+   * @property _toggledCollapse
    * @private
    */
-  @listenTo('collapsed')
-  _collapsed;
+  @tracked
+  _toggledCollapse;
 
   /**
    * Controls whether the wrapping div is a fluid container or not.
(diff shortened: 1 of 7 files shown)


addon/components/bs-navbar.js
```diff
@@ -211,15 +208,19 @@ export default class Navbar extends Component {
    * @event onCollapse
    * @public
    */
-  onCollapse() {}
+  get onCollapse() {
+    return this.args.onCollapse ?? (() => {});
+  }
```
jelhan, line 213: Why do we need a noop function as a default here? Can't we change the invoking side to only invoke it if it is provided?

Same question applies to `onCollapsed`, `onExpand`, and `onExpanded`.

addon/components/bs-navbar.js
```diff
@@ -230,28 +231,18 @@ export default class Navbar extends Component {
...
-    let collapsed = this._collapsed;
-    let active = this.active;
-    if (collapsed !== active) {
-      return;
-    }
-    if (collapsed === false) {
-      this.show();
-    } else {
-      this.hide();
-    }
```
jelhan, line 254: I struggle to understand the impact of this change. Especially I struggle to understand what this code was doing _before_ the change.

* Are `show` and `hide` methods defined somewhere?
* `active` seems to be always `undefined`.

Maybe it was deadcode, which was always stepping out at line 248 as `_collapsed` was a boolean and `active` was always `undefined`? It seems that `collpased !== active` always returned `true` causing the observe to do nothing in practice.

Does that match your findings? Or do I miss something?

addon/components/bs-navbar.js
```diff
@@ -131,23 +128,23 @@ export default class Navbar extends Component {
...
    * @public
    */
 
-  @computed('fluid', 'container')
   get containerClass() {
-    if (this.container) {
-      return `container-${this.container}`;
+    if (this.args.container) {
+      return `container-${this.args.container}`;
     }
```
jelhan, line 148: For readability I tend to prefer destructing.

```suggestion
    const { container } = this.args;
    
    if (container) {
      return `container-${container}`;
    }
```

addon/components/bs-navbar.js
```diff
@@ -295,8 +286,12 @@ export default class Navbar extends Component {
    * @default 'lg'
    * @public
    */
-  @defaultValue
-  toggleBreakpoint = 'lg';
+  get toggleBreakpoint() {
+    if (this.args.toggleBreakpoint === undefined) {
+      return 'lg';
+    }
+    return this.args.toggleBreakpoint;
```
jelhan, line 293: Can we simplify to the following or do I miss something?

```suggestion
    return this.args.toggleBreakpoint ?? 'lg';
```

addon/components/bs-navbar/toggle.js
```diff
@@ -18,17 +16,19 @@ import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
...
+  }
 
   /**
    * @event onClick
    * @public
    */
-  onClick() {}
+  get onClick() {
+    return this.args.onClick ?? (() => {});
+  }
```
jelhan, line 31: Can we conditional apply the modifier instead of providing a noop function to it? https://v5.chriskrycho.com/journal/conditional-modifiers-and-helpers-in-emberjs/

jelhan (approved):
Thanks a lot for following up with the refactoring!

I left a few questions. Especially for removing the existing observer I would like to double check my analysis that it had no impact at all.

The rest are all only optional.

jelhan:
Failing tests for `ember-canary` do not seem to be related. Tests are failing for `<BsAlert>`. Not for a component touched here. Seeing the same errors for PRs created by RenovateBot. So either a regression in Ember canary release or an improvement which highlights an issue in our code. But nothing, which should be addressed in this PR. I'm okay with merging while those tests are failing.

addon/components/bs-navbar.js
SanderKnauff: Good point! I hadn't even considered that as I was too focused on matching exactly what the old `@defaultValue` decorator did.
Removing these fields and invoking them conditionally also will get rid of some documentation however. If we were to add Typescript to this project, we would be able to place that in the Arguments/Signature interface, where it would make more sense, I think.

addon/components/bs-navbar.js
SanderKnauff: This was a pretty tough one. I could not find any documentation on the existence of `this.show` and `this.hide`. Neither in the Ember docs or in the code itself. It seems like these methods just did not exist at all. Similar methods do exist in `bs-collapse` however. Using `console.log` on these functions (and `this.active`) seems to confirm that they are undefined. It is possible in this implementation that someone would add these 3 things through the component parameters though...

For the current implementation, everything relies on derrived data using
```js
  get collapsed() {
    return this._toggledCollapse ?? this.args.collapsed ?? true;
  }
  ```
  With priority being given to manual toggle.

I based the rules of the implementation mostly on the tests

**Comment by SanderKnauff** [comment]
> Failing tests for `ember-canary` do not seem to be related. Tests are failing for `<BsAlert>`. Not for a component touched here. Seeing the same errors for PRs created by RenovateBot. So either a regression in Ember canary release or an improvement which highlights an issue in our code. But nothing, which should be addressed in this PR. I'm okay with merging while those tests are failing.

I've found the root cause of the issue and added commit to fix the issue. It has to do with `@localCopy` from `tracked-toolbox` not working in the `ember-canary` test. A small refactor to more pure Ember reactivity resolved the issue with `bs-alert`. Please let me know if you are fine with it being fixed in the PR or if you want me to extract it into a separate PR.

**Comment by jelhan** [comment]
> I've found the root cause of the issue and added commit to fix the issue. It has to do with `@localCopy` from `tracked-toolbox` not working in the `ember-canary` test. A small refactor to more pure Ember reactivity resolved the issue with `bs-alert`. Please let me know if you are fine with it being fixed in the PR or if you want me to extract it into a separate PR.

Thanks a lot for investigating and fixing;

If it's not causing much more work for you, I would prefer a separate PR. It makes it easier to document the change in changelog and reference it if needed later.

File: addon/components/bs-navbar.js [hunk 6]
```diff
@@ -350,4 +324,24 @@ export default class Navbar extends Component {
...
+
+  /**
+   * The action to be sent when the navbar is about to be expanded.
+   *
+   * You can return false to prevent expanding the navbar automatically, and do that in your action by
+   * setting `collapsed` to false.
+   *
+   * @event onExpand
+   * @public
+   */
```
jelhan, line 346: Can you please restore at original place? Moving it makes it challenging to use Git blame to understand when something changed and why.

jelhan (changes requested):

jelhan (approved):
Thanks a lot!

PR #2027, opened 2023-11-18
Refactor bs-alert to remove @localCopy and avoid mutation after consumption error in Ember canary

SanderKnauff:
The `@localCopy` decorate caused issues on Ember canary builds. By refactoring this component to use pure Ember reactivity we can get rid of the decorator.

### DIFF
--- a/addon/components/bs-alert.js
+++ b/addon/components/bs-alert.js
@@ -5,7 +5,6 @@ import { later } from '@ember/runloop';
 import usesTransition from 'ember-bootstrap/utils/decorators/uses-transition';
 import deprecateSubclassing from 'ember-bootstrap/utils/deprecate-subclassing';
 import arg from 'ember-bootstrap/utils/decorators/arg';
-import { localCopy } from 'tracked-toolbox';
 
 /**
   Implements [Bootstrap alerts](http://getbootstrap.com/components/#alerts)
@@ -75,12 +74,6 @@ export default class Alert extends Component {
   hidden = !this.visible;
 
   /**
-   * This is an unfortunate duplication of the previous property, but this is untracked to avoid causing the dreaded "same computation" assertion in GlimmerVM when reading a tracked property before setting it.
-   * @private
-   */
-  _hidden = !this.visible;
-
-  /**
    * This property controls if the alert should be visible. If false it might still be in the DOM until the fade animation
    * has completed.
    *
@@ -92,14 +85,15 @@ export default class Alert extends Component {
    * @default true
    * @public
    */
-  @arg
-  visible = true;
+  get visible() {
+    return this._visible ?? this.args.visible ?? true;
+  }
 
   /**
    * @property _visible
    * @private
    */
-  @localCopy('visible')
+  @tracked
   _visible;
 
   /**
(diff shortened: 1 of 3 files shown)


addon/components/bs-alert.js
```diff
@@ -200,21 +194,21 @@ export default class Alert extends Component {
...
-            this.hidden = this._hidden = true;
+            this.hidden = true;
             this.args.onDismissed?.();
           }
         },
         this.fadeDuration,
       );
     } else {
-      this.hidden = this._hidden = true;
+      this.hidden = true;
```
jelhan, line 204: I'm surprised that this is not causing a modification after consumption error. The property is read earlier in the same method. I assume all complexity before was only about preventing the modification after consumption. Maybe timing changed and it's not invoked anymore during a rendering cycle? 🤔

addon/components/bs-alert.js
```diff
@@ -92,14 +85,15 @@ export default class Alert extends Component {
...
+    return this._visible ?? this.args.visible ?? true;
+  }
 
   /**
    * @property _visible
    * @private
    */
-  @localCopy('visible')
+  @tracked
   _visible;
```
jelhan, line 97: I wonder if this still works as expected. `localCopy` observes changes to the source and resets the local changes whenever source value changes:

> @localCopy
>
> Creates a local copy of a remote value. The local copy can be updated locally, but will also update if the remote value ever changes
>
> https://github.com/tracked-tools/tracked-toolbox#localcopy

Let's double check that we have sufficient test coverage for the scenario that `@visible` argument changes _after_ the internal visibility changed.

jelhan (commented):
Thanks a lot for picking it up. I'm surprised that it works to be honest. If it does, it would be a great simplification. But let's better double check that test coverage is sufficient before merging.

jelhan:
I think we are missing a test that alert can be shown again _after_ user dismissed it by clicking the close button.

Arrange:

1. Render `<BsAlert>` with `@visible` argument being `true`.
2. Hide the alert by clicking the close button and set property bound to `@visible` argument to `false` in `onDismissed` event.
3. Set property bound to `@visible` argument to `true` again.

Assert:

- Alert is visible.

Do you have time to add such a test? Otherwise I can try doing it the next days.

SanderKnauff:
> I think we are missing a test that alert can be shown again _after_ user dismissed it by clicking the close button.
> 
> Arrange:
> 
> 1. Render `<BsAlert>` with `@visible` argument being `true`.
> 2. Hide the alert by clicking the close button and set property bound to `@visible` argument to `false` in `onDismissed` event.
> 3. Set property bound to `@visible` argument to `true` again.
> 
> Assert:
> 
> * Alert is visible.
> 
> Do you have time to add such a test? Otherwise I can try doing it the next days.

I've added a test. If I actually try to do this on the master branch, it does fail with a modification-after-consumption error. With the refactored component it seems to work if I understand the intent of the test correctly

jelhan (commented):
Thanks a lot for adding the additional test. Changes look good to me.

Did you verified manually that animations are still working as expected? I don't think we have proper test coverage for the animations. After we verified manually that those are still working, I'm happy to merge it.

Thanks a lot for all your work!

Issue #1979, opened 2023-11-06
Migrate to a modern package manager


jelhan:
This project is still using yarn v1. We should migrate to a recent version of NPM, Yarn or PNPM.

This includes:

* Updating the GitHub Actions to use the other package manager
* Regenerating lock file with that new package manager
* Adjusting RenovateBot configuration
* Updating contribution docs

It should be implemented both for the addon as well as the docs app.

simonihmig:
I used pnpm in the more recent projects, would probably suggest that. It's much better at getting peer deps right. Yarn PnP might also be able to do this, however I believe we'll run into issues with the ember ecosystem expecting traditional node_modules on disk...

PR #2028, opened 2023-11-22
#1979 Migrate project to pnpm

SanderKnauff:
This PR addresses issue #1979

pnpm has been chosen over another package manager because the Ember ecosystem seems to be using it over `npm` or newer versions of `yarn`.

For the purposes of keeping development and CI tooling in sync, Volta has also been added to the `package.json`.


package.json
```diff
@@ -80,7 +80,7 @@
     "@glimmer/component": "1.1.2",
     "@glimmer/tracking": "1.1.2",
     "@release-it-plugins/lerna-changelog": "6.0.0",
-    "bootstrap": "5.3.2",
+    "bootstrap": "^5.3.2",
```
jelhan, line 83: Why are you changing the bootstrap dependency to a version range instead of a pinned version?

jelhan (changes requested):
Thanks a lot for picking this up!

It seems you missed deleting the `docs/yarn.lock`. Can you please double check?

And we should configure RenovateBot to dedupe dependencies with pnpm. Deduplicating dependencies for yarn is done by the shared preset: https://github.com/kaliber5/renovate-config/blob/e63aaa1ed189d3aedfcbe79e909e80c7f26bcac2/default.json#L29 For PNPM we need to add the configuration ourselves. RenovateBot documentation on the option is available here: https://docs.renovatebot.com/configuration-options/#postupdateoptions

I assume GitHub actions CI workflow is based on Ember CLI blueprints, correct?

package.json
SanderKnauff: Ah, that must have happened when I tried fixing the `package.json` after a local ember-try run went wrong. I've reverted it.

SanderKnauff:
> Thanks a lot for picking this up!
> 
> It seems you missed deleting the `docs/yarn.lock`. Can you please double check?
> 
> And we should configure RenovateBot to dedupe dependencies with pnpm. Deduplicating dependencies for yarn is done by the shared preset: https://github.com/kaliber5/renovate-config/blob/e63aaa1ed189d3aedfcbe79e909e80c7f26bcac2/default.json#L29 For PNPM we need to add the configuration ourselves. RenovateBot documentation on the option is available here: https://docs.renovatebot.com/configuration-options/#postupdateoptions
> 
> I assume GitHub actions CI workflow is based on Ember CLI blueprints, correct?

Partially, I've tried to keep the current config as much as possible as I was not sure of the implications of larger changes.
The biggest change might be that the `@wyvox/action` handles the checkout and workspace setup.

I've also added the dedupe to `renovate.json` but kept the current preset so we don't have to copy the entire file.

jelhan (approved):
Thanks a lot!

PR #2031, opened 2023-11-26
Update bs-dropdown and related components to Glimmer component

SanderKnauff:


addon/components/bs-dropdown/menu.js
```diff
@@ -85,38 +87,32 @@ export default class DropdownMenu extends Component {
...
+      this.isOpen = value;
     });
-    return value;
   }
 
-  _isOpen = false;
   flip = true;
+
+  @tracked
   _popperApi = null;
```
jelhan, line 114: I struggle understanding what `_popperApi` is used for. I haven't found a place at which it's read. Neither a place at which it's set. Could you please give me a hint?

jelhan (changes requested):
Thanks a lot!

One minor change request to keep an inline comment and one small question as I struggled understanding the existing code.

jelhan (approved):

PR #2032, opened 2023-11-27
Run pre-commit lint hook with pnpm

SanderKnauff:
It seems like this was missed during the yarn -> pnpm conversion (#2028).

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn run lint-staged
+pnpm run lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-pnpm run lint-staged
+pnpm exec lint-staged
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,6 +31,7 @@
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
+    "lint-staged": "lint-staged",
```
jelhan, line 34: This wouldn't be needed if we change to `pnpm exec`. Yarn didn't differentiate between running a script and executing a binary from an installed package. At least yarn v1. PNPM (and NPM) have different commands for it: `pnpm run` to execute a script and `pnpm exec` to execute a binary of an installed dependency.

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn run lint-staged
+pnpm run lint-staged
```
jelhan, line 4: See below for explanation.

```suggestion
pnpm exec lint-staged
```

jelhan (changes requested):
Thanks a lot!

I think we can skip this in changelog as PNPM conversion has not been released yet. Therefore not adding a label. Please let me know if you want to have it in changelog explicitly.

package.json
SanderKnauff: That's awesome! I did not know about `pnpm exec`. I've updated it.

jelhan (approved):

PR #2033, opened 2023-11-28
use pnpm instead of yarn in a few docs

craigteegarden:
adds on to https://github.com/ember-bootstrap/ember-bootstrap/pull/2028

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -31,7 +31,7 @@ Once the prep work is completed, the actual release is straight forward:
 * First, ensure that you have installed your projects dependencies:
 
 ```sh
-yarn install
+pnpm install
 ```
 
 * Second, ensure that you have obtained a
@@ -50,7 +50,7 @@ yarn install
 * And last (but not least 😁) do your release.
 
 ```sh
-npx release-it
+pnpm release-it
 ```
 
 [release-it](https://github.com/release-it/release-it/) manages the actual
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,7 +8,7 @@ You will need the following things properly installed on your computer.
 
 * [Git](https://git-scm.com/)
 * [Node.js](https://nodejs.org/)
-* [Yarn](https://yarnpkg.com/)
+* [pnpm](https://pnpm.io/)
 * [Ember CLI](https://cli.emberjs.com/release/)
 * [Google Chrome](https://google.com/chrome/)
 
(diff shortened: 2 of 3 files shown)


RELEASE.md
```diff
@@ -50,7 +50,7 @@ yarn install
 * And last (but not least 😁) do your release.
 
 ```sh
-npx release-it
+pnpm release-it
```
jelhan, line 53: The PNPM equivalent to execute a binary of a package installed locally is `pnpm exec`.

```suggestion
pnpm exec release-it
```

docs/README.md
```diff
@@ -37,8 +37,8 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Linting
 
-* `yarn lint`
-* `yarn lint:fix`
+* `pnpm lint`
+* `pnpm lint:fix`
```
jelhan, line 41: To execute a script, you need to use `pnpm run` as far as I know.


```suggestion
* `pnpm run lint`
* `pnpm run lint:fix`
```

jelhan (changes requested):
Thanks a lot. I have two minor change requests. Please double check if they make sense. I haven't tested. Just based on what I think I recall about PNPM commands.

docs/README.md
craigteegarden: according to their site (https://pnpm.io/pnpm-cli#commands)

`npm run <cmd>` is equivalent to `pnpm <cmd>`

I do think using `pnpm run` will work, so I can change it if you would prefer that, but I did try it locally and `pnpm lint` works as expected.

RELEASE.md
craigteegarden: I believe either of these will work, according to their docs for the `exec` command, it isn't required if there is no conflict with a built-in pnpm command.

https://pnpm.io/cli/exec
Since either way works, I think it is more of a preference. I can do whichever matches the project's styles better.

jelhan: Thanks a lot for clarifying. I guess I mixed it up with NPM.

Ember CLI blueprints are using the shorthand without `run`: https://github.com/ember-cli/ember-cli/blob/d3b19b5f3db1c811273306f493e1ea3da0c70c74/blueprints/addon/index.js#L180 I think we should follow that convention.

docs/README.md
jelhan: See above.

jelhan (approved):
Thanks a lot!

PR #2034, opened 2023-12-03
Update bs-tab and related components to Glimmer component

SanderKnauff:


jelhan (changes requested):
Thanks a lot. I have only two minor change requests.

jelhan (approved):

Issue #2012, opened 2023-11-12
Refactor to Glimmer components


jelhan:
The addon is still using `@ember/component` for many components. We should refactor to `@glimmer/component`.

Any support is highly appreciated. Let me know if you can pick up one component (or more). I recommend doing only one or a few closely related components in the same PR to ease review and speed up getting it in.

I hope that it does not require any breaking changes as we have prepared for that refactoring long ago.

List of components, which are still using `@ember/component`:

- [x] `<BsCarousel>` (#2038)
- [x] `<BsCarousel::Slide>` (#2038)
- [x] `<BsDropdown>` (#2031)
- [x] `<BsDropdown::Menu>` (#2031)
- [x] `<BsDropdown::Menu::Divider>` (#2031)
- [x] `<BsDropdown::Menu::Item>` (#2031)
- [x] `<BsDropdown::Toggle>` (#2031)
- [x] `<BsLinkTo>` (#2043)
- [x] `<BsNav>` (#2020)
- [x] `<BsNav::Item>` (#2043)
- [x] `<BsNavbar>` (#2025)
- [x] `<BsNavbar::Content>` (#2025)
- [x] `<BsNavbar::Toggle>` (#2025)
- [x] `<BsProgress>` (#2030)
- [x] `<BsProgress::Bar>` (#2030)
- [x] `<BsTab>` (#2034)
- [x] `<BsTab::Pane>` (#2034)

SanderKnauff:
For the `BsNav::Item` and `BsLinkTo` I ran into an issue with the way parent/child lookups are being done.

In most of the new code, we pass through `registerChild/unregisterChild` functions from the parent to the child.
The old code relies on Ember for finding the nearest component of a type using the `nearestOfType` function.

Due to the way Bs::Nav yields children, it is not possible to have it exactly the same way in glimmer. Namely, the following situation will cause issues:
```hbs
<BsNav as |nav|>
  <nav.item>
    <nav.link-to @route='application'>Dummy</nav.link-to>
  </nav.item>
</BsNav>
```

We have no way of telling the `nav.link-to` that it should register itself as a child of `nav.item`.

What should we do in this situation?

jelhan:
Thanks a lot for investigating.

I think the root cause is that the `<BsLinkTo>` component is yielded from `<BsNav>`. But it should be yielded from `<BsNav::Item>` as it is expected to be used as a child of that one.

[...]

PR #2040, opened 2023-12-18
Drop active and disabled classes from . nav-item only expected for BS3

SanderKnauff:
The active and disabled parameters were used in the styling of Bootstrap 3. Since this parent/child component relationship is hindering the switch to Glimmer components and we do not support Bootstrap 3 anymore, we assume it can safely be removed.

For more information, see the discussion in #2012


jelhan (approved):
Thanks a lot!

jelhan:
I consider this a bug fix as it removes HTML markup not expected by BS3 and BS4.

PR #2041, opened 2023-12-19
Run tests on Safari in CI (again)

### LINKED ISSUE #2035: Run tests on Safari again
We aren't running our test suite on Safari anymore. BrowserStack tests are disabled since some time. https://github.com/ember-bootstrap/ember-bootstrap/blob/9e7c590cb01370c786f37e1245710d16e65dd98e/.github/workflows/ci.yml#L102-L127

We should run tests on Safari again in CI. Maybe using BrowserStack again. Maybe another service. We may want to have a look why we disabled BrowserStack in the first place.

In case we cannot enable tests on Safari again, we may want to update our compatibility note. Currently it claims:

> Ember Bootstrap works and is fully tested with (...) all modern evergreen browsers (Chrome, Firefox, Safari, Edge)

SanderKnauff:
Out of curiosity I tried to use Github CI for testing on Safari. It seems to work fine with a basic configuration.

I'm not really too familiar with MacOS, so let me know what you think.

Closes #2035


SanderKnauff:
I just noticed I forgot to add the actual configuration to instruct it to use the Safari testem, so I need to tinker around with it more

SanderKnauff:
It seems to be working now!

testem.safari.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+/* eslint-env node */
+module.exports = {
+  test_page: 'tests/index.html?hidepassed',
+  disable_watching: true,
+  parallel: 1,
```
jelhan, line 7: Do I assume correctly that we need a separate `testem.safari.js` and cannot include it into the `testem.js` because we need to set `parallel = 1` only for Safari?

jelhan (approved):
Thanks a lot! This is awesome!

testem.safari.js
SanderKnauff: I wanted the Safari runner only to run in the MacOS configuration. Placing it in the normal `testem` file would probably cause it to attempt to run on the Linux CI actions as well. 
This setup should be similar to how BrowserStack worked in the past.

There could be a way to have this all in the same file, but this is just the one I stumbled on that worked :sweat_smile:

Issue #1615, opened 2021-09-17
Enable tests for `@onClick` action of `<BsButton>` throws again


simonihmig:
#1581 added tests using custom assertion `assert.rejectsErrorNotHandledByEmberOnError` to test cases of rejected promises. These started to make [CI fail](https://github.com/kaliber5/ember-bootstrap/actions/runs/1244130788) for jobs not using our locked dependencies. Probably some updates to qunit changed some behaviour here!?

@jelhan

jelhan:
See #1616 for tests, which are skipped due to issues with testing Ember code, which throws.

PR #2055, opened 2023-12-21
Remove RSVP

### LINKED ISSUE #2051: Drop RSVP dependency
The `rsvp` dependency should not be needed anymore. It was primarily a polyfill for native `Promise`. All modern browsers support native `Promise` these days. We can simplify our code base and prepare for [planned deprecation](https://github.com/emberjs/rfcs/issues/832) by removing the dependency.

@NullVoxPopuli started the work in #1875 but doesn't have the time to finish it.

RSVP is not used at many places. GitHub code search [lists 11 places](https://github.com/search?q=repo%3Aember-bootstrap%2Fember-bootstrap%20rsvp&type=code).

Main priority should be on code shipped to end user. This affects 3 places:

1. `transitionEnd` utility function (`addon/utils/transition-end.js`)
2. `<BsForm>` component (`addon/components/bs-form.js`)
3. Blueprints (`blueprints/ember-bootstrap/index.js`)
4. Having `rsvp` as a dependency rather than a dev dependency.

The other 7 places affects only code, which is not shipped to consumers:

- docs app (2 files)
- tests (4 files)
- node tests (1 file)

In most cases the transition should be fairly simple. `RSVP.Promise`, `RSVP.reject`, and `RSVP.resolve` can be replaced with their native equivalents. Only `RSVP.defer` does not have a native equivalent. But it could be replaced by an utility function, which I sketched here: https://github.com/ember-bootstrap/ember-bootstrap/pull/1875#issuecomment-1483922699

SanderKnauff:
Continuation of #1875
Fixes #2051


tests/helpers/setup-unhandled-exception-listener.js
```diff
@@ -0,0 +1,44 @@
+import QUnit from 'qunit';
```
jelhan, line 1: The existing `assert-errors.js` and this new test helper seem addressing the same problem. Both are even taking the same approach of overwriting `QUnit.onUncaughtException`. Main difference seems to be that yours also adds an event handler for `unhandledrejection`, which prevents the default error handling. Beside that the public API of both is different as well.

Did you tried using the existing one? Maybe extending it with the `unhandledrejection` event listener.

I fear having to rest helpers for the same issue complicates the code base unnecessarily.

jelhan (changes requested):

tests/helpers/setup-unhandled-exception-listener.js
SanderKnauff: I did not notice the existing one. Strangely enough, there seems to be a difference between `Qunit.onUncaughtException` and `Qunit.onUnhandledRejection`. The second did not seem to work in this case, so I updated the code in `assert-error.js` to use that. It did not seem to influence the other usage of the helper.

jelhan: I done a code search for the existing test helper. It seems to be not used at all.

It's imported in `tests/integration/components/bs-button-test.js`. But all tests using it are currently skipped. We have an open ticket about it: #1615 Let's try activating them again after landing the improved test helper from you.

jelhan (approved):

PR #2068, opened 2024-01-12
support ember-style-modifier v4

jelhan:
Support [`ember-style-modifier` v4](https://github.com/jelhan/ember-style-modifier/releases/tag/v4.0.0), which has been released earlier this week. Breaking changes are limited to supported Ember and node versions. No changes to public API of `{{style}}` modifier were introduced.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "ember-popper-modifier": "^2.0.0 || ^3.0.0 || ^4.0.0",
     "ember-ref-bucket": "^4.0.0 || ^5.0.0",
     "ember-render-helpers": "^0.2.0",
-    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0",
+    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "ember-popper-modifier": "^2.0.0 || ^3.0.0 || ^4.0.0",
     "ember-ref-bucket": "^4.0.0 || ^5.0.0",
     "ember-render-helpers": "^0.2.0",
-    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0",
+    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0",
```
simonihmig, line 64: Do we need to still support the older versions?

simonihmig (approved):

package.json
jelhan: `ember-style-modifier` is used by several addons. Often apps depend on it indirectly through more as one addon. If each of them only support few versions it can get hard or impossible to find a compatible combination. Easy to end up in dependency hell.

Supporting all these versions is not much overhead for us. The API uses in Ember Bootstrap hasn't changed through these versions.

PR #2073, opened 2024-01-15
Convert to monorepo

SanderKnauff:
This PR converts the project to a monorepo in preparation of the V2 conversion.
Part 1 and 2 of [the v2 addon porting guide](https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md) should be covered by this PR.

From my checks during development, we can still link the addon by it's filesystem path.

# Open questions:
Is this the folder structure we want for the monorepo or do we prefer something else?
The two other common structures that I see use:
1. the addon name for the addon directory (E.g. `./ember-bootstrap` instead of `./addon`)
2. The packages nested in a `./packages` directory (E.g. `./packages/addon`, `./packages/test-app`, `./packages/docs`)

# Todo list:
- [x] Fix failing cleanup stage of Node tests
- [x] Check if release scripts are still correct
- [x] Verify if documentation generates correctly
- [x] Verify if all project specific configuration files are in the right place


jelhan:
> Is this the folder structure we want for the monorepo or do we prefer something else? The two other common structures that I see use:
> 
>     1. the addon name for the addon directory (E.g. `./ember-bootstrap` instead of `./addon`)
> 
>     2. The packages nested in a `./packages` directory (E.g. `./packages/addon`, `./packages/test-app`, `./packages/docs`)

I like the current struture. I find it easy to navigate. But I'm not very opinionated on it.

package.json
```diff
@@ -1,150 +1,16 @@
...
-    "ember-render-helpers": "^0.2.0",
-    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0",
-    "findup-sync": "^5.0.0",
-    "fs-extra": "^11.0.0",
-    "resolve": "^1.18.1",
-    "silent-error": "^1.0.1",
-    "tracked-toolbox": "^1.2.3 || ^2.0.0"
+    "prepare": "husky install",
+    "test": "pnpm --filter 'test-app' test",
+    "test:ember": "pnpm --filter 'test-app' test:ember"
```
jelhan, line 6: Should we add `lint` and `lint:fix` scripts as well, which run on all three projects? Might be helpful to quickly resolve linting issues.

package.json
```diff
@@ -1,150 +1,16 @@
...
-    "ember-source": ">=4.8.0"
-  },
-  "engines": {
-    "node": "18.* || >= 20"
+    "lint-staged": "15.2.0"
   },
-  "publishConfig": {
-    "registry": "https://registry.npmjs.org"
+  "lint-staged": {
+    "*.js": "eslint --cache --fix"
```
jelhan, line 13: I think we should fix all fixable linting issues before committing. At least we should run Prettier for Glimmer templates as well. But that's independent from this PR. I only noticed when reviewing. And wanted to keep a reminder.

package.json
```diff
@@ -155,20 +21,13 @@
...
-    "edition": "octane"
-  },
-  "ember-addon": {
-    "configPath": "tests/dummy/config",
-    "versionCompatibility": {
-      "ember": ">=2.3.0"
-    },
-    "demoURL": "https://www.ember-bootstrap.com"
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org"
```
jelhan, line 25: Shouldn't the publish config be in `addon/package.json` only?

test-app/config/ember-try.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
+// eslint-disable-next-line n/no-unpublished-require
```
jelhan, line 3: Can't we fix it in ESLint configuration? Like telling ESLint that this file is not published?

test-app/package.json
```diff
@@ -0,0 +1,137 @@
...
+    "mocha": "10.2.0",
+    "prettier": "3.0.3",
+    "qunit": "2.20.0",
+    "qunit-dom": "3.0.0",
+    "release-it": "16.3.0",
+    "sinon": "17.0.1",
+    "tracked-toolbox": "^1.2.3 || ^2.0.0",
+    "webpack": "5.89.0"
+  },
+  "peerDependencies": {
```
jelhan, line 123: Can an app have peer dependencies?

test-app/package.json
```diff
@@ -0,0 +1,137 @@
...
+    "lint:hbs:fix": "ember-template-lint . --fix",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "nodetest": "mocha node-tests --recursive",
+    "start": "ember serve",
+    "testa": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
+    "test:ember": "ember test",
+    "test:ember-compatibility": "ember try:each"
+  },
+  "dependencies": {
```
jelhan, line 37: Apps should have `devDependencies` only, shouldn't they?

test-app/package.json
```diff
@@ -0,0 +1,137 @@
+{
+  "name": "test-app",
```
jelhan, line 2: It seems to be missing `private: true`. That might be helpful to prevent mistakes.

jelhan (commented):
Thanks a lot for picking this up. Doing this separately from v2 addon migration helps with reviewing for sure.

test-app/package.json
SanderKnauff: I think they can have both, but to be sure I've added all of them to devDependencies

test-app/package.json
SanderKnauff: It can, but probably doesn't make any sense to. I'll move them to the addon package

test-app/config/ember-try.js
SanderKnauff: I've just disabled the rule for these specific cases. It was only actually needed in the node-tests anyway

package.json
SanderKnauff: That indeed sounds interesting. Does lint-staged also support modifying staged files?

docs/package.json
```diff
@@ -90,11 +101,10 @@
   },
   "ember-addon": {
     "paths": [
-      ".."
+      "../addon"
```
simonihmig, line 104: I think we can drop that! IIRC that's for in-repo addons before we had monorepos, wasn't it? Anyways, we should add the addon as a devDependency and let pnpm do the cross linking:

```
"ember-bootstrap": "workspace:*".
```

.github/workflows/ci.yml
```diff
@@ -79,18 +81,30 @@ jobs:
...
       - name: Checkout code
         uses: wyvox/action@v1
       - name: Test
+        working-directory: test-app
         run: pnpm ember try:one ${{ matrix.scenario }}
         env:
           BOOTSTRAPVERSION: ${{ matrix.bootstrap }}
           BROWSER: ${{ matrix.browser }}
 
+  test-node:
```
simonihmig, line 96: Great if this works! But just to mention: when migrating to v2, these tests can be dropped (as the feature they test will be dropped as well), so probably not worth spending much time on these, should they make any trouble now...

simonihmig (commented):
Great work here!

> the addon name for the addon directory (E.g. ./ember-bootstrap instead of ./addon)

I'd prefer this one! It matches what the v2 blueprint would output. Also it enables us to integrate the other ember-bootstrap-* addons into the monorepo should we decide to do so. And `addon/addon` is a bit weird :wink: My 2 cents...

.github/workflows/ci.yml
SanderKnauff: Ah, that would be nice. It feels a bit dirty to have it in the addon directory, but it's the only way that it will work

SanderKnauff:
The CI pipeline for docs seems to be working correctly. As sample can be found on: https://ember-bootstrap.sansk.ooo

SanderKnauff:
I removed the WIP status! The only thing that I have no knowledge of is the release and changelog scripts.
The configuration still exists, but I am not sure if it is pointing at the right place for now. How would I be able to test this?

docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-bootstrap-docs",
-  "version": "0.0.0",
+  "version": "6.1.0",
```
jelhan, line 3: Why do we need to bump the version? When do we need to bump it in future?

jelhan (commented):
I have a few remaining questions just for sanity check.

> The only thing that I have no knowledge of is the release and changelog scripts.
> The configuration still exists, but I am not sure if it is pointing at the right place for now. How would I be able to test this?

Release-it's dry run feature may allow testing this: https://github.com/release-it/release-it/blob/main/docs/dry-runs.md

package.json
```diff
@@ -1,150 +1,20 @@
...
-    "ember-render-helpers": "^0.2.0",
-    "ember-style-modifier": "^0.8.0 || ^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0",
-    "findup-sync": "^5.0.0",
-    "fs-extra": "^11.0.0",
-    "resolve": "^1.18.1",
-    "silent-error": "^1.0.1",
-    "tracked-toolbox": "^1.2.3 || ^2.0.0"
+    "prepare": "husky install",
+    "lint": "pnpm --filter * lint",
+    "lint:fix": "pnpm --filter * lint:fix",
```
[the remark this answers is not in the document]
SanderKnauff: Probably none. I did not know about the `--recursive` option, thanks!

ember-bootstrap/vendor/ember-bootstrap/bs3.css
```diff
@@ -0,0 +1,47 @@
+/* Fix for https://github.com/kaliber5/ember-bootstrap/issues/355 */
```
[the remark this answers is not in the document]
SanderKnauff: Probably, I've had to rebase / merge master into this branch a few times to keep up with renovate

.github/workflows/ci.yml
```diff
@@ -18,8 +18,12 @@ jobs:
     steps:
       - name: Checkout code
         uses: wyvox/action@v1
-      - name: Lint
+      - name: Lint addon
```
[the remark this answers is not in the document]
SanderKnauff: The test-app should be below that, but I've replaced this step with the top level lint that executes on all workspaces

docs/package.json
SanderKnauff: I actually think the version is not needed at all here. I've removed it since it is a private package anyway

SanderKnauff:
> I have a few remaining questions just for sanity check.
> 
> > The only thing that I have no knowledge of is the release and changelog scripts.
> > The configuration still exists, but I am not sure if it is pointing at the right place for now. How would I be able to test this?
> 
> Release-it's dry run feature may allow testing this: https://github.com/release-it/release-it/blob/main/docs/dry-runs.md

Ah yes, the dry-run helps a lot with understanding what happens. I'm currently testing it out with a local NPM repository using https://verdaccio.org/. This way I can just try out 'real' releases without it attempting to go to NPM.

package.json
jelhan: I'm also not that experienced with [lint-staged](https://github.com/lint-staged/lint-staged) to be honest.

I think we are capturing majority of the files being changed in regular commits with our existing linters. So maybe we just want to run linting _always_ regardless and skip filtering on specific files?

I wonder if something like this would work:

```suggestion
    "*": "pnpm run lint:fix"
```

We should quickly check how long it takes. For me lint-staged and similar pre-commit hooks are more annoying than helpful. I have configured my editor to lint on save anyways. But if I recall correctly we primarily added it to reduce barrier for new contributors.

SanderKnauff: The fixing works, but after testing, it seems that using `"*"` as lint-staged target did not work as expected.

I've changed the configuration now so it will run eslint and ember-template-lint. We should include Typescript when we have that as well.

I think that further modifications to the lint-staged configuration should go in a separate PR, the monorepo migration is already difficult to oversee 😅

SanderKnauff:
> > I have a few remaining questions just for sanity check.
> > > The only thing that I have no knowledge of is the release and changelog scripts.
> > > The configuration still exists, but I am not sure if it is pointing at the right place for now. How would I be able to test this?
> > 
> > 
> > Release-it's dry run feature may allow testing this: https://github.com/release-it/release-it/blob/main/docs/dry-runs.md
> 
> Ah yes, the dry-run helps a lot with understanding what is happening during a release. I'm currently testing it out with a local NPM repository using https://verdaccio.org/. This way I can just try out 'real' releases without it attempting to go to NPM.

The release scripts seemed to work correctly in my simulated tests. I was able to deploy to a local NPM registry.
I did not test the GitHub integration though, as I did not want to modify this repository by accident. Nothing has been changed in that aspect though, so I don't think it will break.

The most important configuration is in place, so I don't think even if it is broken, it will take long to fix.

jelhan (approved):
Thanks a lot for doing all this work! I think we can (and should) accept remaining uncertainty.

ember-bootstrap/ember-cli-build.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+const { maybeEmbroider } = require('@embroider/test-setup');
+
+module.exports = function (defaults) {
```
simonihmig, line 6: It seems you deliberately added this here in one of the last commits, but I wonder why? I don't think this is ever going to be used, as this file is purely app-related, and the app-part is now extracted into its own package.

simonihmig (approved):
One minor comment, other than that this looks great to me! Thanks a ton, Sander! :pray:

ember-bootstrap/ember-cli-build.js
SanderKnauff: I added that while debugging the build issues, but forgot to remove it.
I've tested again with and without the file, just to be sure, and it does not make a difference, so I've removed it again.

PR #2077, opened 2024-01-29
Fix several issues with BsCarousel component

SanderKnauff:
The animations in `BsCarousel` have been broken since the Glimmer conversion.
It a few `@tracked` annotations have been forgotten and some `this.set` usages were incorrectly converted, which causes the slide animations to break. 
During the fixes, I also found some BS4/BS5 compatibility issues which now have been fixed

This fixes the following:

* Carousel not animating (Tracking issue)
* Carousel slide indicators not appearing as numbers when using Bootstrap 5
* Carousel slide indicators causing the page to scroll back to the top   
* Carousel controls showing 'next' and 'previous' labels when using Bootstrap 4. These should only be visible for screen readers.
* Fixes line ending in `bs-carousel.hbs` and `bs-carousel/slide.hbs` (For reviewing these files, I recommend turning off whitespace changes in the diff view)


jelhan (changes requested):
Thanks a lot for fixing all those defects!

Would be great if you could add test coverage for the features to ensure that we won't break them again in the next refactoring.

SanderKnauff:
> Thanks a lot for fixing all those defects!
> 
> Would be great if you could add test coverage for the features to ensure that we won't break them again in the next refactoring.

I've adapted some tests to check for classes during transitions. This combined with the fixes I had to do to make the tests aware of the difference between Bs4 and Bs5 should cover the changes

ember-bootstrap/addon/components/bs-carousel.hbs
```diff
@@ -1,47 +1,66 @@
...
+  {{on "mouseleave" this.handleMouseLeave}}
+  {{did-update this.childSlidesObserver this.childSlides this.autoPlay}}
+  {{did-update this.indexObserver this.index}}
+>
+  {{#if this.showIndicators}}
+    {{#if (macroCondition (macroGetOwnConfig "isBS5"))}}
+      <div class="carousel-indicators">
+        {{#each this.indicators as |indicator _index|}}
+          <button
+            data-bs-target
```
jelhan, line 17: What is this data attribute used for?

ember-bootstrap/addon/components/bs-carousel.hbs
```diff
@@ -1,47 +1,66 @@
...
+            type="button"
+            class={{if (bs-eq this.currentIndex _index) "active"}}
+            aria-current="{{bs-eq this.currentIndex _index}}"
+            {{on "click" (fn this.toSlide _index)}}
+          />
```
jelhan, line 22: The button is missing the `aria-label` attribute, which should be set accordingly to [Bootstrap docs](https://getbootstrap.com/docs/5.3/components/carousel/#indicators). But as that's not a regression, we can address it in a follow-up. To not forget about it, we should create a GitHub issue for it.

ember-bootstrap/addon/components/bs-carousel.hbs
SanderKnauff: It's required for the indicator CSS to work.
This is the relevant Boostrap 5 stylesheet: https://github.com/twbs/bootstrap/blob/11272eca59e426906f1ee105371fe8477266c24b/scss/_carousel.scss#L169

ember-bootstrap/addon/components/bs-carousel.js
```diff
@@ -228,6 +228,8 @@ export default class Carousel extends Component {
    * @private
    * @type string
    */
+  @tracked
+  orderClassName = this.args.orderClassName ?? null;
```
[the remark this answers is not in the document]
SanderKnauff: There were more of those, so I have removed them there as well

ember-bootstrap/addon/components/bs-carousel.hbs
SanderKnauff: I've created #2079 to track the issue

jelhan (approved):
Looks good to me.

PR #2078, opened 2024-01-31
upgrade husky to v9 per migration guides

jelhan:
Following upgrade guide for Husky v9: https://github.com/typicode/husky/releases/tag/v9.0.1

Husky v9 is backward compatible. The current configuration works. This is only to be future proof. And because I just had read that changelog entry when merging the upgrade MR.


SanderKnauff (approved):
Looks good to me!

I've checked the docs as well and it does seem like a pretty safe upgrade.

Issue #2064, opened 2024-01-08
Deprecate inserting wormhole target


jelhan:
Ember Bootstrap inserts a `<div id="ember-bootstrap-wormhole"></div>` into consuming app's `index.html` unless `insertEmberWormholeElementToDom` is `false`. The element is later used by Ember Bootstrap as wormhole target.

We decided to drop this functionality when migrating to a v2 addon (#1980). We should deprecate it as soon as possible educating consumers about the path forward.

[Our documentation](https://www.ember-bootstrap.com/getting-started/setup) already documents how consumers can insert a wormhole target themselves:

> [insertEmberWormholeElementToDom] Will automatically insert a div into your index.html, where Bootstrap's modals and tooltips will be rendered to. If you set this to false, you will have to add an `<div id="ember-bootstrap-wormhole"></div>` element by yourself, e.g. into your `application.hbs`. 

Ember Bootstrap should log a build-time deprecation when it inserts the element into the build educating the consumer about the path forward. It could be implemented here: https://github.com/ember-bootstrap/ember-bootstrap/blob/c963f59ed28345655fbd1e7eb437fa0163d4c622/index.js#L230-L238

The deprecation message may be something like this:

> Inserting the element used by Ember Bootstrap as wormhole target for modals and tooltips is deprecated. Please add a `<div id="ember-bootstrap-wormhole"></div>` to either the `app/index.html` or `app/templates/application.hbs` and set Ember Bootstrap's `insertEmberWormholeElementToDom` configuration option to `false`. Please find more details about Ember Bootstrap's configuration here: https://www.ember-bootstrap.com/getting-started/setup

jelhan:
Implemented by #2083.

PR #2083, opened 2024-02-01
Deprecate inserting wormhole target

SanderKnauff:
#2064

### DIFF
--- a/docs/app/templates/getting-started/setup.hbs
+++ b/docs/app/templates/getting-started/setup.hbs
@@ -6,6 +6,15 @@
 
 <pre class="highlight">ember install ember-bootstrap</pre>
 
+<p>
+  To allow Bootstrap to render modals and tooltips, add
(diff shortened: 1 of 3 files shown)


simonihmig (approved):
LGTM, thank you!

I'll remove the `breaking` label though, as I think we only need to flag this as breaking when actually removing the feature in the next major, right?

SanderKnauff:
> LGTM, thank you!
> 
> I'll remove the `breaking` label though, as I think we only need to flag this as breaking when actually removing the feature in the next major, right?

Sounds good! Should we create a checklist so we don't forget to remove the things we drop support for in 7.0.0?

jelhan:
> > LGTM, thank you!
> > I'll remove the `breaking` label though, as I think we only need to flag this as breaking when actually removing the feature in the next major, right?
> 
> Sounds good! Should we create a checklist so we don't forget to remove the things we drop support for in 7.0.0?

Creating a tracking issue for v7 would be great!

PR #2087, opened 2024-02-02
Convert docs components to native classes

SanderKnauff:
This resolves a number of linting warnings in the docs app.
It is primarily focused on fixing the `ember/no-classic-classes` lint rule.

There is also a small fix for the rendering of the markdown changelogs, due to `ember-cli-showdown` 8.0.0 removing the ability to use positional parameters.

### DIFF
--- a/docs/app/models/login.js
+++ b/docs/app/models/login.js
@@ -1,6 +1,7 @@
 // BEGIN-SNIPPET login-model
 import EmberObject from '@ember/object';
 import { validator, buildValidations } from 'ember-cp-validations';
+import { tracked } from '@glimmer/tracking';
 
 const Validations = buildValidations({
   password: [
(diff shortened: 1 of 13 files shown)


SanderKnauff:
I see that my merge from master into my working branch caused some chaos. I'm going to rebuild the branch to clean it up a bit.

docs/app/models/login.js
```diff
@@ -18,9 +19,8 @@ const Validations = buildValidations({
   email: [validator('presence', true), validator('format', { type: 'email' })],
 });
 
-export default EmberObject.extend(Validations, {
-  email: null,
-  password: null,
-  rememberMe: false,
```
jelhan, line 24: Good catch that `rememberMe` wasn't actually used in the template. :clap:

docs/app/templates/application.hbs
```diff
@@ -1,85 +1,78 @@
-{{#unless this.hideNav}}
```
jelhan, line 1: Good catch that `hideNav` could never be `true` as there isn't any route containing `'acceptance'`. :clap:

jelhan (approved):
Looks good to me!

Thanks a lot for addressing this long outstanding technical debt!

PR #2091, opened 2024-02-10
Deprecate `importBootstrapCSS`, remove `importBootstrapTheme` and `importBootstrapFont`

SanderKnauff:
The CSS and SASS configuration should be manually configured by the consuming addon.
`importBootstrapTheme` and `importBootstrapFont` have been removed instead of deprecated as these only exists for Bootstrap 3, which we do not support.

### DIFF
--- a/docs/app/templates/getting-started/assets.hbs
+++ b/docs/app/templates/getting-started/assets.hbs
@@ -28,14 +28,27 @@
 
 <p>
   When not using a CSS preprocessor (see below) you will need to import the
-  static Bootstrap CSS into your ember-cli build:
+  static Bootstrap CSS into yourself.
+</p>
+<p>
+  For classic apps, add the following lines to
+  <code>ember-cli-build.js</code>
+
+  {{! prettier-ignore }}
+  <pre class="highlight"><code>//your-bootstrap-app/ember-cli-build.js
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, { /* Configuration */ });
+  app.import('node_modules/bootstrap/dist/css/bootstrap.css');
+  app.toTree();
+};</code></pre>
 </p>
 
-<div>
+<p>
+  For Embroider apps, import the CSS file in
+  <code>app/app.js</code>.
   {{! prettier-ignore }}
-  <pre class="highlight"><code>// app/app.js
-import 'bootstrap/dist/css/bootstrap.css';</code></pre>
-</div>
+  <pre class="highlight"><code>import 'bootstrap/dist/css/bootstrap.css';</code></pre>
+</p>
 
 <h2>Using a CSS preprocessor</h2>
 
--- a/docs/app/templates/getting-started/assets.hbs
+++ b/docs/app/templates/getting-started/assets.hbs
@@ -27,21 +27,17 @@
 <h2 id="importing-css-assets">Importing static CSS Assets</h2>
 
 <p>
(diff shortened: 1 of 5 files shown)


docs/app/templates/getting-started/assets.hbs
```diff
@@ -27,21 +27,18 @@
...
 <div>
+  {{! prettier-ignore }}
   <pre class="highlight"><code>//your-bootstrap-app/ember-cli-build.js
-      module.exports = function(defaults) { let app = new EmberApp(defaults, {
-      'ember-bootstrap': { 'importBootstrapCSS': false } }); return
-      app.toTree(); };
-    </code></pre>
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, { /* Configuration */ });
+  app.import('node_modules/bootstrap/dist/css/bootstrap.css');
```
simonihmig, line 39: I don't think we should suggest folks to use `app.import()` as that is a legacy EmberCLI API. I would rather suggest to use the "new" way using the ember-auto-import/webpack bundling pipeline, which is also how things would work natively under Embroider. Which is to import the CSS in JS code and let webpack handle the rest.

Haven't tried this out, but I think if you add this to your app's `app.{j,t}s`, then it _should_ work (we need to confirm!):

```
import 'bootstrap/dist/css/bootstrap.css';
```

SanderKnauff: Aha, I was not aware of that. Luckily your suggestion seems to work like a charm both in a classic and Embroider build.

jelhan: Ember CLI docs document `app.import()` to pull in third-party CSS: https://guides.emberjs.com/release/addons-and-dependencies/#toc_css I think we should point consumers to that official documentation.

I'm okay with an additional note that this may change in future. And different patterns available today already.

I see the main intent for deprecating this (beside unblocking v2 conversion) in getting rid of the responsibility. I fear we would somehow keep them if recommending an unofficial CSS import pattern.

jelhan: > Ember CLI docs document `app.import()` to pull in third-party CSS: https://guides.emberjs.com/release/addons-and-dependencies/#toc_css I think we should point consumers to that official documentation.
> 
> I'm okay with an additional note that this may change in future. And different patterns available today already.
> 
> I see the main intent for deprecating this (beside unblocking v2 conversion) in getting rid of the responsibility. I fear we would somehow keep them if recommending an unofficial CSS import pattern.

P.S.: Upgrading Ember CLI docs is likely best mid- to long-term solution here. 😅

jelhan (approved):
Looks good to me.

docs/app/templates/getting-started/assets.hbs
```diff
@@ -27,22 +27,28 @@
...
-</div>
+module.exports = function(defaults) {
+  let app = new EmberApp(defaults, { /* Configuration */ });
+  app.import('node_modules/bootstrap/dist/css/bootstrap.css');
+  app.toTree();
+};</code></pre>
+</p>
+
+<p>
+  For Embroider apps, import the CSS file in
```
simonihmig, line 47: This is not depending on the build system, because both patterns work both under classic and Embroider. It's more about legacy vs. next-gen pattern IMHO. But we can tweak this later...

simonihmig (approved):

PR #2092, opened 2024-02-10
Add initial typescript configuration to addon project

SanderKnauff:
This adds Typescript support to the addon project.
At this moment, it does not touch the test-app and docs projects yet.

### DIFF
--- a/ember-bootstrap/.eslintrc.js
+++ b/ember-bootstrap/.eslintrc.js
@@ -2,7 +2,7 @@
 
 module.exports = {
   root: true,
-  parser: '@babel/eslint-parser',
+  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 'latest',
     sourceType: 'module',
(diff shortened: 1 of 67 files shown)


simonihmig (approved):
LGTM! Any reason to have this tagged as WIP actually? 

We don't expose the types in the published package yet, [that stuff](https://github.com/ember-cli/ember-cli/blob/v5.4.1/blueprints/addon/index.js#L95-L107) is still missing. But I assume this was intentional, to only expose the types when we are ready to?

SanderKnauff:
> LGTM! Any reason to have this tagged as WIP actually?
> 
> We don't expose the types in the published package yet, [that stuff](https://github.com/ember-cli/ember-cli/blob/v5.4.1/blueprints/addon/index.js#L95-L107) is still missing. But I assume this was intentional, to only expose the types when we are ready to?

I put it as WIP because I wasn't yet sure if I had missed anything, such as the `typesVersions` object that I actually did miss. 😅 

Shall I add it in? I'm not sure if it would be a big issue if we ship partial types until we're completely done.

simonihmig:
>  I'm not sure if it would be a big issue if we ship partial types until we're completely done.

Yeah, probably it doesn't hurt really. So we can add that I guess!

simonihmig:
I think we will need some more tweaking of the config. But this is better done when we actually have some TS code to play with. I am currently working on some initial TS conversion, so will probably bring up some tweaks with the next PR! 

So let's get this merged first! Thanks @SanderKnauff!

simonihmig:
> so will probably bring up some tweaks with the next PR

See https://github.com/ember-bootstrap/ember-bootstrap/pull/2093/commits/e641f84b77f8ca932ad0c0e386a9f26c092b8912

PR #2093, opened 2024-02-10
Convert BsButton to TypeScript

simonihmig:
Converts our "leaf-most" component to TS, including type checking of its template and emitting declarations.

When reviewing, check [this commit](https://github.com/ember-bootstrap/ember-bootstrap/pull/2093/commits/e641f84b77f8ca932ad0c0e386a9f26c092b8912) specifically, as it builds on top of #2092 to tweak the tsconfig such that emitting declarations with Glint actually work...


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "glint.libraryPath": "ember-bootstrap"
```
simonihmig, line 2: The Glint VSCode plugin expect to find the glint libraries in the root's node_modules, where they aren't due to pnpm not hoisting dependencies. So this is making the plugin work ootb when setting up the repo locally. Normally, I wouldn't but code editor configs into the repo, but in this case maybe it makes sense?

simonihmig:
When reviewing, check [this commit](https://github.com/ember-bootstrap/ember-bootstrap/pull/2093/commits/e641f84b77f8ca932ad0c0e386a9f26c092b8912) specifically, as it builds on top of #2092 to tweak the tsconfig such that emitting declarations with Glint actually work...

jelhan (changes requested):
I feel we should convert the corresponding tests to TypeScript as well. Without that we don't have test coverage for the types, which makes reviewing and maintaining the type signatures more difficult.

jelhan:
> I feel we should convert the corresponding tests to TypeScript as well. Without that we don't have test coverage for the types, which makes reviewing and maintaining the type signatures more difficult.

I started converting `<BsAccordion>` to TypeScript. Including the tests. And noticed late that the test app is not setup for TypeScript yet. I feel we should do that as another preparation step before.

simonihmig:
I agree in general, but I wonder if we need to have everything in place in a single PR. I think we will only be able to call the types to be safe and stable when we have converted _everything_, all components and all tests. Like a future PR that converts ButtonGroup for example could reveal that some (internal) button types were missing or wrong. But making these changes land sooner than later unblocks those other migrations, as we have to convert basically from the leaf components upwards (or downwards, however you think of the orientation of that tree :wink:). So not sure if we need to hold back iterative improvements like this one?

jelhan:
> I agree in general, but I wonder if we need to have everything in place in a single PR.

I don't think we need to have everything in one single PR. Actually I think we should split it into small chunks. I just have bad experience from splitting feature implementation and tests for it in separate PRs.

I see that there could be edge cases, in which converting the corresponding tests with the component to TypeScript may not be easy. E.g. if the tests render another component, which has not been converted yet for testing compatibility. I agree that we need to be pragmatic in those cases. But that doesn't seem to be the case for `<BsButton>`.

@SanderKnauff already created a PR converting the test app to TypeScript. This should unblock converting corresponding types together with the components. I hope I find the time later today pulling it into #2095. Getting some real-world experience on overhead due to converting tests as well, will likely help us with decision making.

simonihmig:
Just merged #2096, so will update this PR to have the tests covered as well!

jelhan:
> Just merged #2096, so will update this PR to have the tests covered as well!

Have a look at #2095. I converted some test infrastructure, which you may need as well, to TypeScript already. Especially the `test-app/tests/helpers/bootstrap`. Not sure if the types are correct. I only done minimal work to get it passing.

simonihmig:
> I converted some test infrastructure, which you may need as well, to TypeScript already.

Oh no, just did that as well! 🙈 Should looked at your work first 😅

jelhan (approved):
Looks good to me.

simonihmig:
CI is finally happy.! :tada:

Feel free to have a final look, but I hope it's good now! Keeping this open for a bit, in case @SanderKnauff you want to also review before merging. But besides the actual button-related migration, this also contains a bit of infrastructure (test-helpers, CI/scripts setup) that is needed for follow-up migrations, so feel free to hit the merge button when anyone needs this to unblock their work! :)

ember-bootstrap/addon/components/bs-button.hbs
```diff
@@ -1,8 +1,14 @@
-{{! @glint-nocheck }}
 <button
   disabled={{this.__disabled}}
   type={{if @attrTypePrivateWorkaround @attrTypePrivateWorkaround "button"}}
-  class="btn {{if @active "active"}} {{if (macroCondition (macroGetOwnConfig "isNotBS5")) (if this.block "btn-block")}} {{bs-size-class "btn" @size}} {{bs-type-class "btn" @type default="secondary" outline=@outline}}"
+  class="btn
```
SanderKnauff, line 4: While this does work, I think the newlines will be included while rendering.
Is this something we want to look out for?

ember-bootstrap/addon/utils/decorators/arg.ts
```diff
@@ -0,0 +1,33 @@
+import type Component from '@glimmer/component';
+
+type DecoratorPropertyDescriptor = PropertyDescriptor & {
+  initializer?: () => void;
+};
+
+/**
+ * @internal
+ */
+function arg(target: object, key: string): void;
```
SanderKnauff, line 10: With generics, is it possible to type args like `<T extends object>(target: T, key: keyof T)`?

SanderKnauff (changes requested):
Awesome start! Just a few small comments.
I do hope that after the initial boilerplate work is done, the changesets will become a bit more manageable.

ember-bootstrap/addon/utils/decorators/arg.ts
simonihmig: IIRC I tried that, but at some point I gave up. `key` has to be a keyof this.args actually, see L23. But given this is used as a decorator, which has its own expected signature (with TS actually disagreeing on our implementation), this was not working easily, or I was missing something...

ember-bootstrap/addon/components/bs-button.hbs
simonihmig: I just fixed that!

I think our prettier setup is not right, wanted to come up with a PR to fix this. With pnpm and no hoisting, we don't have prettier in the root node_modules, but VS Code expects it there to work with the prettier plugin. I believe as this was not working ootb, it would reformat the code, but not based on prettier. Had to run `lint:fix` sometimes, which is unexpected, as usually prettier-in-the-editor should be in-sync with prettier-in-lint-scripts.

simonihmig:
> I do hope that after the initial boilerplate work is done, the changesets will become a bit more manageable.

Sure! 😅

ember-bootstrap/addon/components/bs-button.hbs
SanderKnauff: hmm, should the prettier configuration be in the root project?

That aside, I did like the indentation as it was before, it's just a shame that it generates all those newlines. Maybe we can add a helper in the future that strips all the unneccesary whitespaces from such class lists.

ember-bootstrap/addon/utils/decorators/arg.ts
SanderKnauff: I'm not that familiar with creating decorators, so I believe you if you say that it's not working. :sweat_smile:

SanderKnauff (approved):

ember-bootstrap/addon/components/bs-button.hbs
jelhan: Having a build plugin which strips whitespaces would be great. But I'm not sure if something like that already exists in Ember ecosystem.

simonihmig: There is this: https://github.com/mainmatter/ember-hbs-minifier 

But not sure if it's able to only run on our addon. Also when migrating to v2, we would want to have something that runs ahead of time as part of our rollup build...

jelhan: Yeah. I have more thinking in a v2 addon world. I feel it's too risky and unnecessary churn introducing such tooling before v2 conversion.

PR #2094, opened 2024-02-10
[docs] Resolve remaining lint warnings

SanderKnauff:
All ESLint and ember-template-lint rules that were set to warn have now been enabled.

There were some remaining places where the lint issue could not be solved, but those places have been annotated with `{{! template-lint-disable }}` and a comment stating the reason for ignoring the lint rules.

### DIFF
--- a/docs/.eslintrc.js
+++ b/docs/.eslintrc.js
@@ -22,13 +22,6 @@ module.exports = {
   env: {
     browser: true,
   },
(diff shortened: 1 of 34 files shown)


jelhan (approved):
Looks good to me.

Thanks a lot for the detailed and precise comments in the source code. They helped a lot with reviewing.

PR #2095, opened 2024-02-11
WIP: convert BsAccordion to TypeScript

jelhan:
Converts `<BsAccordion>` to TypeScript. Including the tests. But the test app is not yet setup for TypeScript. Keeping it as work in progress until we landed that one in a preparation PR.


test-app/tests/integration/components/bs-collapse-test.ts
```diff
@@ -1,13 +1,26 @@
...
-import hbs from 'htmlbars-inline-precompile';
+import {
+  render,
+  settled,
+  waitFor,
+  type TestContext,
+} from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
 import { delay, test, visibilityClass } from '../../helpers/bootstrap';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
```
jelhan, line 11: Converting `setupNoDeprecations` test helper to TypeScript is the only task pending for this PR as far as I'm aware. @simonihmig: Are you already working on that for #2093?

test-app/tests/integration/components/bs-collapse-test.ts
```diff
@@ -44,18 +51,18 @@ module('Integration | Component | bs-collapse', function (hooks) {
...
-    this.actions.shown = shownAction;
+  test('setting collapse to false expands this item', async function (this: Context, assert) {
+    const showAction = sinon.spy();
+    const shownAction = sinon.spy();
+    this.show = showAction;
+    this.shown = shownAction;
 
     this.set('collapsed', true);
-    await render(
+    await render<Context>(
```
SanderKnauff, line 61: Something a colleague of mine came up with is to create a helper function to render the elements, while providing an object with the arguments. Would it make sense to introduce something like that here as well?

It usually looks something like this

```ts
renderWithArgs(args: BsCollapseSignature['Args']) {
  setAttributes(args) // I think this just loops over the properties of `args` and sets them using `this.set`.
  
  return render<BsCollapseSignature['Args']>(hbs`<BsCollapse
      {{! All arguments }}
    </BsCollapse>`);
}
```

This can then be used for all tests in the same class:

```ts
await renderWithArgs({
  show: false,
  shown: () => {},
  // etc...
});
```

This way the boilerplate of Glint context setting is handled by the helper function, and you don't need to bother with it in every `test` method.
Could something like this be interesting for `ember-bootstrap` as well?

Of course, if we decide to go for `gts`, the render context will just be the same as the function scope, but that might still be some time away.

SanderKnauff (changes requested):
Just one small question / idea that could make writing tests a bit easier. Let me know what you think!

It's great to see some action happening in this front.
I'll be sure to start working on some components myself as well!

test-app/tests/integration/components/bs-collapse-test.ts
jelhan: Interesting proposal!

Let's discuss it in #2093. I think we should land that one first. Will update the patterns here accordingly.

simonihmig: But you would also need to render arbitrary things specific to the test, not just the component invocation with some args, like block content, named blocks, yielded stuff etc., and you can't do that with a template that's inside that helper? Or do I misunderstand?

SanderKnauff: > But you would also need to render arbitrary things specific to the test, not just the component invocation with some args, like block content, named blocks, yielded stuff etc., and you can't do that with a template that's inside that helper? Or do I misunderstand?

Nope, that is one of the downsides. For components without block data it works like a dream...
I did forget that most components here do have yielded blocks, so it may not as useful as I initially thought 😅

jelhan:
I merged latest master with the setup from @simonihmig in #2093 into this one. All is working fine now. Ready for final review from my side.

ember-bootstrap/addon/components/bs-collapse.ts
```diff
@@ -233,6 +264,10 @@ export default class Collapse extends Component {
    * @protected
    */
   hide() {
+    if (!this._element) {
+      throw new Error('Reference to collapse div element is missing.');
```
simonihmig, line 268: Should we use `assert` from `@ember/debug` here? This would strip the code in a production build, but still make TS happy.

test-app/tests/integration/components/bs-collapse-test.ts
```diff
@@ -1,13 +1,26 @@
...
+  type TestContext,
+} from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
 import { delay, test, visibilityClass } from '../../helpers/bootstrap';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
 import a11yAudit from 'ember-a11y-testing/test-support/audit';
 import { skipTransition } from 'ember-bootstrap/utils/transition-end';
 import sinon from 'sinon';
 
+interface Context extends TestContext {
```
simonihmig, line 16: During my button PR I was actually thinking if we should introduce gts for the test-app already? It would eliminate the need for all that context ceremony just to please TS, as templates can directly access variables in scope, so you don't have to assign to `this`. 🤔

test-app/tests/test-helper.ts
```diff
@@ -7,10 +7,10 @@ import { start } from 'ember-qunit';
 import setupSinon from 'ember-sinon-qunit';
 import { setRunOptions } from 'ember-a11y-testing/test-support';
 
-setApplication(Application.create(config.APP));
-
 setup(QUnit.assert);
 
+setApplication(Application.create(config.APP));
```
simonihmig, line 12: Was there a reason to change this?

simonihmig (changes requested):
LGTM! The only thing I would like to see here is to also remove the no-check directive [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/7329f018566fadd7b54e31b32721772e82c228c8/ember-bootstrap/addon/components/bs-collapse.hbs#L1)

test-app/tests/integration/components/bs-collapse-test.ts
SanderKnauff: That might indeed be a good place to start including GTS. I don't think the setup would change that much.
Hopefully it will also allow for a bit neater template formatting. I find that template-lint can be a tad aggressive at times.

SanderKnauff (approved):
Looks good to me!

test-app/tests/integration/components/bs-collapse-test.ts
jelhan: I'm open to introduce gts. Especially for the tests. But I think we shouldn't couple/block this PR by it. I'm open to refactor of we land gts before merging this one though. At the end of the day it's just a question of we refactor to `<template>` tag before or after merging this PR.

Refactoring to gts has only one trade-off, which I'm aware of: We would loose test coverage for template registry as all components are imported explicitly. I'm happy to accept that trade-off. Especially as we gain test coverage for explicit imports at the same time.

test-app/tests/test-helper.ts
jelhan: I cannot recall any specific reason. :see_no_evil: Good catch. I reverted the change.

ember-bootstrap/addon/components/bs-collapse.ts
jelhan: Done in a8a06c4a2b61380768df14feeecc0219cf079a4d. Wasn't aware that `assert` from `@ember/debug` is able to narrow the type.

test-app/tests/integration/components/bs-collapse-test.ts
simonihmig: > But I think we shouldn't couple/block this PR by it

I didn't mean to say that. Just came to my mind again after reviewing...

simonihmig:
> The only thing I would like to see here is to also remove the no-check directive [here](https://github.com/ember-bootstrap/ember-bootstrap/blob/7329f018566fadd7b54e31b32721772e82c228c8/ember-bootstrap/addon/components/bs-collapse.hbs#L1)

@jelhan have you seen that comment? That's the thing that I was hoping we could change as part of this PR.

simonihmig (approved):

PR #2096, opened 2024-02-12
Add initial typescript configuration to test-app project

SanderKnauff:
This configration should allow us to use Typescript in the tests as well.
This will unblock the PR's that are currently awaiting Typescript types.


test-app/types/global.d.ts
```diff
@@ -0,0 +1,3 @@
+import 'ember-source/types';
+import 'ember-source/types/stable';
+import '@glint/environment-ember-loose';
```
jelhan, line 3: It should pull in the template registry from `ember-bootstrap`, shouldn't it?

jelhan (approved):
Looks good to me. Just one question.

Thanks a lot for picking it up!

test-app/types/global.d.ts
simonihmig: It's #2093 which sets up the registry, so assuming we will merge this first, we have to wait a bit. Or cherry-pick my changes? But honestly, I wouldn't overcomplicate this, we will need many more PRs to get everything properly sorted out...

SanderKnauff: It should, but I think that since Simon implemented it in #2093 already, it should be save to omit here.
An empty registry is possible, but would just look a bit odd.

simonihmig (approved):

PR #2099, opened 2024-02-14
Convert BsLinkTo to TypeScript

jelhan:
Converts `<BsLinkTo>` to TypeScript.

It seems that we are not having any tests explicitly for `<BsLinkTo>`. It is only implicitly tested due to usage in other components. Therefore this is not touching tests.

### DIFF
--- a/ember-bootstrap/addon/components/bs-link-to.hbs
+++ b/ember-bootstrap/addon/components/bs-link-to.hbs
@@ -3,6 +3,7 @@
   @models={{this.models}}
   @query={{this.query}}
   @disabled={{@disabled}}
+  class={{@attrClassInternal}}
   ...attributes
 >
   {{yield}}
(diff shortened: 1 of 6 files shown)


ember-bootstrap/addon/components/bs-link-to.hbs
```diff
@@ -1,10 +1,8 @@
-{{! @glint-nocheck }}
 <LinkTo
   @route={{@route}}
   @models={{this.models}}
   @query={{this.query}}
   @disabled={{@disabled}}
-  class={{@class}}
```
jelhan, line 7: `<BsLinkTo>` is a private component. We didn't used that argument anywhere. So I think we can just drop it. Likely a left-over from a very old version.

jelhan: Please ignore. Tests are telling me that I haven't searched correctly. Will fix.

simonihmig (approved):

PR #2103, opened 2024-02-16
exclude dev files from published package

jelhan:
I noticed that unnecessary files were included in the generated NPM package. It seems we lost the `.npmignore` file in #2073.

I added the [`.npmignore` file generated by Ember CLI for addons](https://github.com/ember-cli/ember-addon-output/blob/v5.6.0/.npmignore) as a starting point. I than added additional files after reviewing the generated package. It may contain some files not needed in our case. But I felt that does not hurt much.


SanderKnauff:
I think this file has accidentally been moved to the test-app directory. Could you remove it from there as well?

jelhan:
> I think this file has accidentally been moved to the test-app directory. Could you remove it from there as well?

Good catch! I have done that.

We now also have a nice diff between what we had pre-monorepo conversion and what we have now.

jelhan:
One thing I noticed after merging and I'm not sure about established best practices:

Should the `CHANGELOG.md` be included in the published package or not?

SanderKnauff:
> One thing I noticed after merging and I'm not sure about established best practices:
> 
> Should the `CHANGELOG.md` be included in the published package or not?

I think it is nice to have when inspecting the local files in node_modules, but I can also understand excluding it

PR #2108, opened 2024-02-28
Convert BsSpinner to TypeScript

Baltazore:
Converts <BsSpinner> to TypeScript. Including the tests.


Baltazore:
@jelhan thanks for review, updated.

jelhan (approved):
Looks good to me. 👍

Will give Simon and Sander the chance to review as well before hitting the merge button. In case there isn't an update in next 2-3 days, please ping me. I may forget about it without notification.

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
SanderKnauff, line 1: Huh, I wonder why this was updated since the rest of the lockfile seems the same...

SanderKnauff (approved):
Looks good to me! I have just a minor uncertainty about the lockfile change, but that might just have been pnpm updating that.

Since CI is fine with it, it probably is not that big of an issue.

pnpm-lock.yaml
jelhan: Good catch. I ignored that.

PNPM updates the lockfile version when running `pnpm install`. But `6.1` is a value, which we don't want. That's from an old PNPM version and was rolled back. See https://github.com/orgs/pnpm/discussions/6857.

Let's reset that change:

```suggestion
lockfileVersion: '6.0'
```

@Baltazore: Looks as if you need to upgrade PNPM on your machine.

Baltazore: yeah I had pnpm 8.6 locked due to ember-data, sorry ...

PR #2109, opened 2024-02-28
[WIP] Convert BsDropdown to TypeScript

jelhan:
Converts `<BsDropdown>` and its sub-components to TypeScript.

PR #2110, opened 2024-02-28
Dropdown should not add .btn-group class

jelhan:
`<BsDropdown>` used the `.btn-group` in some cases on the wrapping `<div>` element. But this should be fully controlled by the consumer.

Per Bootstrap docs:

> Wrap the dropdown’s toggle (your button or link) and the dropdown menu within `.dropdown`, or another element that declares `position: relative;`.
>
> https://getbootstrap.com/docs/5.3/components/dropdowns/#examples and https://getbootstrap.com/docs/4.6/components/dropdowns/#examples

Having the `.dropdown` class always fulfills the requirement, while consumer can add `.btn-group` or similar classes _additionally_.

The `.btn-group` class was added if

1. a `<button>` element was used as toggle and
2. the `<button>` was _not_ rendered in block mode.

This seems to be a leftover from BS3 support. For BS3 a `.btn-group` was required in this case:

> Use any button to trigger a dropdown menu by placing it within a `.btn-group` and providing the proper menu markup.
>
> https://getbootstrap.com/docs/3.4/components/#btn-dropdowns

I run into it when refactoring how `<BsDropdown::Button>` extends `<BsButton>` as preparation work for #2109.

This change _may_ break some apps, which rely on Ember Bootstrap adding the `.btn-group` class in that case. Not sure how likely it is that an app relies on 1) the difference between `.dropdown` and `.btn-group` _and_ 2) Ember Bootstrap adding the `.btn-group` class implicitly. I feel we should consider it a bug fix as Ember Bootstrap does not render the _expected_ markup for BS4 and BS5.

### DIFF
--- a/test-app/tests/integration/components/bs-dropdown-test.js
+++ b/test-app/tests/integration/components/bs-dropdown-test.js
@@ -50,35 +50,14 @@ module('Integration | Component | bs-dropdown', function (hooks) {
     assert.dom(dropleftClass).exists('has dropleft class');
   });
 
-  test('dropdown container with dropdown button has btn-group class', async function (assert) {
+  test('supports setting btn-group class', async function (assert) {
     await render(
-      hbs`<BsDropdown as |dd|><dd.button>Dropdown
+      hbs`<BsDropdown class='btn-group' as |dd|><dd.button>Dropdown
     <span class='caret'></span></dd.button><dd.menu><li><a
         href='#'
       >Something</a></li></dd.menu></BsDropdown>`,
     );
     assert.dom('.btn-group').exists('has btn-group class');
-  });
-
-  test('dropdown container with block dropdown button has dropdown class', async function (assert) {
-    await render(
-      hbs`<BsDropdown as |dd|><dd.button @block={{true}}>Dropdown
-    <span class='caret'></span></dd.button><dd.menu><li><a
-        href='#'
-      >Something</a></li></dd.menu></BsDropdown>`,
-    );
-    assert.dom('.dropdown').exists('has dropdown class');
-  });
-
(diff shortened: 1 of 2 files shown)


jelhan:
@SanderKnauff Do you have time reviewing this one as you recently working on the dropdown?

test-app/tests/integration/components/bs-dropdown-test.js
```diff
@@ -50,35 +50,14 @@ module('Integration | Component | bs-dropdown', function (hooks) {
...
     await render(
-      hbs`<BsDropdown as |dd|><dd.button>Dropdown
-    <span class='caret'></span></dd.button><dd.menu><li><a
-        href='#'
-      >Something</a></li></dd.menu></BsDropdown>`,
-    );
-    assert.dom('.btn-group').exists('has btn-group class');
-  });
-
-  test('dropdown container with block dropdown button has dropdown class', async function (assert) {
```
SanderKnauff, line 63: Is there a reason for removing the dropdown and dropup tests? According to the jsdoc of `bs-dropdown containerClass`, they should still be supported

SanderKnauff (changes requested):
I'm only unsure of the removal of the tests, but besides that it looks fine!

test-app/tests/integration/components/bs-dropdown-test.js
jelhan: The tests were asserting that `.btn-group` CSS class is used in some cases instead of `.dropdown`. That logic is removed in the PR. It is now using `.dropdown` CSS class only. `.btn-group` is not used in any case anymore.

There is another test (`dropdown container has dropdown class`), which covers that the markup is as expected.

PR #2111, opened 2024-02-28
Extend `<BsButton>` component by invoking it in template rather than extending JS class

jelhan:
Extending the JavaScript class of `<BsButton>` instead of invoking it in the template has several disadvantages:

1. Only the JavaScript class of the component is extended but the template is duplicated.
2. It's makes it much harder to understand `<BsDropdown::Button>` as it pulls in all complexity from `<BsButton>`'.
3. `<BsDropdown::Button>` has features not needed at all by the dropdown toggle. E.g. the yielded properties for the state of triggered async `@onClick` handler are useless as the `@onClick` handler for the dropdown toggle is not async.

Last but not least: I'm not sure if we could type it correctly. At least I haven't found a way how to modify the component signature from the extended class. For that reason I consider this refactoring a prerequisite for #2109.

It includes #2110, which we should align on first. Keeping it as WIP until than.


jelhan:
> It includes https://github.com/ember-bootstrap/ember-bootstrap/pull/2110, which we should align on first. Keeping it as WIP until than.

Missed that #2110 is not merged yet. I'm sorry for the noise.

jelhan:
Now it is actually ready for review. :smile:

ember-bootstrap/addon/components/bs-dropdown/button.hbs
```diff
@@ -1,21 +1,15 @@
...
-  {{#if this.icon}}<i class={{this.icon}}></i> {{/if}}{{this.text}}{{yield
-    (hash
-      isFulfilled=this.isFulfilled
-      isPending=this.isPending
-      isRejected=this.isRejected
-      isSettled=this.isSettled
-    )
-  }}
-</button>
+  {{yield}}
```
SanderKnauff, line 14: This seems like a change in API. The settledness states do not get yield anymore.
Is it possible to add `as |buttonState|` to `<BsButton>` so it can be passed to the `{{yield}}`? (`{{yield buttonState}}`)

SanderKnauff (changes requested):
There is a small change in API that might need to be looked at. It does look like quite a simple fix.
Apart from that, the changes looks fine!

ember-bootstrap/addon/components/bs-dropdown/button.hbs
jelhan: Only supported usage of `<BsDropdown::Button>` is as the `button` component yielded by `<BsDropdown>`. That component is the toggle for the dropdown. The `@onClick` action can never be async.

The `@onClick` action is hardcoded to be the `toggleDropdown` method of `<BsDropdown>`: https://github.com/ember-bootstrap/ember-bootstrap/blob/9c964025afc85a9d389b70b00a28aa1cfd78ff9a/ember-bootstrap/addon/components/bs-dropdown.hbs#L12-L14

The `toggleDropdown` method does not return a promise: https://github.com/ember-bootstrap/ember-bootstrap/blob/9c964025afc85a9d389b70b00a28aa1cfd78ff9a/ember-bootstrap/addon/components/bs-dropdown.js#L292-L299

It seems that yielding promise state here was done by mistake. Likely an issue due to copying over `<BsButton>`'s template to this template for customization.

SanderKnauff: Ah, I had not realized that the called was hardcoded. In that case it is fine!

SanderKnauff (approved):

SanderKnauff:
Is it possible to get a release out after this is merged? I currently have a PR at work that uses a workaround for one of the bugs that were fixed...

jelhan:
> Is it possible to get a release out after this is merged? I currently have a PR at work that uses a workaround for one of the bugs that were fixed. 😅

I just released [v6.3.1](https://github.com/ember-bootstrap/ember-bootstrap/releases/tag/v6.3.1) including the recent bug fixes.

SanderKnauff:
> > Is it possible to get a release out after this is merged? I currently have a PR at work that uses a workaround for one of the bugs that were fixed. 😅
> 
> I just released [v6.3.1](https://github.com/ember-bootstrap/ember-bootstrap/releases/tag/v6.3.1) including the recent bug fixes.

Thank you! 🎉

PR #2112, opened 2024-02-29
Convert Tooltip/Popover to TypeScript

simonihmig:


jelhan:
`<BsTooltip>` was converted to TypeScript in #2166, including the `BsContextualHelp` base class and the `isFastBoot` util. The `{{bs-default}}` helper has been converted in #2164. But the DOM utils (`ember-bootstrap/addon/utils/dom.ts`) have not been converted to TypeScript yet. We may want to pull that part out of this PR before closing.

PR #2113, opened 2024-02-29
Drop unused argument @ariaRole from <BsDropdown::Menu>

jelhan:
The `@ariaRole` argument was required for BS3. It is not used anymore since https://github.com/ember-bootstrap/ember-bootstrap/commit/8bc57449335dd39c759e9b14372555382401f5c5#diff-5f64d3b3af0f75bb6cb845fecfc7dff2583cdbc9b8fc9a54abb456ff4e498a41L39. 

I noticed it when typing the component. It seems that the TypeScript already pays off by uncovering old leftovers. :smile: 

I feel #2109 will get huge. Therefore trying to pull out as many small changes as possible and landing them before.

As the argument did not had any effect, I think we can consider it an internal change.

### DIFF
--- a/ember-bootstrap/addon/components/bs-dropdown/menu.js
+++ b/ember-bootstrap/addon/components/bs-dropdown/menu.js
@@ -24,16 +24,6 @@ export default class DropdownMenu extends Component {
   @ref('menuElement') menuElement = null;
 
   /**
-   * @property ariaRole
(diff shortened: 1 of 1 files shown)


SanderKnauff (approved):
Nice catch! Since the template doesn't use this, I feel like this is safe to merge!

PR #2114, opened 2024-02-29
Refactor tests for dropdown menu

jelhan:
Dropdown is aligned based on CSS class if used within a `<BsNav>` or `<BsNavbar>`. Bootstrap 5 expects the `.dropdown-menu-end` CSS class for the alignment. But Ember Bootstrap rendered the `.dropdown-menu-right` CSS class for Bootstrap 5 as well. That one is only correct for Bootstrap 4.

Additionally this fixes a bug rendering `.dropdown-menu-false` CSS class if `@align` argument was not set. That was introduced in https://github.com/ember-bootstrap/ember-bootstrap/commit/94ed2643071a6ada6500d2d13b13d28c3dd7f782#diff-6edad28cc2b90ce452b543d7e652aea3319372185fdc0b84ebad39a812820252L46-R48 but didn't had any visual impact.


ember-bootstrap/addon/components/bs-dropdown/menu.js
```diff
@@ -99,7 +100,13 @@ export default class DropdownMenu extends Component {
   }
 
   get alignClass() {
-    return this.align !== 'left' ? `dropdown-menu-${this.align}` : undefined;
+    if (this.align === 'right') {
```
jelhan, line 103: We only need to render a CSS class if `@align` argument is `"right"`.

Bootstrap supports responsive alignment as well: https://getbootstrap.com/docs/5.3/components/dropdowns/#responsive-alignment But that was also not supported by Ember Bootstrap before. It would require multiple changes:

1. Allow `@align` to take multiple values, e.g. `["left", "lg-right"]`.
2. Add `data-bs-display` (BS5) or `data-display` attribute with value `"static"` to the `.dropdown-menu` element if responsive alignment is used.

Changing the assertion from `this.align !== 'left'` to `this.align === 'right'` does not drop any supported use case. It only ensures that we don't render an invalid CSS class for unsupported values `@align`.

test-app/tests/integration/components/bs-dropdown/menu-test.js
```diff
@@ -76,23 +73,33 @@ module('Integration | Component | bs-dropdown/menu', function (hooks) {
...
+      hbs`<BsDropdown as |dd|><dd.menu
     @isOpen={{true}}
     @toggleElement={{this.element}}
     class='custom-class-1'
     data-test-menu
   >Something</dd.menu></BsDropdown>`,
-      );
+    );
+
+    assert.dom('.dropdown-menu').exists('menu has dropdown-menu class');
```
SanderKnauff, line 86: These assertions can be chained like the ones on L20-27

ember-bootstrap/addon/components/bs-dropdown/menu.js
SanderKnauff: Should we add support for this? If so could you create an issue?

SanderKnauff (changes requested):
Looks good! Just a minor nit comment, so if you fix that it's ready to be merged.

ember-bootstrap/addon/components/bs-dropdown/menu.js
jelhan: I created https://github.com/ember-bootstrap/ember-bootstrap/issues/2121 to track it.

jelhan:
@SanderKnauff: I think this was addressing the same problem as #2120. I'm sorry that I caused you doing the same work just by not following up on this one. Honestly speaking I only noticed when resolving merge conflicts that it was about the same issue.

This one is no limited to refactoring the tests. Would be great if you could review anyways. I think it is still a small improvement.

SanderKnauff:
> @SanderKnauff: I think this was addressing the same problem as #2120. I'm sorry that I caused you doing the same work just by not following up on this one. Honestly speaking I only noticed when resolving merge conflicts that it was about the same issue.
> 
> This one is no limited to refactoring the tests. Would be great if you could review anyways. I think it is still a small improvement.

No worries! I also completely forgot about this one. The changes look good, so this can be merged!

SanderKnauff (approved):

PR #2115, opened 2024-03-12
Convert BsModal related classes to TS

SanderKnauff:
WIP PR to track work of the conversion of BsModal and it's related components.

This also touches the `bs-default` and `bs-noop` helpers.


ember-bootstrap/addon/components/bs-modal/header.hbs
```diff
@@ -1,20 +1,15 @@
...
-        (hash
-          title=Title
-          close=Close
-        )
+      {{yield (hash title=Title close=Close)
       }}
     {{else}}
       {{#if (has-block)}}
-        {{yield}}
+        {{yield (hash title=Title close=Close)}}
```
jelhan, line 12: I don't think we need to yield here. We are checking before of the block should receive parameters.

If Glimmer does not support that well, I think we should just remove the `has-block-params` code path. Simplifying the implementation may be good anyways. At least I don't see benefits in conditionally yuelding the component. I doubt that Embroider is able to detect it and tree-shake the child component if it's not needed.

ember-bootstrap/addon/helpers/bs-default.ts
```diff
@@ -0,0 +1,5 @@
+export default function bsDefaultHelper<PRIMARY, SECONDARY>(
```
jelhan, line 1: Please have a look at https://github.com/ember-bootstrap/ember-bootstrap/blob/2f0049c23d56a0b671f46981fb808ce7bcbe1797/ember-bootstrap/addon/helpers/bs-default.ts for a more advanced implementation, which does a null check as well.

ember-bootstrap/addon/components/bs-modal/footer.hbs
```diff
@@ -1,15 +1,14 @@
-{{! @glint-nocheck }}
 {{#let (ensure-safe-component (bs-default @buttonComponent (component "bs-button"))) as |Button|}}
   <div class="modal-footer" ...attributes {{on "submit" (bs-default @onSubmit (bs-noop))}}>
     {{#if (has-block)}}
       {{yield}}
     {{else}}
       {{#if @submitTitle}}
         <Button @onClick={{@onClose}}>{{bs-default @closeTitle "Ok"}}</Button>
-        <Button @type={{bs-default @submitButtonType "primary"}} onClick={{@onSubmit}} disabled={{bs-default @submitDisabled false}}>{{@submitTitle}}</Button>
+        <Button @type={{bs-default @submitButtonType "primary"}} @onClick={{@onSubmit}} disabled={{bs-default @submitDisabled false}}>{{@submitTitle}}</Button>
```
jelhan, line 8: Good catch!

Is this causing a bug for consumers? If so, we may want to fix it in a separate PR and add missing test coverage.

jelhan (changes requested):
Just a few things I noticed skimming through it.

ember-bootstrap/addon/components/bs-modal/footer.hbs
SanderKnauff: I've addressed this in #2123. The missing @ does not seem to be breaking anything, but it is nicer to have it there just to be sure.

SanderKnauff:
I've returned this PR to draft status as I found out I forgot some classes that need to be converted as well.
Probably why it was a draft PR in the first place. 😅

SanderKnauff:
@jelhan It's been a while, but I found some time to continue the work on the TS conversion.

All classes for bs-modal and related have now been changed (if I did not miss any) and this PR is ready for review.

ember-bootstrap/addon/components/bs-modal/dialog.ts
```diff
@@ -52,9 +84,9 @@ export default class ModalDialog extends Component {
    * @private
    */
   @action
-  getOrSetTitleId(modalNode) {
-    //Title element may be set by user so we have to try and find it to set the id
-    let nodeId = null;
+  getOrSetTitleId(modalNode: Element) {
+    //Title element may be set by user, so we have to try and find it to set the id
```
jelhan, line 88: While we fix the typo, let's also add the missing white space at the beginning.

```suggestion
    // Title element may be set by user, so we have to try and find it to set the id
```

ember-bootstrap/addon/template-registry.ts
```diff
@@ -1,9 +1,36 @@
...
+}
+
+export default interface EmberBootstrapRegistry extends BsModalRegistry {
+  'bs-button': typeof BsButton;
   BsButton: typeof BsButton;
+  'bs-collapse': typeof BsCollapse;
   BsCollapse: typeof BsCollapse;
+  'bs-spinner': typeof BsSpinner;
   BsSpinner: typeof BsSpinner;
 }
```
jelhan, line 36: Why do we add all the dasherized versions? Why do we add all the private components yielded such as `BsModall::Header`? I expected that it only adds `BsModal: typeof BsModal;` to the existing list.

This review comment from @simonihmig seems to be touching the same topic in another PR: https://github.com/ember-bootstrap/ember-bootstrap/pull/2164/files#r1868407700

ember-bootstrap/addon/utils/is-fastboot.ts
```diff
@@ -0,0 +1,11 @@
+import { getOwner } from '@ember/application';
+import type FastBootService from 'ember-cli-fastboot/services/fastboot';
+
+export default function isFastBoot(context: object) {
+  const owner = getOwner(context);
+  if (!owner) {
+    return false;
+  }
+  const fastbootService = owner.lookup('service:fastboot') as FastBootService;
```
jelhan, line 9: I think the type is not fully correct here. `Owner.lookup` returns `undefined` or `null` (cannot recall anymore) if no factory has been registered with that name. Sadly the return type in that case doesn't seem to be documented well by Ember.

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -26,8 +29,8 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
     assert
       .dom('.modal-footer button')
       .hasClass('btn-primary', 'Button is a primary button.');
-    assert.ok(
-      this.element.querySelector('.modal-footer button').getAttribute('type'),
+    assert.strictEqual(
+      this.element.querySelector('.modal-footer button')?.getAttribute('type'),
       'button',
       'Submit button is of type submit.',
     );
```
jelhan, line 36: TypeScript shows it value. :smile:

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -37,15 +40,18 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
     assert.strictEqual(close.callCount, 1, '@onClose is called exactly once');
   });
 
+  // TODO The logs and expected values in the assertions for the button type do not agree on the expected button type.
```
jelhan, line 43: Is this something we still need to do? Or has it been fixed?

test-app/tests/integration/components/bs-modal/header-test.gts
```diff
@@ -88,24 +97,34 @@ module('Integration | Component | bs-modal/header', function (hooks) {
...
+    await render(
+      <template>
+        {{! template-lint-disable no-unused-block-params }}
+        <BsModalHeader as |header|>
+          <div id='custom'>Test</div>
+        </BsModalHeader>
+        {{! template-lint-enable no-unused-block-params }}
+      </template>,
+    );
+    /* eslint-enable @typescript-eslint/no-unused-vars */
```
jelhan, line 121: Why can we just not use the yielded value?

```suggestion
    await render(
      <template>
        <BsModalHeader>
          <div id='custom'>Test</div>
        </BsModalHeader>
      </template>,
    );
```

ember-bootstrap/addon/components/bs-modal.ts
```diff
@@ -3,20 +3,71 @@ import { assert } from '@ember/debug';
...
 
 function afterRender() {
-  return new Promise((resolve) => schedule('afterRender', resolve));
+  return new Promise((resolve) => schedule('afterRender', resolve, undefined));
+}
+
+export type ModalPosition = 'top' | 'center' | null;
+
+interface Signature {
+  Args: {
```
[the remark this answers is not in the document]
SanderKnauff: Ah I missed those. I'm overlooked the @args annotated fields. Personally I'm not too fond of them as they don't seem hide the typescript arguments.

That also brings up another point, the API docs could probably use a rework in how they are generated. It now picks up all the JSDoc comments with redundant information floating around in the components and creates the docs from that. Wouldn't they be more accurate if they were generated from the TS definitions?
If we were to do that, the docs would also be more visible when inspecting the signatures in an IDE.

This of course can only be done once all components have been updated to Typescript. I'm also not sure if and how YuiDoc would support something like this.

ember-bootstrap/addon/components/bs-modal/header.hbs
```diff
@@ -14,7 +13,12 @@ as |Title Close|
       }}
     {{else}}
       {{#if (has-block)}}
-        {{yield}}
+        {{yield
+          (hash
+            title=Title
+            close=Close
+          )
+        }}
```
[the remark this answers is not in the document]
SanderKnauff: These yields are added to align with the Block section of the bs-header signature. The codepath above this one yields both title and close as well.
The alternative would be to make the yield output be potentially undefined, but that would mean that any consumer must perform an existence check.

I cannot quickly thing of anything that might break by adding yield parameters. They are optional for the consumers to use so I think it's fine.

ember-bootstrap/addon/template-registry.ts
SanderKnauff: You're right. The dasherized versions should not be part of the public API. These need to be moved to the internal types until we can get rid of the component helper usages.

ember-bootstrap/addon/utils/is-fastboot.ts
SanderKnauff: The null check after that makes more sense like that. I'll update it.

test-app/tests/integration/components/bs-modal/footer-test.gts
SanderKnauff: This one is solved, I just forgot to remove the todo

test-app/tests/integration/components/bs-modal/header-test.gts
SanderKnauff: I think this is testing the has-block-params codepath, so this might have been intentional in the original test. Converting it to GTS just makes the linter more strict

ember-bootstrap/addon/components/bs-modal.ts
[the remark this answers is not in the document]
jelhan: Yes. Fully agree. We should write our API docs on top of TypeScript types as a follow-up step.

test-app/tests/integration/components/bs-modal/header-test.gts
jelhan: If this is about testing the block params we should also use them. Otherwise compiler may refactor them away as they are not used and we don't notice it.

SanderKnauff: That makes sense. I've updated the test to use the title out of the block params

SanderKnauff:
@jelhan can you take another look at this PR. I believe it's in a mergable state, though it might need a rebase after this time...

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -55,9 +60,9 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
     assert
       .dom('.modal-footer button:first-child')
       .hasClass(defaultButtonClass(), 'Close button is a default button.');
-    assert.ok(
+    assert.strictEqual(
```
jelhan, line 63: We should refactor here to `assert.dom('.modal-footer button:first-child').hasAttribute('type',   'button', 'Submit button is of type submit.')` as well for readability at some point. But not necessarily in this PR.

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -68,12 +73,12 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
     assert
       .dom('.modal-footer button:last-child')
       .hasClass('btn-primary', 'Submit button is a primary button.');
-    assert.ok(
+    assert.strictEqual(
```
jelhan, line 76: Same as in the two other cases.

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -134,31 +145,43 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
...
+          @submitTitle='submit'
+          @submitDisabled={{state.disabled}}
+        />
+      </template>,
     );
 
     assert
       .dom('.modal-footer button')
       .exists({ count: 2 }, 'Modal footer has two button.');
     assert.notOk(
```
jelhan, line 165: This is a great use case for the `isDisabled` DOM assertion. But could be also done in a follow-up.

jelhan (changes requested):
I'm sorry for not following up on this earlier. I actually had a pending review. But never finished it. Thanks for the reminder.

ember-bootstrap/addon/components/bs-modal.ts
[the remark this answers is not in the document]
SanderKnauff: Yes, from my testing, neither `glint` or `tsc` cares about the existence of the actual `arg` on the component's signature.
This may be a bug from how the arg decorator is typed, but I'm not sure if it can be fixed.

I understand its use, but to me it hides some pretty common EmberJS patterns.

```ts
@arg
foo = 'default text';
```
Could be just as well be written as
```ts
// If should be a reassignable default
@tracked
foo = this.args.foo ?? 'default text'

// or if immutable and should react to changes
get foo() {
  return this.args.foo ?? 'default text';'
}
```

ember-bootstrap/addon/modifiers/bs-conditional-attribute.ts
```diff
@@ -2,18 +2,22 @@ import { modifier } from 'ember-modifier';
 
 interface Signature {
   Args: {
-    Positional: [attribute: string, condition: boolean, value: string];
+    Positional: [
+      attribute: string,
+      condition: boolean | undefined,
+      value: unknown,
```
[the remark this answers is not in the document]
SanderKnauff: Yeah, that might be better. I've changed it.

ember-bootstrap/addon/utils/dom.d.ts
```diff
@@ -0,0 +1 @@
+export function getDestinationElement(owner: object): Element;
```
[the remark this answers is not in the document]
SanderKnauff: And `null` as well! I've added them to the return type.

ember-bootstrap/types/ember-focus-trap/modifiers/focus-trap.d.ts
```diff
@@ -0,0 +1,18 @@
+import { ModifierLike } from '@glint/template';
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    'focus-trap': ModifierLike<{
+      Args: {
+        Named: {
+          shouldSelfFocus?: boolean;
+          focusTrapOptions?: {
```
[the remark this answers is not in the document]
SanderKnauff: Ah, I was focussed (ha!) on the ember-focus-trap implementation. This is better.

test-app/tests/integration/components/bs-modal/footer-test.gts
```diff
@@ -26,8 +29,8 @@ module('Integration | Component | bs-modal/footer', function (hooks) {
     assert
       .dom('.modal-footer button')
       .hasClass('btn-primary', 'Button is a primary button.');
-    assert.ok(
-      this.element.querySelector('.modal-footer button').getAttribute('type'),
+    assert.strictEqual(
```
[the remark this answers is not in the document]
SanderKnauff: Changed it to make use of Qunit DOM

test-app/tests/integration/components/bs-modal/footer-test.gts
SanderKnauff: Changed it to make use of Qunit DOM

test-app/tests/integration/components/bs-modal/footer-test.gts
SanderKnauff: Changed it to make use of Qunit DOM

SanderKnauff:
Thanks for the feedback @jelhan. I've implemented your suggestions!

ember-bootstrap/addon/components/bs-modal.ts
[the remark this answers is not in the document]
jelhan: I agree that the `@arg` decorator causing more complexity and making it harder for new contributors to understand what's happening. If it's also causing TypeScript issue, I tend to refactor it out entirely.

ember-bootstrap/addon/components/bs-modal/footer.hbs
```diff
@@ -8,7 +8,7 @@
         <Button
           @type={{bs-default @submitButtonType "primary"}}
           @onClick={{@onSubmit}}
-          {{bs-conditional-attribute "disabled" @submitDisabled @submitDisabled}}
+          {{bs-conditional-attribute "disabled" @submitDisabled (if @submitDisabled "true" "false")}}
```
jelhan, line 11: This seems to be wrong. The value is only taken into consideration if the condition is true. The condition is `@submitDisabled`. The value can be an empty string for boolean attributes such as `disabled`.

SanderKnauff: Ah, you're right. I was too preoccupied with looking for a toString inside the template that I overlooked that it only uses the "true" value anyway

ember-bootstrap/addon/components/bs-modal.ts
```diff
@@ -639,17 +697,36 @@ export default class Modal extends Component {
...
-    let modalEl = this.modalElement;
+    const modalEl = this.modalElement;
+    if (!modalEl.parentNode) {
+      return 0;
+    }
     modalEl.parentNode.insertBefore(scrollDiv, modalEl.nextSibling);
-    let scrollbarWidth = scrollDiv.offsetWidth - scrollDiv.clientWidth;
-    scrollDiv.parentNode.removeChild(scrollDiv);
+    const scrollbarWidth = scrollDiv.offsetWidth - scrollDiv.clientWidth;
+    scrollDiv.parentNode?.removeChild(scrollDiv);
```
jelhan, line 708: Small leftover from previous refactoring.

```suggestion
    scrollDiv.parentNode.removeChild(scrollDiv);
```

ember-bootstrap/addon/components/bs-modal/footer.hbs
```diff
@@ -9,7 +8,7 @@
         <Button
           @type={{bs-default @submitButtonType "primary"}}
           @onClick={{@onSubmit}}
-          {{bs-conditional-attribute "disabled" @submitDisabled @submitDisabled}}
+          {{bs-conditional-attribute "disabled" @submitDisabled "true"}}
```
jelhan, line 11: MDN recommends using an empty string or the attribute's name for boolean attributes.

>  Boolean attributes are considered to be true if they're present on the element at all. You should set value to the empty string ("") or the attribute's name, with no leading or trailing whitespace.

I think we should follow that recommendation. Even though any string would work.

```suggestion
          {{bs-conditional-attribute "disabled" @submitDisabled "disabled"}}
```

test-app/.eslintrc.js
```diff
@@ -53,7 +53,9 @@ module.exports = {
         'plugin:@typescript-eslint/eslint-recommended',
         'plugin:@typescript-eslint/recommended',
       ],
-      rules: {},
+      rules: {
+        'no-undef': 'off',
```
jelhan, line 57: Why do we need to disable this rule?

jelhan (changes requested):

test-app/.eslintrc.js
SanderKnauff: There are some global DOM types that ESLint does not recognize. Specifically `(animation as CSSTransition)` in `bs-modal-test.gts`. The docs from typescript-eslint mention that it is recommended to disabled `no-undef` in Typescript files, since Typescript checks for it anyway.

https://typescript-eslint.io/troubleshooting/faqs/eslint/#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

ember-bootstrap/addon/components/bs-modal.ts
SanderKnauff: This one was actually not null-checked yet, so I've added the check for it as well.

ember-bootstrap/addon/components/bs-modal/footer.hbs
SanderKnauff: Ah, good point!

jelhan (approved):
Looks good to me. Thanks a lot!

PR #2120, opened 2024-04-10
Fix alignment of dropdown menus when using Bootstrap 5

SanderKnauff:
Bootstrap 5 uses start/end instead of left/right. The menu was still using the value from the align argument. In addition to this, the styles used in Bootstrap 5 check for the existence of the [data-bs-popper] attribute. This has been added to the menu as well.

I noticed this issue while converting the app at work from BS4 to BS5.

Before fix: 
image

After fix:
image

### DIFF
--- a/ember-bootstrap/addon/components/bs-dropdown/menu.js
+++ b/ember-bootstrap/addon/components/bs-dropdown/menu.js
@@ -90,18 +90,12 @@ export default class DropdownMenu extends Component {
   }
 
   get alignClass() {
-    if (macroCondition(getOwnConfig().isBS4)) {
-      return this.align !== 'left' ? `dropdown-menu-${this.align}` : undefined;
+    if (this.align === 'right') {
+      const alignClass = macroCondition(getOwnConfig().isBS4) ? 'right' : 'end';
+      return `dropdown-menu-${alignClass}`;
     }
 
-    switch (this.align) {
-      case 'left':
-        return 'dropdown-menu-start';
-      case 'right':
-        return 'dropdown-menu-end';
-      default:
-        return undefined;
-    }
+    return undefined;
   }
 
   @tracked
--- a/ember-bootstrap/addon/components/bs-dropdown/menu.js
+++ b/ember-bootstrap/addon/components/bs-dropdown/menu.js
@@ -4,6 +4,7 @@ import { next } from '@ember/runloop';
 import { getDestinationElement } from 'ember-bootstrap/utils/dom';
 import { ref } from 'ember-ref-bucket';
 import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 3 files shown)


jelhan (approved):
Good catch!

ember-bootstrap/addon/components/bs-dropdown/menu.js
```diff
@@ -89,7 +90,18 @@ export default class DropdownMenu extends Component {
...
 
   get alignClass() {
-    return this.align !== 'left' ? `dropdown-menu-${this.align}` : undefined;
+    if (macroCondition(getOwnConfig().isBS4)) {
+      return this.align !== 'left' ? `dropdown-menu-${this.align}` : undefined;
+    }
+
+    switch (this.align) {
+      case 'left':
+        return 'dropdown-menu-start';
```
jelhan, line 99: I double checked in Bootstrap docs for v5. If I get it right, `.dropdown-menu-start` is not needed as that's the default: https://getbootstrap.com/docs/5.3/components/dropdowns/#menu-alignment

test-app/tests/integration/components/bs-dropdown/menu-test.js
```diff
@@ -26,10 +31,28 @@ module('Integration | Component | bs-dropdown/menu', function (hooks) {
...
+  >Something</dd.menu></BsDropdown>`,
+    );
+
+    let alignmentClass = versionDependent(
+      'dropdown-menu-left',
+      'dropdown-menu-start',
+    );
+    assert
+      .dom('.dropdown-menu')
+      .hasClass(alignmentClass, `menu has ${alignmentClass} class`);
```
jelhan, line 53: I think we do t expect it to have a class in BS4 and 5.

- https://getbootstrap.com/docs/5.3/components/dropdowns/#menu-alignment
- https://getbootstrap.com/docs/4.6/components/dropdowns/#menu-alignment

I think we should assert that it does not have any class except `.dropdown-menu`. Not sure how well that's supported by qunit-dom.

jelhan (changes requested):
I was too quick with my first review. Only noticed failing CI afterwards. And that led me to some change requests.

ember-bootstrap/addon/components/bs-dropdown/menu.js
SanderKnauff: Ah, I had not noticed that. I will remove the left/start

test-app/tests/integration/components/bs-dropdown/menu-test.js
SanderKnauff: QUnit DOM may not, but I'll do the assertions on the classlist.

jelhan (approved):

PR #2122, opened 2024-04-10
modernize hbs import in tests

jelhan:
Importing `hbs` from `htmlbars-inline-precompile-template` is an outdated pattern. And it does not support TypeScript. This PR modernizes all imports at once to TypeScript conversion and lower the barrier of entrance.


SanderKnauff (approved):
Nice! This will certainly help with Typescript/Glint conversion!

PR #2123, opened 2024-04-10
Add @ to `onClick` in BsModal::Footer submit button

SanderKnauff:
As part of the Typescript conversion of `bs-modal`, I noticed that the `@` on `@onClick` was missing for the submit button.

I have added tests for the submit and close button, and they seem to be fine with or without the `@`, but to keep in line with Ember's usual style, we should put the `@` in there anyway.

### DIFF
--- a/ember-bootstrap/addon/components/bs-modal/footer.hbs
+++ b/ember-bootstrap/addon/components/bs-modal/footer.hbs
@@ -6,7 +6,13 @@
     {{else}}
       {{#if @submitTitle}}
         <Button @onClick={{@onClose}}>{{bs-default @closeTitle "Ok"}}</Button>
-        <Button @type={{bs-default @submitButtonType "primary"}} @onClick={{@onSubmit}} disabled={{bs-default @submitDisabled false}}>{{@submitTitle}}</Button>
+        <Button
+          @type={{bs-default @submitButtonType "primary"}}
+          @onClick={{@onSubmit}}
+          {{bs-conditional-attribute "disabled" @submitDisabled @submitDisabled}}
+        >
+          {{@submitTitle}}
+        </Button>
       {{else}}
         <Button @type="primary" @onClick={{@onClose}}>{{bs-default @closeTitle "Ok"}}</Button>
       {{/if}}
--- a/ember-bootstrap/addon/components/bs-modal/footer.hbs
+++ b/ember-bootstrap/addon/components/bs-modal/footer.hbs
@@ -6,7 +6,7 @@
     {{else}}
       {{#if @submitTitle}}
         <Button @onClick={{@onClose}}>{{bs-default @closeTitle "Ok"}}</Button>
-        <Button @type={{bs-default @submitButtonType "primary"}} onClick={{@onSubmit}} disabled={{bs-default @submitDisabled false}}>{{@submitTitle}}</Button>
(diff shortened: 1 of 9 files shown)


jelhan:
Great catch. My assumption would be that `onClick` seems to work as well for most use cases as Ember sets the `onclick` property of the `button` element.

If that assumptions is correct, the `handleClick` logic of `<BsButton>` would be skipped: https://github.com/ember-bootstrap/ember-bootstrap/blob/7b66df9df01f80baa6634b4bd0e1fc76299f1a13/ember-bootstrap/addon/components/bs-button.ts#L421-L457 This should have the affect that promise state is not handled correctly with the button.

The promise state is used in `<BsButton>` to control the `disabled` argument. If my assumption is correct, the button is _not_ disabled while a promise triggered by it is still running. That would be a bug, which is just not covered by our existing tests.

Would be great if you could double check that. Maybe even add a test case for it.

SanderKnauff:
To get the pending state to work as expected, I had to create a modifier for conditionally setting an attribute on an element. The `disabled` attribute on the Footer submit button was interfering with the default `disabled` behavior on BsButton.
In the changed situation, the `disabled` attribute will only exist on the footer if  `@submitDisabled` is set. In all other cases, `BsButton`, or the button component provided by the consumer, will handle the `disabled` state.

jelhan (changes requested):
Thanks a lot. Didn't had the time for a detailed review yet. But what I have seen so far looks great. Only one small change to dependency management.

SanderKnauff:
> Thanks a lot. Didn't had the time for a detailed review yet. But what I have seen so far looks great. Only one small change to dependency management.

That's fine! I don't see the comment about the dependency management though. Should something be changed in `ember-bootstrap/package.json`?

jelhan:
> > Thanks a lot. Didn't had the time for a detailed review yet. But what I have seen so far looks great. Only one small change to dependency management.
> 
> That's fine! I don't see the comment about the dependency management though. Should something be changed in `ember-bootstrap/package.json`?

Arghs. Somehow my review comment got lost. In short:

I think ember-modifier should be a peer dependency. It is added as a dependency for every app in Ember CLI blueprints.

Ember Bootstrap introduces a peer dependency on ember-modifier indirectly already today. Ember Bootstrap depends on ember-style-modifier. Ember-style-modifier has a peer dependency on ember-modifier.

Supported range for ember-modifier peer dependency should be the same as ember-style-modifier supports. In that case it wouldn't be a breaking change. At least in my understanding of our informal SemVer commitment for dependency management.

SanderKnauff:
> Ember Bootstrap introduces a peer dependency on ember-modifier indirectly already today. Ember Bootstrap depends on ember-style-modifier. Ember-style-modifier has a peer dependency on ember-modifier.
>
> Supported range for ember-modifier peer dependency should be the same as ember-style-modifier supports. In that case it wouldn't be a breaking change. At least in my understanding of our informal SemVer commitment for dependency management.

Ah, that makes sense. I've check `ember-style-modifier`, but that does not seem to specifically declare `ember-modifier` as peer dependency. I've copied version the range of `ember-modifier` that `ember-style-modifier` supports now

jelhan:
> I've check `ember-style-modifier`, but that does not seem to specifically declare `ember-modifier` as peer dependency.

Good catch! I assumed it does. But it does not. Same for ember-popper-modifier. Both declare it as dependencies only. I guess that should be changed upstream as well at some point.

jelhan (approved):
Looks good to me. Especially the improved test coverage. Great job!

PR #2127, opened 2024-05-13
Fix syntax in BsForm API docs

bracke:


ember-bootstrap/addon/components/bs-form.js
```diff
@@ -97,7 +97,7 @@ import { ref } from 'ember-ref-bucket';
   It could be used for visual feedback about last submission:
 
   ```hbs
-  <BsForm @onSubmit={{action 'save}} as |form|>
+  <BsForm @onSubmit={{action "save"}} as |form|>
```
jelhan, line 100: Actually it should not use `{{action}}` helper anymore. But that could be done in a later refactoring. Especially as it affects more places.

jelhan (approved):
Thanks a lot for fixing!

jelhan:
Merging regardless of failing tests as they are not related.

Issue #2131, opened 2024-05-28
Remove usage of action template helper and modifier


jelhan:
The `(action)` template helper and `{{action}}` modifier has been deprecated. See [RFC 1006](https://rfcs.emberjs.com/id/1006-deprecate-action-template-helper/) for details.

We are still relying on both at many places. In the addon itself, in the test app, and in the documentation app. We should refactor away from it especially in the addon code shipped to consumers as soon as possible.

These are the files where the action helper is still used:
- [x] `test-app/tests/integration/components/bs-accordion/item-test.js` #2150
- [x] `test-app/tests/integration/components/bs-alert-test.js` #2150
- [x] `test-app/tests/integration/components/bs-dropdown-test.js` #2146
- [x] `test-app/tests/integration/components/bs-dropdown/toggle-test.js` #2146
- [x] `test-app/tests/integration/components/bs-form-test.js` #2148
- [x] `test-app/tests/integration/components/bs-form/element-test.js` #2148
- [x] `test-app/tests/integration/components/bs-modal-simple-test.js` #2147
- [x] `test-app/tests/integration/components/bs-modal-test.js` #2147
- [x] `test-app/tests/integration/components/bs-nav/item-test.js` #2149
- [x] `test-app/tests/integration/components/bs-navbar-test.js` #2149
- [x] `test-app/tests/integration/components/bs-popover-test.js` #2150
- [x] `test-app/tests/integration/components/bs-tab-test.js` #2150
- [x] `test-app/tests/integration/components/bs-tooltip-test.js` #2150

The JSDoc comments also still use the action helper the following places:
- [x] `ember-bootstrap\addon\components\bs-tooltip.js` #2152
- [x] `ember-bootstrap\addon\components\bs-tab.js` #2152
- [x] `ember-bootstrap\addon\components\bs-popover.js` #2152
- [x] `ember-bootstrap\addon\components\bs-navbar.js` #2152
- [x] `ember-bootstrap\addon\components\bs-modal.js` #2152
- [x] `ember-bootstrap\addon\components\bs-form.js` #2152
- [x] `ember-bootstrap\addon\components\bs-dropdown.js` #2152
- [x] `ember-bootstrap\addon\components\bs-form\element.js` #2152
- [x] `ember-bootstrap\addon\components\bs-button-group.js` #2152
- [x] `ember-bootstrap\addon\components\bs-button.ts` #2163 

After that, the configuration should be updated to re-enable the warning:
[...]

PR #2136, opened 2024-07-20
Tests: Add support for writing tests in Gjs/Gts

SanderKnauff:
When changing the tests to get rid of the action helpers (#2131), it might be a good time to also update the tests to Gjs/Gts. The changes for the action helper will be influencing how we access scope variables in the tests and so will using ember-template-import syntax.

 This PR prepares the test-app for template-imports and converts one file to gjs and one to gts as a proof-of-concept and to verify that the configuration of tools such as ESLint and Prettier still work correctly.


test-app/.eslintrc.js
```diff
@@ -77,13 +97,5 @@ module.exports = {
       },
       extends: ['plugin:n/recommended'],
     },
-    {
-      // test files
-      files: ['tests/**/*-test.{js,ts}'],
-      extends: [
-        // @todo enable this while fixing linting errors in tests...
-        // 'plugin:qunit/recommended'
```
jelhan, line 85: We should keep track that we should enable `plugin:qunit/recommended` for the tests at some point of time. Do you mind moving those lines into the two new blocks for test files?

test-app/.template-lintrc.js
```diff
@@ -38,5 +22,11 @@ module.exports = {
         'require-presentational-children': 'warn',
       },
     },
+    {
+      files: ['tests/**/*.(gjs|gts)'],
+      rules: {
+        prettier: 'off',
+      },
+    },
```
jelhan, line 30: Why do we disable Prettier for `.gjs` and `.gts` files? Are we running Prettier for the `<template>` tags with ESLint now?

test-app/tests/integration/components/bs-accordion-test.gjs
```diff
@@ -80,12 +96,18 @@ module('Integration | Component | bs-accordion', function (hooks) {
...
   test('changing selected item expands this item', async function (assert) {
-    this.set('selected', 1);
-    await render(hbs`<BsAccordion @selected={{this.selected}} as |acc|>
-  <acc.item @value={{1}} @title='TITLE1'>CONTENT1</acc.item>
-  <acc.item @value={{2}} @title='TITLE2'>CONTENT2</acc.item>
-</BsAccordion>`);
-    this.set('selected', 2);
+    const data = new (class {
+      @tracked selected = 1;
+    })();
```
jelhan, line 101: Are there any established best practices within the community how to create data objects for tests? Personally I feel this style is hard to read. Explicitly declaring the class before may be easier:

```suggestion
    class TestData {
      @tracked selected = 1;
    };
    
    const data = new TestData();
```

But I'm not much opinionated about it.

jelhan (approved):
Thanks a lot for picking this one up. I have two questions and a minor change request. But nothing major. I'm also fine with merging as is.

PR #2143, opened 2024-08-03
Run tests for fastboot with ember-canary and Embroider

SanderKnauff:
The mentioned issue has been resolved and the tests work after removing the option

### DIFF
--- a/test-app/config/ember-try.js
+++ b/test-app/config/ember-try.js
@@ -60,10 +60,6 @@ module.exports = async function () {
         },
         env: {
           // FAIL_ON_DEPRECATION: true,
-
-          // TODO: Enable again when FastBoot is ready for Ember 5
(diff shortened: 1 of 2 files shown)


SanderKnauff:
It looks like the Fastboot tests also work for Embroider. Should we enable them across the board?

@simonihmig according to `git blame` you introduced the disable flag. Do you think it is okay to enable them all?

jelhan (approved):
Looks good to me.

jelhan:
> It looks like the Fastboot tests also work for Embroider. Should we enable them across the board?

If the tests pass, we should enable them in my opinion. I think an incompatibility between Embroider and Fastboot was the only reason those tests were skipped when introducing Embroider tests. At least that's how I read the PR description in #930.

PR #2150, opened 2024-08-09
Tests: Remove remaining action helpers from tests and convert to gjs

SanderKnauff:
These remaining ones did not have obvious groups, so I put them together to not create a large dump of PR's.
I can split them up if preferred.

### DIFF
--- a/test-app/tests/integration/components/bs-tab-test.gjs
+++ b/test-app/tests/integration/components/bs-tab-test.gjs
@@ -1,20 +1,17 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { click, render, settled } from '@ember/test-helpers';
-import { hbs } from 'ember-cli-htmlbars';
 import setupNoDeprecations from '../../helpers/setup-no-deprecations';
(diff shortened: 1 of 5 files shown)


test-app/tests/integration/components/bs-tab-test.gjs
```diff
@@ -274,73 +329,89 @@ module('Integration | Component | bs-tab', function (hooks) {
...
-    await render(hbs`<BsTab @fade={{false}} @activeId={{this.paneId}} as |tab|>
-  <tab.pane @id='pane1' @title='Tab 1'>
-    tabcontent 1
-  </tab.pane>
-  <tab.pane @id='pane2' @title='Tab 2'>
-    tabcontent 2
-  </tab.pane>
-</BsTab>`);
+    class State {
+      @tracked paneId = 'pane1';
```
jelhan, line 379: I _think_ this doesn't need to be tracked. We assert that it isn't changed. If it would be changed, the test would fail. We don't care if UI would be in sync with the state. Or do I miss something?

jelhan (approved):
Looks good to me.

I noticed one case in which tracked input state may not be needed. But I'm not fully sure myself to be honest. Please feel free to ignore that comment.

PR #2151, opened 2024-08-17
Tests: Re-enable template-lint rules and fix remaining issues

SanderKnauff:


test-app/tests/integration/components/bs-form-test.gjs
```diff
@@ -1314,10 +1314,10 @@ module('Integration | Component | bs-form', function (hooks) {
     );
 
     await triggerEvent('form', 'submit');
-    assert.dom('form').hasText('isSubmitted');
+    assert.dom('form').hasText('Input: isSubmitted');
```
jelhan, line 1317: The `Input:` prefix makes it hard to read the test. What do you think about wrapping the assertion string in an element to have a better test selector?

jelhan (approved):
Looks good to me. I only have a minor comment to improve readability.

test-app/tests/integration/components/bs-form-test.gjs
SanderKnauff: That could work. Alternatively, we could just add ignore comments for this linting rule, as I do not see much benefit for most of these cases.

test-app/tests/helpers/setup-no-deprecations.ts
```diff
@@ -7,16 +7,7 @@ const deprecations = new Set<string>();
...
-  // @todo remove when https://github.com/alexlafroscia/ember-popper-modifier/issues/452 is fixed
-  /ember-modifier/,
-  // @todo investigate what is still triggering the deprecation, might be some internals like
-  // ember-cli-app-version or ember-export-application-global
-  /@ember\/string/,
-  // @todo https://github.com/ember-bootstrap/ember-bootstrap/issues/2131
-  /Usage of the `\(action\)` helper is deprecated/,
-  /Usage of the `{{action}}` modifier is deprecated/,
-];
+const ignoredDeprecations: string[] = [];
```
simonihmig, line 10: 😍

simonihmig (approved):

PR #2152, opened 2024-08-17
Replace {{action}} helper in doc comments with modern patterns

SanderKnauff:


simonihmig (approved):
Uh gosh, so much old cruft still around! 🙈

Thank you! 😅

PR #2156, opened 2024-10-16
Widen support for ember-render-helpers to v1

jelhan:
Support for ember-render-helpers@^1.0.0 has been added in #2141 already. But accidentally support for ember-render-helpers@^0.2.0 has been dropped.  We don't want that as it would make upgrading more complex for consumers. And may be considered a breaking change.

The change has not been released yet. So it is safe to just fix it before releasing.

Adding a changelog entry because there isn't one for #2141 yet.

I decided adding an additional ember-try scenario for ember-render-helpers@^0.2.0 because some code were changed on the upgrade as well.


SanderKnauff (approved):
Looks good! I like the extra ember-try scenario to verify the compatibility

PR #2157, opened 2024-10-16
Fix compatibility with ember-render-helpers v0.2

jelhan:
#2156 should have restored compatibility with ember-render-helpers@^0.2.0 but it hasn't. The test had a bug and therefore leading to a false sense of security. The helpers were renamed in the v1.0 release _and_ the new names weren't available in v0.2.0.

This PR requires [ember-render-helpers v0.2.1](https://github.com/buschtoens/ember-render-helpers/releases/tag/v0.2.1) which backports the new helper names.

It removes the test as it was not working as expected anyways. See https://github.com/ember-bootstrap/ember-bootstrap/pull/2157#issuecomment-2417696426 and https://github.com/ember-cli/ember-try/issues/970 for context.

### DIFF
--- a/ember-bootstrap/package.json
+++ b/ember-bootstrap/package.json
@@ -57,7 +57,7 @@
     "ember-on-helper": "^0.1.0",
     "ember-popper-modifier": "^2.0.0 || ^3.0.0 || ^4.0.0",
     "ember-ref-bucket": "^4.0.0 || ^5.0.0",
-    "ember-render-helpers": "^0.2.0 || ^1.0.0",
+    "ember-render-helpers": "^0.2.1 || ^1.0.0",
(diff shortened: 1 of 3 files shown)


jelhan:
I'm not sure if we can get that compatibility test reliable. Overrides (and resolutions) are only supported at the root of the project. But as far as I know Ember Try does not support modifying the `package.json` at the root of the project in a monorepo.

With the current setup it is coincidence if the ember-render-helpers version from the test-app or from ember-bootstrap package gets pulled into the build.

SanderKnauff (approved):
Looks good! Should these helpers be refactored away in the future?

jelhan:
> Looks good! Should these helpers be refactored away in the future?

I think so. Conceptually they are the same as @ember/render-modifiers.

I haven't looked into why we are using them to be honest.

Issue #2158, opened 2024-10-23
Modal Focus-Trap Deactivates on "Escape" or "Click" Even when relevant options Are Disabled


kdagnan:
---

Output from `ember bootstrap:info`:

```
ember-bootstrap: ^5.1.1 -> 5.1.1
ember-cli: ~4.11.0 -> 4.11.0
bootstrap: ^5.3.2 -> 5.3.2
bootstrap-sass: n/a
ember-cli-sass: ^11.0.1 -> 11.0.1
ember-cli-less: n/a
Bower packages:
bootstrap: n/a
bootstrap-sass: n/a
ember-bootstrap configuration:
bootstrapVersion: 5
importBootstrapCSS: false
```

The focus-trap on BSModal is deactivated by pressing escape, even if @keyboard={{false}}, or by clicking outside of the dialog, even if @backdropClose={{false}}

Will submit a PR for fix.

kdagnan:
Addressed by #2159 and fixed in 6.4.1.

PR #2159, opened 2024-10-23
Modal should keep focus on escape key if @keyboard=false and on backdrop click @backdropClose=false

kdagnan:
Solves #2158

### DIFF
--- a/ember-bootstrap/addon/components/bs-modal/dialog.hbs
+++ b/ember-bootstrap/addon/components/bs-modal/dialog.hbs
@@ -35,7 +35,7 @@
     <div
       class="modal-content"
       tabindex="-1"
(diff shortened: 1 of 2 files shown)


jelhan:
Thanks a lot for reporting and fixing this! Do you mind adding a test to ensure that we don't introduce a regression?

jelhan (approved):
Looks good to me. Thanks a lot for adding test cases as well.

jelhan (changes requested):
Please take care of the linting errors.

jelhan (changes requested):
Linting is still failing.

kdagnan:
Sorry, I was expecting that `npm test` ran linting too. Should *actually* be good now

jelhan (approved):

jelhan:
Released as [v6.4.1](https://github.com/ember-bootstrap/ember-bootstrap/releases/tag/v6.4.1).

PR #2162, opened 2024-11-20
Cleanup outdated information about Bootstrap 3 from BsNavbar API docs

jelhan:
Support for Bootstrap 3 has been dropped long ago. The reference to it in API docs should be removed as well. I just noticed that we had it here when looking something up in the docs.

### DIFF
--- a/ember-bootstrap/addon/components/bs-navbar.js
+++ b/ember-bootstrap/addon/components/bs-navbar.js
@@ -13,10 +13,8 @@ import { isBlank } from '@ember/utils';
 
   ```hbs
   <BsNavbar as |navbar|>
-    <div class="navbar-header">
-      <navbar.toggle />
-      <a class="navbar-brand" href="#">Brand</a>
-    </div>
+    <a class="navbar-brand" href="#">Brand</a>
+    <navbar.toggle />
     <navbar.content>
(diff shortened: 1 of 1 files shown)


ember-bootstrap/addon/components/bs-navbar.js
```diff
@@ -13,10 +13,8 @@ import { isBlank } from '@ember/utils';
 
   ```hbs
   <BsNavbar as |navbar|>
-    <div class="navbar-header">
-      <navbar.toggle />
-      <a class="navbar-brand" href="#">Brand</a>
-    </div>
+    <a class="navbar-brand" href="#">Brand</a>
+    <navbar.toggle />
```
jelhan, line 17: `.navbar-brand` link is before the toggle button in [Bootstrap docs](https://getbootstrap.com/docs/5.3/components/navbar/). I switched the order here accordingly.

SanderKnauff (approved):
Looks good! Are these the only BS3 references?

jelhan:
> Are these the only BS3 references?

I haven't checked all API docs. I only checked the once in those files by searching for "3". I felt I should clean up those left overs when I noticed them. But I didn't had time looking if there is more.

PR #2164, opened 2024-11-28
Convert BsProgress to TypeScript

SanderKnauff:


ember-bootstrap/addon/template-registry.ts
```diff
@@ -1,9 +1,11 @@
 import type BsButton from './components/bs-button';
 import type BsCollapse from './components/bs-collapse';
 import type BsSpinner from './components/bs-spinner';
+import type BsProgressBar from 'ember-bootstrap/components/bs-progress/bar';
 
 export default interface EmberBootstrapRegistry {
   BsButton: typeof BsButton;
   BsCollapse: typeof BsCollapse;
   BsSpinner: typeof BsSpinner;
+  'bs-progress/bar': typeof BsProgressBar;
```
simonihmig, line 10: I think this should have `BsProgress` (not dasherized), and only that. The bar component is yielded, but should not be invoked directly (IIRC)

ember-bootstrap/addon/components/bs-progress/bar.ts
```diff
@@ -1,5 +1,22 @@
...
+export interface ProgressBarSignature {
+  Args: {
+    minValue?: number;
+    maxValue?: number;
+    value?: number;
+    showLabel?: boolean;
+    striped?: boolean;
+    animate?: boolean;
+    roundDigits?: number;
+    type?: 'default' | 'success' | 'info' | 'warning' | 'danger' | string;
```
simonihmig, line 12: This union type is effectively just `string`, so does not help much. Any typo like `defualt` would still be valid. Can't we list only the types Bootstrap supports ootb?

simonihmig (changes requested):
Thanks for working on this!

ember-bootstrap/addon/template-registry.ts
SanderKnauff: Ah, these should've been added to the internal types I think.
We still use the dasherized version with the `component` helper. Once everything is in GJS/GTS we should be able to get rid of those.

ember-bootstrap/addon/components/bs-progress/bar.ts
SanderKnauff: We could change the `string` to `string & {}`[^1]. This would allow for users to use their own defined bar types while still enjoying the benefits of auto-complete, however I'm not sure how often that case is used. I'm fine with being more strict/correct

[^1]: https://stackoverflow.com/questions/74467392/autocomplete-in-typescript-of-literal-type-and-string

simonihmig: Interesting, never heard of that!

But I still think being strict is probably better. Users that have this customized can still choose to not add our default types to their template registry, but have their of for a particular component they have extended, or extend our types. That is explicitly covered in the Glint docs [here](https://typed-ember.gitbook.io/glint/environments/ember/using-addons#using-glint-enabled-addons).

simonihmig (approved):
:tada:

PR #2171, opened 2025-02-16
remove unused arguments introduced by mistake

jelhan:
I noticed some arguments, which are either documented as internal or not documented at all, when refactoring `<BsDropdown>` and related components to TypeScript. Those are not used internally. They seem to be introduced by mistake when refactoring to `@glimmer/component`. I think we can drop them.


SanderKnauff (approved):
Great to hear you're working on converting these components to Typescript!

I don't see any problems with the changes, especially since the tests still all seem to be running.

PR #2177, opened 2025-02-26
Convert bs-accordion to Typescript

SanderKnauff:
There are quite a view `unknown` values here, but they do not seem to impact type checking too much.
The reason for this is that, while generic signatures do work, it down the line mashes with the fallback behavior for when an accordion item has no value, in which case it will fallback to `guidFor` (See `addon/components/bs-accordion/item.ts:55`).


jelhan (changes requested):
Looks good to me overall.

I'm not a TypeScript expert. I fear I cannot help with the `unknown`. I wouldn't have an idea how it would look like using generics to be honest.

Only change request is adding it to the template registry. I fear that's missing.

SanderKnauff:
> Looks good to me overall.
> 
> I'm not a TypeScript expert. I fear I cannot help with the `unknown`. I wouldn't have an idea how it would look like using generics to be honest.
> 
> Only change request is adding it to the template registry. I fear that's missing.

Ah yes, I forgot about that one!

Should we only put the main components in the registry, or the child components such as the bs-accordion/item, bs-accordion/item/title and bs-accordion/item/body as well?

In that case there are some items related to BsModal missing as well.

jelhan:
> Should we only put the main components in the registry, or the child components such as the bs-accordion/item, bs-accordion/item/title and bs-accordion/item/body as well?

We expect consumers to only invoke the child components when they are yielded from the main components. Therefore we only need the main components in the registry.

jelhan (approved):

PR #2178, opened 2025-03-01
Convert bs-carousel to Typescript

SanderKnauff:


ember-bootstrap/addon/components/bs-carousel.ts
```diff
@@ -624,14 +674,18 @@ export default class Carousel extends Component {
   }
 
   @action
-  registerChild(element) {
+  //  This means any component. Component and Component<unknown> don't work for this purpose.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  registerChild(element: Component<any>) {
```
SanderKnauff, line 679: These two methods _could_ be stricter and only accept `bs-carousel/slide`s, but for some reason, the children array could contain anything, but was only used by the `get childSlides` function that filters using an `instanceof` on `CarouselSlide`.

This feels like something that should be deprecated and cleaned up, as we only expect our provded slide components. This also would mean that the `@slideComponent` parameter on `BsCarousel` can be removed, since any slides using that would not be included in the carousel anyway...

jelhan: I'm not sure if I can follow. Defining your own yielded slides component can be helpful to customize the look and feel. But this custom slides component must be API compliant. E.g. it must call the `registerChild` method as expected.

We don't provide the same stability guarantees for those customization APIs as we do for the common APIs. I think we have introduced breaking changes for consumers providing custom components in the past. I feel that's okay as only a few power users are using it. And it would otherwise slow down development a lot.

In this case, can't we use `ComponentLike<Carousel slide>`?

SanderKnauff: > I'm not sure if I can follow. Defining your own yielded slides component can be helpful to customize the look and feel. But this custom slides component must be API compliant. E.g. it must call the registerChild method as expected.

It is, but even if you register anything, we seem to filter out only the Components that extend from CarouselSlide.
```ts
  // bs-carousel.ts:133
  get childSlides(): CarouselSlide[] {
    return this.children.filter(
      (view) => view instanceof CarouselSlide,
    ) as unknown as CarouselSlide[];
  }
```

As for `ComponentLike<CarouselSlideSignature>`, that does not work since an instance of `CarouselSlide` is not assignable to `ComponentLike<CarouselSlideSignature>`, as odd as that might sound.
I'm not sure if there is a way to properly type this.
I'll check in the EmberJS discord tomorrow, perhaps they have some ideas on how to do this.

jelhan:
I went through the rest of the code. All looks good beside the `@slideComponent` argument issue and the minor refactoring to avoid `isNone`, which I already mentioned in the last review.

SanderKnauff:
> I went through the rest of the code. All looks good beside the `@slideComponent` argument issue and the minor refactoring to avoid `isNone`, which I already mentioned in the last review.

Over the last two weeks I have done some attempts to get this working without the any, but no success however. I did add a test to prove that it is possible to have custom slide components. To keep the conversion effort going, we might want to just merge it like this as there is some type checking at least on the component level. It's just the registration callback that is a bit more loose than we want. Perhaps we can find an alternative or solution in the future.

PR #2180, opened 2025-03-03
support @ember/render-modifiers v3

jelhan:
This adds support for `@ember/render-modifiers` v3. `@ember/render-modifiers` v2 does not support Ember v6.


SanderKnauff (approved):
Looks good! No issues with merging this.

jelhan:
It seems that `{{create-ref}}` from [ember-ref-bucket](https://github.com/lifeart/ember-ref-bucket) is broken in Embroider after this change. I don't know yet. It seems that many transitive dependencies were upgraded by this change. Likely one caused an incompatibility. It's especially surprising as it was working before merging latest master in. I will investigate later.

jelhan:
Regenerating the lockfile solved the issue.

PR #2191, opened 2025-06-23
resolve deprecation "do not import inject from @ember/service"

### LINKED ISSUE #2190: Importing `inject` from `@ember/service` is deprecated
DEPRECATION: Importing `inject` from `@ember/service` is deprecated. Please import `service` instead. [deprecation id: importing-inject-from-ember-service] This will be removed in ember-source 7.0.0. See https://deprecations.emberjs.com/id/importing-inject-from-ember-service for more details.

ZacharyMohler:
closes #2190

### DIFF
--- a/docs/app/controllers/demo.js
+++ b/docs/app/controllers/demo.js
@@ -1,8 +1,6 @@
+import { service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isEmpty } from '@ember/utils';
-import * as emberService from '@ember/service';
-
-const service = emberService.service ?? emberService.inject;
 
 export default class DemoController extends Controller {
   @service
--- a/docs/app/controllers/demo.js
+++ b/docs/app/controllers/demo.js
@@ -1,6 +1,8 @@
-import { inject as service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isEmpty } from '@ember/utils';
+import * as emberService from '@ember/service';
+
+const service = emberService.service ?? emberService.inject;
 
 export default class DemoController extends Controller {
   @service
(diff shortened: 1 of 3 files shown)


docs/app/controllers/demo.js
```diff
@@ -1,6 +1,8 @@
-import { inject as service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isEmpty } from '@ember/utils';
+import * as emberService from '@ember/service';
+
+const service = emberService.service ?? emberService.inject;
```
jelhan, line 5: This isn't needed. Ember Bootstrap requires Ember 4.8 or above. The named import `service` is available since Ember 4.1. We can simplify use the new import.

jelhan (changes requested):
Thanks a lot for not only reporting but also providing a PR for it.

jelhan:
It seems some other deprecations which have been added in Ember 6.5 break the build. We need to address those as well to have a green build again.

Do you have time fixing those as well? If not, you can add them to the ignore lust to fix the CI: https://github.com/ember-bootstrap/ember-bootstrap/blob/6d84115c77540429235adbfaece2ac8e15c2e895/test-app/tests/helpers/setup-no-deprecations.ts#L10

ZacharyMohler:
It's currently nearing the end of my day—I'd be happy to try my hand and at fixing them if they're still waiting for me tomorrow

jelhan:
I'm going to merge this even though CI is failing for some jobs. The failing jobs are limited to new releases. And they are caused by other deprecations not fixed here. This PR improves the situation. Even though it does not fully solve it.

PR #2193, opened 2025-06-27
Fix deprecations triggered by importing from ember barrel file

locks:
- `Ember.testing`
- `Ember.ViewUtils.getViewBounds`
- `import { inject as service } from '@ember/service';`
- `Ember.libraries`

### DIFF
--- a/docs/app/controllers/demo.js
+++ b/docs/app/controllers/demo.js
@@ -1,4 +1,4 @@
-import { inject as service } from '@ember/service';
+import { service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isEmpty } from '@ember/utils';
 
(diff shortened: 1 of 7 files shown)


SanderKnauff (changes requested):

jelhan (approved):
Deprecations still cause the CI to fail. Likely those are triggered by dependencies and only affect our test suite. I think we should add them to the ignore list. Either in this PR or a follow-up.

https://github.com/ember-bootstrap/ember-bootstrap/blob/0ea05521b2833cf6d71355f09290b1c1a2a7fcb9/test-app/tests/helpers/setup-no-deprecations.ts#L9-L10

locks:
> Deprecations still cause the CI to fail. Likely those are triggered by dependencies and only affect our test suite. I think we should add them to the ignore list. Either in this PR or a follow-up.
> 
> https://github.com/ember-bootstrap/ember-bootstrap/blob/0ea05521b2833cf6d71355f09290b1c1a2a7fcb9/test-app/tests/helpers/setup-no-deprecations.ts#L9-L10

I added the `VERSION` deprecation and have an unpushed commit with more deprecation ids I found in the CI logs.

jelhan (approved):
Tests for Safari are still failing. But that's not related to the changes in this PR. And this PR improves the as-is state a lot. I'm going to merge it.

PR #2195, opened 2025-07-25
Disable modal animation tests in Safari

SanderKnauff:
I do not have a Mac, and Epiphany (a Webkit browser) does not result in the same test failures.

For now we can opt to disable these tests when running in Safari.

I have opened #2197 to address this

### DIFF
--- a/test-app/tests/integration/components/bs-modal-test.gts
+++ b/test-app/tests/integration/components/bs-modal-test.gts
@@ -383,86 +383,90 @@ module('Integration | Component | bs-modal', function (hooks) {
   );
 
   // TODO #2197: Fix modal animation tests on Safari
-  module.if('it animates opening and closing the modal', !isSafari, function () {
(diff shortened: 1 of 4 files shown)


test-app/tests/integration/components/bs-modal-test.gts
```diff
@@ -380,86 +382,89 @@ module('Integration | Component | bs-modal', function (hooks) {
     },
   );
 
-  module('it animates opening and closing the modal', function () {
-    hooks.before(function () {
-      skipTransition(false);
-    });
+  // TODO #2197: Fix modal animation tests on Safari
+  if (!isSafari) {
+    module('it animates opening and closing the modal', function () {
```
SanderKnauff, line 387: Qunit docs state that `modal.if` should be an option, however the @types/qunit packages does not have definitions for this.

PR #2203, opened 2025-09-26
remove blueprints not supported in v2 addon

jelhan:
This removes support for setting up Ember Bootstrap automatically when installing it via `ember install ember-bootstrap`. 

Motivation:

* The blueprints used for doing so, are not supported in v2 addons. We need to drop it as part of preparing for publishing Ember Bootstrap as a v2 addon.
* Setting up Ember Bootstrap in a project is a one time task. It boils down to a few steps, which can be easily done manually:
   1. Install Bootstrap NPM package (and ember-cli-sass if SASS preprocessor should be used)
   2. Import the Bootstrap CSS or SASS
   3. Configure Ember Bootstrap if needed

This is not a breaking change. It removes functionality. But functionality only used when introducing Ember Bootstrap to a new project or altering its configuration. The functionality is nothing a consumer depends on.


docs/app/templates/getting-started/assets.hbs
```diff
@@ -64,13 +64,12 @@ module.exports = function(defaults) {
     href="https://github.com/adopted-ember-addons/ember-cli-sass"
     target="_blank"
     rel="noopener noreferrer"
-  >ember-cli-sass</a>) as a CSS preprocessor. When it detects it upon first
-  installation of ember-bootstrap, it will install the necessary packages and
-  add an
+  >ember-cli-sass</a>) as a CSS preprocessor. Ensure that you have the Bootstrap
+  package installed and add an
   <code>@import</code>
```
SanderKnauff, line 69: Should we still recommend using `@import`?
It seems like SASS is trying to slowly get rid of it in favor of `@use` and `@forward`.

docs/app/templates/getting-started/setup.hbs
```diff
@@ -49,86 +53,49 @@ See the
...
+  option to
+  <code>4</code>
+  in your
+  <code>ember-cli-build.js</code>:
+
+  {{! prettier-ignore }}
+  <pre class="highlight"><code
+>module.exports = function (defaults) {
+  const app = new EmberApp(defaults, {
+    'ember-bootstrap': {
```
SanderKnauff, line 83: I wonder if these fields should be changed to `@embroider-macros`' `setConfig` field.

I'm not well versed yet in how Vite apps / future EmberJS apps should behave. Is this change something we can attempt through the ember-try scenarios?

docs/app/templates/getting-started/setup.hbs
```diff
@@ -203,9 +170,18 @@ for more information about including assets into your app.
...
-    'bs-accordion'] } }); return app.toTree(); };
-  </code></pre>
\ No newline at end of file
+
+module.exports = function(defaults) {
+  const app = new EmberApp(defaults, {
+    'ember-bootstrap': {
+      bootstrapVersion: 4,
+      importBootstrapCSS: false,
+      exclude: ['bs-popover', 'bs-accordion'],
```
SanderKnauff, line 181: Not related to blueprints, but will this `excludes/includes` structure work in a V2 addon?

SanderKnauff (approved):
The changes look good!
Also very nice to see that some lingering BS3 references have been updated.

I just have a few small questions, but I think those are more related to the V2 conversion as V2 conversion itself.

docs/app/templates/getting-started/setup.hbs
jelhan: No. It won't work. But it's also not needed anymore in modern Embroider apps. Embroider only pulls those components into the build which are actually used. No need to provide a list manually. We should deprecate the feature and recommend relying on Embroider's tree shaking capabilities instead.

docs/app/templates/getting-started/assets.hbs
jelhan: Bootstrap still uses `@import` in its official documentation: https://getbootstrap.com/docs/5.3/customize/sass/ I feel we should stay aligned with Bootstrap documentation in that regard.

jelhan:
The failing tests for Ember beta and canary scenarios are not related. I merge this one ignoring those. Let's fix them in a separate PR.

PR #2187, opened 2025-04-30
fix browser Uncaught (in promise) error when we use a popover with the hover trigger and delay

stukalin:
Closes #2186

PR #2204, opened 2025-10-02
Fix popover/tooltip crash when hiding before element renders

### LINKED ISSUE #2186: Popover throws the `Uncaught (in promise)` error when we user the hover trigger with the delay
**Steps**

1. Use a popover with the following combination of parameters: `@delayShow={{some relatively large number e.g. 1000}} @triggerEvents="hover"`
2. Hover over the triggering element and immediately leave before the popover is rendered

**Expected result**
Nothing should happen

**Actual result**
There is the 

Image

error in the console

emmanuel-ferdman:
## PR Summary
This PR fixes a race condition that occurs when using tooltips or popovers with `@triggerEvents="hover"` and `@delayShow`. When a user hovers and quickly leaves before the delay completes, the component may attempt to hide before the overlay element has fully rendered, causing an "Uncaught (in promise) undefined" error. This fix adds null checks for `overlayElement` in both `_show()` and `hide()` methods before calling `transitionEnd()`, ensuring the component gracefully handles cases where the element hasn't been created yet.

Fixes #2186.

### DIFF
--- a/test-app/tests/integration/components/bs-popover-test.gjs
+++ b/test-app/tests/integration/components/bs-popover-test.gjs
@@ -2,6 +2,7 @@ import { module, skip } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { click, render, settled, triggerEvent } from '@ember/test-helpers';
 import {
+  delay,
   popoverArrowClass,
   popoverPositionClass,
   test,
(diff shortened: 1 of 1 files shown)


emmanuel-ferdman:
@jelhan good idea - I took a look at the tests from #2187 and they look solid for covering the race condition. I've added them to this PR with some minor improvements.

Issue #2205, opened 2025-10-04
Deprecate tree-shaking via exclude / include options


jelhan:
Ember Bootstrap provides `exclude` and `include` options. Those allow a consuming build to only pull in those components needed to optimize the bundle size.

Manually listing the needed components is not needed anymore for apps using Embroider. Embroider supports tree-shaking out of the box. Only those components actually used in the app are pulled into the build.

We cannot support manual tree-shaking via `exclude` and `include` options when converting to a v2 addon. V2 addons do not have the capabilities to manipulate the consuming app's build.

We should deprecate the `exclude` and `include` options. The depreciation message should recommend migrating to Embroider and enabling`staticInvocables` option.

jelhan:
Implemented by #2206

PR #2206, opened 2025-10-17
feat: deprecate include and exclude configuration options

kpj2006:
fix:#2205
super of :#2203

### Summary of Changes

1. **Added Deprecation Warnings (`index.js`)**

   * Updated the `filterComponents()` method to emit deprecation warnings when `include` or `exclude` options are used.
   * Warnings recommend migrating to **Embroider** with the `staticInvocables` or `staticComponents` options for automatic tree-shaking.
   * Deprecations indicate these options will be **removed in v7.0.0**.

2. **Updated Documentation (`setup.hbs`)**

   * Marked both `include` and `exclude` configuration options as **deprecated** with visual warning badges.
   * Added clear migration guidance pointing users to Embroider.
   * Explained that these options will not work with V2 addons and that automatic tree-shaking is the recommended approach.

3. **Added Tests (`index-tests.js`)**

   * Added three new test cases verifying that deprecation warnings are properly emitted:

     * Test for `include` option
     * Test for `exclude` option
     * Test for both options used together

4. **Updated CHANGELOG (`CHANGELOG.md`)**

### DIFF
--- a/ember-bootstrap/CHANGELOG.md
+++ b/ember-bootstrap/CHANGELOG.md
@@ -22,15 +22,6 @@
 
 
 
-
-
-## Unreleased
-
-#### :warning: Deprecation
-* Deprecated the `include` and `exclude` configuration options. Manual tree-shaking via these options is no longer supported as Ember Bootstrap migrates to v2 addons. These options will be removed in v7.0.0.
-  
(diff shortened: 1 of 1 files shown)


jelhan (changes requested):
Thanks a lot for picking this up.

PR #2207, opened 2025-11-13
skip Fastboot tests failing due to a regression in Ember >= 6.8

jelhan:


SanderKnauff (approved):
Looks good after the `test.if` change

PR #2243, opened 2025-11-26
upgrade node version used for development to fix tooling

jelhan:
[lint-staged](https://github.com/lint-staged/lint-staged) used in pre-commit hook was broken due to an incompatibility of a transitive dependency with Node 18. Trying to commit failed with the following error:

```
$ git commit -m "support @glimmer/component v2"
gitfile:///home/main/Code/open-source/ember-bootstrap/node_modules/.pnpm/string-width@8.1.0/node_modules/string-width/index.js:16
const zeroWidthClusterRegex = /^(?:\p{Default_Ignorable_Code_Point}|\p{Control}|\p{Mark}|\p{Surrogate})+$/v;
                              ^

SyntaxError: Invalid regular expression flags
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:152:18)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:299:14)

Node.js v18.20.8
 husky - pre-commit script failed (code 1)
```

This could only be bypassed by setting `--no-verify` flag.

Upgrading the node version pinned with Volta to 20 seems the most reasonable solution to me. We still ensure compatibility with Node 18 via CI. We will drop support for Node 18 in the next major anyways as it reached end of life (EOL) in April 2025.


SanderKnauff (approved):
Since this is mostly for development, is a newer node version than 20 risky?

Anyway, Node 20 is better already so I'm also fine with the change just being this.

jelhan:
> Since this is mostly for development, is a newer node version than 20 risky?

As long as we are a v1 addon there is a risk that a change breaks compatibility with projects using older node versions. I tend toward keeping doing as we have done for years until we finally migrated to a v2 addon and can stop worrying about node compatibility at all.

PR #2249, opened 2025-11-29
drop support for importing Bootstrap CSS and Bootstrap SCSS alias

jelhan:
This removes support for

1. the deprecated `importBootstrapCSS` configuration options and
2. the `ember-bootstrap/bootstrap` alias for Bootstrap SCSS.

Consuming application must pull in Bootstrap CSS in their app themselves going forward. Consuming applications are free doing so however they like.

We may want to improve the documentation how Bootstrap CSS can be pulled in by applications in a follow-up PR. There are many different options today:

1. Consuming Bootstrap CSS as plain CSS or via SCSS using a CSS preprocessor.
2. Using Ember Bootstrap in a classic app, in a v1 app with Embroider (+ Webpack), in a v2 app with Vite.

Especially usage in an Ember v2 app build with Vite does not seem to be covered yet by the existing docs. But I see improving the docs as out of scope for this PR.

### DIFF
--- a/ember-bootstrap/README.md
+++ b/ember-bootstrap/README.md
@@ -6,7 +6,7 @@
 
 An [ember-cli](http://www.ember-cli.com) addon for using [Bootstrap](http://getbootstrap.com/) 4 und 5 in Ember applications.
 
-The addon it provides a set of native Ember components for Bootstrap CSS. It does not rely on any Bootstrap JavaScript.
+The addon provides a set of native Ember components for Bootstrap CSS. It does not rely on any Bootstrap JavaScript.
 
 _See [www.ember-bootstrap.com](http://www.ember-bootstrap.com/) for full documentation._
 


ember-bootstrap/index.js
```diff
@@ -60,32 +54,14 @@ module.exports = {
...
-    }
 
-    // import custom addon CSS
-    this.import(path.join(vendorPath, `bs${options.bootstrapVersion}.css`));
+    // import custom addon CSS required to use Popper.js v2 with Bootstrap v4
+    if (options.bootstrapVersion === 4) {
+      let vendorPath = path.join('vendor', 'ember-bootstrap');
+
+      // import custom addon CSS
+      this.import(path.join(vendorPath, `bs4-overwrites-for-popper-js-v2.css`));
```
jelhan: I renamed the file to clarify why we have it. When migrating to a v2 addon, we must request the consumers to import it explicitly in their apps. I added that to the tracking issue for v7.

docs/app/templates/getting-started/assets.hbs
```diff
@@ -53,19 +52,13 @@ module.exports = function(defaults) {
...
+  You can use
   <a
     href="https://github.com/adopted-ember-addons/ember-cli-sass"
     target="_blank"
     rel="noopener noreferrer"
-  >ember-cli-sass</a>) as a CSS preprocessor. Ensure that you have the Bootstrap
-  package installed and add an
+  >Ember CLI SASS</a>) if you want to use a CSS preprocessor. Ensure that you
+  have the Bootstrap package installed and add an
   <code>@import</code>
```
SanderKnauff: This should be an @use or @forward in modern SCSS.

PR #2254, opened 2026-01-08
move BOOSTRAPVERSION env variable to test-app

jelhan:
Ember Bootstrap allowed setting the bootstrap version via `BOOTSTRAPVERSION` environment variable. This was not documented. And was not a public API. Theoretically consumers were able to use it. But likely the test app was the only one using the functionality to set the Bootstrap version depending on the test scenario.

This change moves the logic to the test app to which it belongs.

### DIFF
--- a/ember-bootstrap/index.js
+++ b/ember-bootstrap/index.js
@@ -40,10 +40,6 @@ module.exports = {
       );
     }
 
-    if (process.env.BOOTSTRAPVERSION) {
-      // override bootstrapVersion config when environment variable is set
-      options.bootstrapVersion = parseInt(process.env.BOOTSTRAPVERSION);
-    }
     this.bootstrapOptions = options;
 
     this.validateDependencies();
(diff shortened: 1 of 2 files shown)


SanderKnauff (approved):
Ah, I was under the assumption that the macros needed this. Do they just read from the consuming app's package.json?

Why can't the test app do the same?

The changes look good though, so I will approve!

jelhan:
> Ah, I was under the assumption that the macros needed this. Do they just read from the consuming app's package.json?
> 
> Why can't the test app do the same?
> 
> The changes look good though, so I will approve!

The macros need the Bootstrap version. But that one is typically set in the configuration of the consuming app:

```js
module.exports = function (defaults) {
  const app = new EmberApp(defaults, {
    'ember-bootstrap': {
      bootstrapVersion: 4,
    },
  });

  // ...
};
```

The `BOOTSTRAPVERSION` environment variable was overruling that setting (if being set). I think it was needed before test-app was moved out to a dedicated app because we haven't had an `ember-cli-build.js` file to configure the test-app.

PR #2259, opened 2026-01-20
support ember-concurrency v5

jelhan:


acoppola:
There are news on this one?

jelhan:
> There are news on this one?

We still haven't found a reliable way to support both ember-concurrency versions. We will support ember-concurrency v5 in next major. But it's unclear if we can port that support back to the current major.

PR is very welcome.

PR #2272, opened 2026-04-04
Update ember focus trap to 2 0 0

SanderKnauff:
The 1.2.0 version promised typescript types, but was broken.

2.0.0 fixes that and also upgrades the addon to v2, which reduces our v1 child-dependencies as well


jelhan (approved):
Looks good to me.

PR #2273, opened 2026-04-04
Fix iOS mouseover listener leak when tooltip/popover is destroyed while visible

glitch-ux:
## Summary

On touch-enabled devices (iOS), `bs-contextual-help` adds empty `mouseover` listeners to all `document.body` children in `_show()` as a [workaround for broken iOS event delegation](https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html) (mirroring [Bootstrap's own fix](https://github.com/twbs/bootstrap/pull/22481)).

These listeners were only cleaned up in `hide()`, but if the component is destroyed without `hide()` being called — for example during a route transition while a tooltip is visible — the listeners leak permanently. Over time this can accumulate thousands of stale event listeners, degrading performance and memory.

**Fix:** Extract the iOS listener cleanup into `_removeiOSMouseoverListeners()` and call it from both `hide()` and `willDestroy()`, ensuring cleanup regardless of how the component is torn down.

## Changes

- **`ember-bootstrap/addon/components/bs-contextual-help.ts`**: Extract iOS cleanup into `_removeiOSMouseoverListeners()` method, call it from `hide()` and `willDestroy()`
- **`test-app/tests/integration/components/bs-tooltip-test.gts`**: Add test that verifies iOS mouseover listeners are cleaned up when the component is destroyed without `hide()` being called

## Test plan

- [ ] Existing tests pass (no behavior change for the `hide()` path)
- [ ] New test verifies cleanup on destroy: simulates a touch device, renders a visible tooltip, destroys it via DOM removal (not `hide()`), and asserts `removeEventListener('mouseover', ...)` was called on body children

### DIFF
--- a/test-app/tests/integration/components/bs-tooltip-test.gts
+++ b/test-app/tests/integration/components/bs-tooltip-test.gts
@@ -772,4 +772,51 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     await a11yAudit();
     assert.ok(true, 'A11y audit passed');
   });
+
+  test('it cleans up iOS mouseover listeners on destroy', async function (assert) {
+    // Simulate a touch-enabled device so the iOS workaround code path is exercised
+    const hadOntouchstart = 'ontouchstart' in document.documentElement;
+    if (!hadOntouchstart) {
+      // Setting to null makes `'ontouchstart' in document.documentElement` return true
+      (document.documentElement as HTMLElement & { ontouchstart: unknown }).ontouchstart = null;
+    }
+
+    const bodyChild = document.body.children[0]!;
+    const removeListenerSpy = sinon.spy(bodyChild, 'removeEventListener');
+
+    class State {
+      @tracked show = true;
+    }
+    const state = new State();
+
+    try {
+      await render(
+        <template>
+          {{#if state.show}}
+            <button type='button'>
+              Test
+              <BsTooltip @title='Dummy' @visible={{true}} />
+            </button>
+          {{/if}}
+        </template>,
+      );
+      await settled();
+
+      // Destroy the component by removing it from the DOM.
+      // This triggers willDestroy() without calling hide() first,
+      // which previously leaked the iOS mouseover listeners.
+      state.show = false;
+      await settled();
+
+      assert.ok(
+        removeListenerSpy.calledWith('mouseover'),
+        'iOS mouseover listeners are removed when component is destroyed without hide()',
+      );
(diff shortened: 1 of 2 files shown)


test-app/tests/integration/components/bs-tooltip-test.gts
```diff
@@ -772,4 +772,51 @@ module('Integration | Component | bs-tooltip', function (hooks) {
...
     assert.ok(true, 'A11y audit passed');
   });
+
+  test('it cleans up iOS mouseover listeners on destroy', async function (assert) {
+    // Simulate a touch-enabled device so the iOS workaround code path is exercised
+    const hadOntouchstart = 'ontouchstart' in document.documentElement;
+    if (!hadOntouchstart) {
+      // Setting to null makes `'ontouchstart' in document.documentElement` return true
+      (document.documentElement as HTMLElement & { ontouchstart: unknown }).ontouchstart = null;
+    }
```
jelhan, line 782: Can we move the setup and clean-up logic into QUnit `before` and `after` hooks to improve readability? Let's introduce a module dedicated for iOS compatibility related tests to encapsulate those.

test-app/tests/integration/components/bs-tooltip-test.gts
```diff
@@ -772,4 +772,51 @@ module('Integration | Component | bs-tooltip', function (hooks) {
...
+        <template>
+          {{#if state.show}}
+            <button type='button'>
+              Test
+              <BsTooltip @title='Dummy' @visible={{true}} />
+            </button>
+          {{/if}}
+        </template>,
+      );
+      await settled();
```
jelhan, line 803: Why do we need to await settled here? There wasn't any state change after render.

test-app/tests/integration/components/bs-tooltip-test.gts
```diff
@@ -772,4 +772,51 @@ module('Integration | Component | bs-tooltip', function (hooks) {
     await a11yAudit();
     assert.ok(true, 'A11y audit passed');
   });
+
+  test('it cleans up iOS mouseover listeners on destroy', async function (assert) {
```
jelhan, line 776: Do we have a test that those are registered in the first place? Or do we test that explicitly by running our test suite in Safari?

jelhan (changes requested):
Good catch.

I have a few change requests. Please have a look.

Please run Prettier in addition to fix formatting which breaks CI.

